PR #22, opened 2024-01-13
Config unit tests for CI

### LINKED ISSUE #10: Add CI workflows
Add static checks, unit / integration test workflows

mikealfare:
resolves #10

Setup unit tests to run automatically against `main`

### Checklist


mikealfare:
Requires #27 to work

mikealfare:
Requires #28

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,49 @@
...
+        run: hatch run unit-tests
+        shell: bash
+
+      - name: Publish results
+        uses: ./.github/actions/publish-results
+        if: always()
+        with:
+          source-file: "results.csv"
+          file-name: "unit_results"
+          python-version: ${{ matrix.python-version }}
```
emmyoop, line 49: Why are you doing this?

.changes/unreleased/Under the Hood-20240112-230236.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Under the Hood
+body: Implement unit testing in CI
```
emmyoop, line 2: Generally we don't have changelogs for CI changes.

emmyoop (approved):
This looks fine.  Just a question on why you save the results.

We should find out if we need both the matrix for python version in the workflow and the python version tests defined in pyroject.toml.  Not a blocker to getting tests running on your PRs though.

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
```
emmyoop, line 30:
```suggestion
        uses: actions/checkout@v4
```

.changes/unreleased/Under the Hood-20240112-230236.yaml
mikealfare: Fair, force of habit. We can clean that up afterwards.

.github/workflows/unit_tests.yml
mikealfare: I'm mimicking the original workflow. I can remove this if it's not needed.

mikealfare:
>We should find out if we need both the matrix for python version in the workflow and the python version tests defined in pyroject.toml.

Agreed. My concern is that I don't know how to run integration tests in parallel using the hatch env matrix. They would run in parallel on the machine, but not as separate workflows in GHA. We're already running into memory issues for Redshift on Windows, so I want to split across machines. There might be a plugin for hatch that does this, I just haven't looked into it.

PR #43, opened 2024-01-23
update dbt-adapters to include unit test feature updates

colin-rogers-dbt:
port unit testing adapters changes from https://github.com/dbt-labs/dbt-core/pull/8411

### Checklist

### DIFF
--- a/dbt/adapters/base/relation.py
+++ b/dbt/adapters/base/relation.py
@@ -228,7 +228,7 @@ class BaseRelation(FakeAPIObject, Hashable):
     def create_ephemeral_from(
         cls: Type[Self],
         relation_config: RelationConfig,
-        limit: Optional[int],
+        limit: Optional[int] = None,
     ) -> Self:
         # Note that ephemeral models are based on the name.
         identifier = cls.add_ephemeral_prefix(relation_config.name)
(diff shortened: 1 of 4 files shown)


MichelleArk:
Looks good regarding the macro changes! Something small I noticed was that `adapterse/base/relation.py` was also updated by https://github.com/dbt-labs/dbt-core/pull/8411 and those changes don't seem to be reflected on this branch: [relation.py changes on unit testing branch](https://github.com/dbt-labs/dbt-core/pull/8411/files#diff-8b301fa58dbc12c519f4bf6e3056605c09aea9da46b21d2ec5fd979cfd4f38ad)

I don't entirely recall why/if this change was strictly necessary. I don't see any usage of the method in dbt-core so I suspect it was just a bit of clean-up to reflect the optionality of `limit` -- but @gshank may have a better recollection ([source commit of relation.py changes](https://github.com/dbt-labs/dbt-core/pull/9184/files#))

mikealfare:
@colin-rogers-dbt Are there any associated tests that also need to move over? Should I look into that as part of the testing migration?

colin-rogers-dbt:
> Are there any associated tests that also need to move over? Should I look into that as part of the testing migration?

@mikealfare I'm guessing so, I think that will happen naturally as part of the PR to delete dbt-adapter-tests from core/when you break out dbt-postgres

mikealfare:
> > Are there any associated tests that also need to move over? Should I look into that as part of the testing migration?
> 
> @mikealfare I'm guessing so, I think that will happen naturally as part of the PR to delete dbt-adapter-tests from core/when you break out dbt-postgres

Sounds good. I added an item to our migration list to make sure we don't lose it. So consider the testing half of this feature as out of scope for this PR.

mikealfare (approved):
These look like all of the functional code changes from the linked PR.

PR #53, opened 2024-01-25
Fix _get_cache_schemas by removing relation identifiers

### LINKED ISSUE #52: [Bug] dbt commands running the same metadata query way too many times

### Current Behavior

Screen Shot 2024-01-25 at 1 41 49 PM

Check out all the `show terse objects ...` queries above.

### Expected Behavior

Run the `show terse objects ...` query a minimal number of times.

### Steps To Reproduce

1. Using the bleeding edge version in dbt Cloud
2. Using internal analytics project
3. Run `dbt clone`

### Relevant log output

```shell
See screenshot above.
```

### Environment

```markdown
- OS:
- Python:
- dbt:
```

snowflake

### Additional Context

_No response_

jtcohen6:
resolves #52

### Problem

We were running one caching query per relation (resource) defined in the project

### Solution

We should only run one caching query per schema, by dropping the relation-specific identifier within `_get_cache_schemas`

### Checklist


dbt/adapters/base/impl.py
```diff
@@ -421,7 +421,7 @@ def _get_cache_schemas(self, relation_configs: Iterable[RelationConfig]) -> Set[
         populate.
         """
         return {
-            self.Relation.create_from(quoting=self.config, relation_config=relation_config)
```
MichelleArk, line 424: Looks like we unintentionally removed .without_identifier as part of the adapter decoupling: https://github.com/dbt-labs/dbt-core/pull/8906/files#diff-aa1e9e602a0f1e2930b465b30a35a28778fbe131da4fbf2c4e88dd52ca44f140R422-R423

MichelleArk (approved):

dbt/adapters/base/impl.py
mikealfare: @colin-rogers-dbt @McKnight-42 I wonder if this was the issue we were seeing in that `dbt-snowflake` test. It feels like more than a coincidence.

PR #55, opened 2024-01-29
case-insensitive comparisons in unit testing, base unit testing test

MichelleArk:
resolves https://github.com/dbt-labs/dbt-core/issues/9467

[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/

### Problem
No base test exists for adapter-specific functionality for the new unit testing framework in dbt-core. This functionality includes pieces like:
* mocking out inputs with different types
* case-insensitive 'actual vs expected' query building
* input validation

### Solution

* Introduce base tests for types & case insensitivity 
* Update unit testing macros to generate a case-insensitive unit test query + introduce a pattern for implementing a common cast macro that safe_cast can leverage as appropriate. 

### Checklist

Note: This repo does not yet run integration tests, but I've tested these in a local venv built from dbt-core but pointing to these changes for postgres. Other 1p adapters inherit the base tests and are run in separate repos


MichelleArk:
Snowflake: https://github.com/dbt-labs/dbt-snowflake/pull/896
BigQuery: https://github.com/dbt-labs/dbt-bigquery/pull/1031
Redshift: https://github.com/dbt-labs/dbt-redshift/pull/680

gshank (dismissed):
Looks really good. Nice error message for invalid column names.

gshank (dismissed):

dbt/include/global_project/macros/utils/cast.sql
```diff
@@ -0,0 +1,7 @@
+{% macro cast(field, type) %}
```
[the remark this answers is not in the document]
mikealfare: How do we define "adapter-agnostic" here? We have to assume some flavor of SQL in order to have a default. Does this scenario differ from other places where we've taken the stance that there should be no default implementation (e.g. relation create statements)?

dbt/tests/adapter/unit_testing/test_case_insensitivity.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def test_case_insensitivity(self, project):
+        results = run_dbt(["run"])
+        assert len(results) == 2
+
+        results = run_dbt(["test"])
+
+
+class TestPosgresUnitTestCaseInsensitivity(BaseUnitTestCaseInsensivity):
+    pass
```
mikealfare, line 50: EOF

mikealfare (commented):
Nothing alarming. I have some cause for concern that we're bypassing quoting and case-sensitivity. But that's literally in the title and seems to be the intention, so I'll chalk that up as me lacking context.

dbt/include/global_project/macros/utils/cast.sql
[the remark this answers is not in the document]
dbeatty10: In this case, `cast` function is specified in ISO/IEC 9075-2 (part 2 of the SQL standard), so there is a natural adapter-agnostic implementation.

dbt/tests/adapter/unit_testing/test_case_insensitivity.py
```diff
@@ -0,0 +1,50 @@
+import pytest
+from dbt.tests.util import run_dbt
+from dbt.tests.fixtures.project import *
```
mikealfare, line 3: I would like to avoid using `from <module> import *` in general. It's less likely to cause an issue here since there are so few imports, but I'd still rather avoid it. Also, I don't see why this is needed here. The fixtures should already be registered via `tests/functional/conftest.py` or `tests/conftest.py`.

dbt/tests/adapter/unit_testing/test_case_insensitivity.py
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+    def test_case_insensitivity(self, project):
+        results = run_dbt(["run"])
+        assert len(results) == 2
+
+        results = run_dbt(["test"])
+
+
+class TestPosgresUnitTestCaseInsensitivity(BaseUnitTestCaseInsensivity):
```
mikealfare, line 49: There shouldn't be any actual test cases in `dbt-tests-adapter`. This should be implemented in the adapter implementation (e.g. `dbt-postgres`). One of the issues with creating a test case like this is that it cannot be imported without `pytest` running it. So it either doesn't get imported, gets imported and not used (accept to run it as-is), or gets imported because it needs to be modified to run. In the last case, the imported version would likely fail since the point is that it needs to be modified.

dbt/tests/adapter/unit_testing/test_invalid_input.py
```diff
@@ -0,0 +1,63 @@
+import pytest
+from dbt.tests.util import run_dbt, run_dbt_and_capture
+from dbt.tests.fixtures.project import *
```
mikealfare, line 3: same as above

dbt/tests/adapter/unit_testing/test_invalid_input.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+        assert "Invalid column name: 'invalid_column_name' in unit test fixture for 'my_upstream_model'." in out
+        
+        _, out = run_dbt_and_capture(
+            ["test", "--select", "test_name:test_invalid_expect_column_name"], expect_pass=False
+        )
+        assert "Invalid column name: 'invalid_column_name' in unit test fixture for expected output." in out
+
+
+class TestPostgresUnitTestInvalidInput(BaseUnitTestInvalidInput):
```
mikealfare, line 62: same as above

dbt/tests/adapter/unit_testing/test_types.py
```diff
@@ -0,0 +1,85 @@
+import pytest
+
+from dbt.tests.util import write_file, run_dbt
+from dbt.tests.fixtures.project import *
```
mikealfare, line 4: same as above

dbt/tests/adapter/unit_testing/test_types.py
```diff
@@ -0,0 +1,85 @@
...
+            results = run_dbt(["run", "--select", "my_upstream_model"])
+            assert len(results) == 1
+
+            try:
+                run_dbt(["test", "--select", "my_model"])
+            except Exception:
+                raise AssertionError(f"unit test failed when testing model with {sql_value}")
+
+
+class TestPostgresUnitTestingTypes(BaseUnitTestingTypes):
```
mikealfare, line 84: same as above

dbt/include/global_project/macros/unit_test_sql/get_fixture_sql.sql
```diff
@@ -59,18 +61,32 @@ union all
...
+            {{ exceptions.raise_compiler_error(
+                "Invalid column name: '" ~ column_name ~ "' in unit test fixture for " ~ fixture_name ~ "."
+                "\nAccepted columns for " ~ fixture_name ~ " are: " ~ (column_name_to_data_types.keys()|list)
+            ) }}
+        {%- endif -%}
+
+        {%- set column_type = column_name_to_data_types[column_name] %}
+        
+        {#-- wrap yaml strings in quotes, apply cast --#}
+        {%- if column_value is string -%}
```
mikealfare, line 81: Since this conditional only really cares about `column_value` and effectively maps it to be piped into `safe_cast`, would you consider something like:
```sql
{#-- wrap yaml strings in quotes --#}
{%- if column_value is string -%}
    {%- set column_value_clean = dbt.string_literal(dbt.escape_single_quotes(column_value)) -%}
{%- elif column_value is none -%}
    {%- set column_value_clean = 'null' -%}
{%- else -%}
    {%- set column_value_clean = column_value -%}
{%- endif -%}

{%- set row_update = {column_name: safe_cast(column_value_clean, column_type) } -%}

```

mikealfare (commented):
Sorry for the double review. I just missed these I guess.

dbt/include/global_project/macros/unit_test_sql/get_fixture_sql.sql
MichelleArk: sure, i like this in terms of readability

dbt/tests/adapter/unit_testing/test_case_insensitivity.py
MichelleArk: Totally agree! Discussed offline: I've left these here as part of this work because I understand dbt-postgres is not yet set up to be running its integration tests, and because there will be a migration of postgres tests defined in dbt-adapters into dbt-postgres and moving these can be part of that more unified lift.

mikealfare (approved):

PR #62, opened 2024-02-01
Point check back to dbt-core version instead of the adapter version

mikealfare:
Addresses failed CI in dbt-labs/dbt-bigquery#1082

### Problem

The existing integration tests in `dbt-core` validated against the `dbt-core` version, but the new integration tests tested against adapter version. 

### Solution

Point the test back to `dbt-core` for now since that's our testing dependency anyway.

### Checklist

### DIFF
--- a/dbt/tests/adapter/basic/test_docs_generate.py
+++ b/dbt/tests/adapter/basic/test_docs_generate.py
@@ -3,10 +3,6 @@ from datetime import datetime
 
 import pytest
 
-try:
-    from dbt.version import __version__ as DBT_VERSION
-except ImportError:
-    DBT_VERSION = "ERROR: dbt-core is not installed or `dbt.version.__version__` does not exist"
 from dbt.tests.adapter.basic import expected_catalog
 from dbt.tests.fixtures.project import write_project_files
 from dbt.tests.util import run_dbt, rm_file, get_artifact, check_datetime_between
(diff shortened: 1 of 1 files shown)


colin-rogers-dbt:
I believe dbt_version is set here: https://github.com/dbt-labs/dbt-bigquery/blob/e86609a1e15a766eb764a535f277100ca10ee67b/dbt/adapters/bigquery/connections.py#L42

So it should be the adapter version not dbt-adapters or dbt-core

colin-rogers-dbt:
see similar test patch: https://github.com/dbt-labs/dbt-adapters/commit/8f259f56e298c999337062589249f8673425f5db

Maybe we should add the adapter version as a global fixture?

mikealfare:
> I believe dbt_version is set here: https://github.com/dbt-labs/dbt-bigquery/blob/e86609a1e15a766eb764a535f277100ca10ee67b/dbt/adapters/bigquery/connections.py#L42
> 
> So it should be the adapter version not dbt-adapters or dbt-core

If you look at 1.7, this used to be the version of `dbt-core`. This was updated to the adapter version during the migration, but I don't think we updated the tests to refer to the adapter version. They still pull in the core version.

colin-rogers-dbt:
offline investigation shows that the `dbt_version` set here is set by artifacts: https://github.com/dbt-labs/dbt-core/blob/ad723a6db897d9779f817da522758c05dce65730/core/dbt/artifacts/schemas/base.py#L62

we need to settle on a more consistent approach to this field but for now I don't _think_ adapters need to validate core's artifacts are generated correctly

mikealfare:
> offline investigation shows that I'm incorrect and that the `dbt_version` set here is set by artifacts: https://github.com/dbt-labs/dbt-core/blob/ad723a6db897d9779f817da522758c05dce65730/core/dbt/artifacts/schemas/base.py#L62
> 
> we need to settle on a more consistent approach to this field but for now I don't _think_ adapters need to validate core's artifacts are generated correctly

Are you suggesting we just remove the assertion (line 339) entirely? I'm also alright with this approach for exactly the argument you made.

colin-rogers-dbt (approved):

PR #66, opened 2024-02-06
Add an implementation issue template for maintainers

mikealfare:
### Problem

We need a template for implementation issues.

### Solution

Reuse `dbt-core`'s template with some modification.

### Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/implementation-ticket.yml
+++ b/.github/ISSUE_TEMPLATE/implementation-ticket.yml
@@ -1,0 +1,66 @@
+name: 🛠️ Implementation
+description: This is an implementation ticket intended for use by the maintainers of `dbt-adapters`
+title: "[<project>] <title>"
+labels: ["user docs","refinement"]
+body:
+  - type: markdown
+    attributes:
+      value: This is an implementation ticket intended for use by the maintainers of `dbt-adapters`
+
+  - type: checkboxes
+    attributes:
+      label: Housekeeping
+      description: >
+        A couple friendly reminders:
+          1. Remove the `user docs` label if the scope of this work does not require changes to https://docs.getdbt.com/docs: no end-user interface (e.g. yml spec, CLI, error messages, etc) or functional changes.
+          2. Will this change need to be backported? Add the appropriate `backport 1.x.latest` label(s).
+          3. Is this an `enhancement`, `bug`, or `regression`? Add the appropriate label.
+      options:
+        - label: I am a maintainer of `dbt-adapters`
+          required: true
+
+  - type: textarea
+    attributes:
+      label: Short description
+      description: |
+        Describe the scope of the ticket, a high-level implementation approach and any tradeoffs to consider.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Context
+      description: |
+        Provide the "why", motivation, and alternative approaches considered -- linking to previous refinement issues, spikes and documentation as appropriate
+    validations:
+      required: false
+
+  - type: textarea
+    attributes:
+      label: Acceptance criteria
+      description: |
+        What is the definition of done for this ticket? Include any relevant edge cases and/or test cases.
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/implementation-ticket.yml
```diff
@@ -0,0 +1,66 @@
...
+title: "[<project>] <title>"
+labels: ["user docs","refinement"]
+body:
+  - type: markdown
+    attributes:
+      value: This is an implementation ticket intended for use by the maintainers of `dbt-adapters`
+
+  - type: checkboxes
+    attributes:
+      label: Housekeeping
```
mikealfare, line 12: I removed the blocking issues item since we do not have that field in our project.
I moved backporting up to here since we have labels for it. We also use enhancement/bug/regression labels, so I added that item.

.github/ISSUE_TEMPLATE/implementation-ticket.yml
```diff
@@ -0,0 +1,66 @@
...
+    attributes:
+      label: Short description
+      description: |
+        Describe the scope of the ticket, a high-level implementation approach and any tradeoffs to consider.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Context
```
mikealfare, line 32: I moved the context section to before the AC. I think it helps to have this in your head when reading through the AC. This is similar to our Problem / Solution setup for PRs.

.github/ISSUE_TEMPLATE/implementation-ticket.yml
```diff
@@ -0,0 +1,66 @@
...
+        Provide scenarios to test. Include both positive and negative tests if possible. Link to existing similar tests if appropriate.
+      placeholder: |
+         1. Test with no `materialized` field in the model config. Expect pass.
+         2. Test with a `materialized` field in the model config that is not valid. Expect ConfigError.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Consequences
```
mikealfare, line 59: I broadened this section from impacted teams to consequences in general (good or bad). We don't always capture this for the discussion so I want to call it out explicitly. I see team impact as a proper subset of consequences. We could break impact out separately, or include it in housekeeping section since the rest of the labels are there. But it feels intuitive to leave it here with consequences.

.github/ISSUE_TEMPLATE/implementation-ticket.yml
martynydbt: for bugs we use the bug template, Core only uses this for feature work.

.github/ISSUE_TEMPLATE/implementation-ticket.yml
martynydbt: either way works for me as long as we are calling it out and ensuring those teams are looped in. If we don't explicitly call out other team/system impacts then we can forget.

martynydbt (approved):
Sounds good! This is a living doc so as we use it we can always adapt! Thanks for the PR Mike!

.github/ISSUE_TEMPLATE/implementation-ticket.yml
mikealfare: Ah, great. I'll update this to be enhancement-focused then, which removes some of the either/or steps. And I'll look at the bug template and add that to this PR.

.github/ISSUE_TEMPLATE/implementation-ticket.yml
mikealfare: I'll put it at the top with housekeeping then, and reference that here. Sometimes you don't realize it's impacting until after some detail is fleshed out.

martynydbt: true! Thanks for these adjustments and pulling it in!

martynydbt (approved):

PR #75, opened 2024-02-08
Update internal issue templates

mikealfare:
If both `placeholder` and `value` properties are provided, `value` wins. This means examples cannot be put in the `placeholder` property in these scenarios. Examples were moved into the `description` property.

A header is not needed for `tasklist` fields as it duplicates the field header. They were removed.

More fields were updated to `tasklist` fields. All `tasklist` fields use the `value` property to setup the tasklist structure.

Implementation tasks were moved up to be after Objectives for epics. These fields are more closely related than the previous structure suggested.

It may be easier to see the new versions by looking at the rendered forms:
- [epic](https://github.com/dbt-labs/dbt-adapters/blob/config/internal-issue-template/.github/ISSUE_TEMPLATE/internal-epic.yml)
- [feature](https://github.com/dbt-labs/dbt-adapters/blob/config/internal-issue-template/.github/ISSUE_TEMPLATE/internal-feature-implementation.yml)

Note that these rendered forms show the description below the input field. This is not how it appears when using the form.

### DIFF
--- a/.github/ISSUE_TEMPLATE/implementation-ticket.yml
+++ b/.github/ISSUE_TEMPLATE/implementation-ticket.yml
@@ -1,0 +1,66 @@
+name: 🛠️ Implementation
+description: This is an implementation ticket intended for use by the maintainers of `dbt-adapters`
+title: "[<project>] <title>"
+labels: ["user docs","refinement"]
+body:
+  - type: markdown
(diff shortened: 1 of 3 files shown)


martynydbt (approved):
thanks for taking them for a spin!

PR #91, opened 2024-02-15
Ignore adapter-level support for `custom` constraints

### LINKED ISSUE #90: [CT-3181] [Bug] `ConstraintType.custom` doesn't mix with adapter `CONSTRAINT_SUPPORT`

### Current Behavior

`ConstraintType.custom` is missing from `CONSTRAINT_SUPPORT`, leading to a `KeyError` during `process_parsed_constraint`.

We could fix this in either method, but we'll need to account for adapter-level overrides.

https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/adapters/base/impl.py#L226-L232

https://github.com/dbt-labs/dbt-core/blob/eac13e3bd38b7f29401444ca77ee243ec4d8f29d/core/dbt/adapters/base/impl.py#L1441-L1462

### Expected Behavior

There should not be a `KeyError`.

Custom constraints should be `ConstraintSupport.NOT_ENFORCED` by default. In the future, we could maybe let users tell us if a constraint is enforced or not (i.e. whether removing that constraint constitutes a breaking change).

### Steps To Reproduce

It's possible to use custom constraints for any attributes that need application within the column spec of the `create table` DDL. For example, custom compression values on columns in Redshift.

1. Create a model:
```sql
-- models/my_model.sql
select 1 as a_number
```

2. Define a contract for that model:
```yaml
models:

  - name: my_model
    config:
      materialized: table
      contract:
        enforced: true

    columns:
      - name: a_number
        data_type: integer
        constraints:
          - type: custom
            expression: "encode bytedict"
            warn_unenforced: False
            warn_unsupported: False
```

### Relevant log output

```shell
File "/Users/jerco/dev/product/dbt-core/core/dbt/adapters/base/impl.py", line 1434, in render_raw_columns_constraints
    c = cls.process_parsed_constraint(constraint, cls.render_column_constraint)
  File "/Users/jerco/dev/product/dbt-core/core/dbt/adapters/base/impl.py", line 1459, in process_parsed_constraint
    if cls.CONSTRAINT_SUPPORT[parsed_constraint.type] != ConstraintSupport.NOT_SUPPORTED:
KeyError: <ConstraintType.custom: 'custom'>
```

### Environment

```markdown
- OS: macOS 13.4.1
- Python: 3.10.11
- dbt: 1.7.0-b2 (main)
```

_No response_

### Additional Context

_No response_

jtcohen6:
resolves #90

I believe that fixing this bug would enable a workaround for defining tags / masking policies as part of Snowflake model contracts. This has recently come up as a blocking issue for multiple folks internally (see [internal slack thread](https://dbt-labs.slack.com/archives/C05CJT5MJF2/p1707781284556729)).

### Problem

`ConstraintType.custom` is missing from `CONSTRAINT_SUPPORT`, leading to a `KeyError` during `process_parsed_constraint`.

### Solution

Completely ignore "custom" constraints when checking for adapter-specific support in `process_parsed_constraint`.

### Checklist

### DIFF
--- a/dbt/adapters/base/impl.py
+++ b/dbt/adapters/base/impl.py
@@ -1535,6 +1535,9 @@ class BaseAdapter(metaclass=AdapterMeta):
         parsed_constraint: Union[ColumnLevelConstraint, ModelLevelConstraint],
         render_func,
     ) -> Optional[str]:
+        # skip checking enforcement if this is a 'custom' constraint
+        if parsed_constraint.type == ConstraintType.custom:
+            return render_func(parsed_constraint)
         if (
             parsed_constraint.warn_unsupported
             and cls.CONSTRAINT_SUPPORT[parsed_constraint.type] == ConstraintSupport.NOT_SUPPORTED
--- a/.changes/unreleased/Fixes-20240215-141545.yaml
+++ b/.changes/unreleased/Fixes-20240215-141545.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Ignore adapter-level support warnings for 'custom' constraints
+time: 2024-02-15T14:15:45.764145+01:00
+custom:
+  Author: jtcohen6
+  Issue: "90"
--- a/tests/unit/test_base_adapter.py
+++ b/tests/unit/test_base_adapter.py
@@ -40,6 +40,7 @@ class TestBaseAdapterConstraintRendering:
             ["column_name integer references other_table (c1)"],
         ),
         ([{"type": "check"}, {"type": "unique"}], ["column_name integer unique"]),
+        ([{"type": "custom", "expression": "-- noop"}], ["column_name integer -- noop"]),
     ]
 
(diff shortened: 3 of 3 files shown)


dbt/adapters/base/impl.py
```diff
@@ -1535,6 +1535,9 @@ def process_parsed_constraint(
         parsed_constraint: Union[ColumnLevelConstraint, ModelLevelConstraint],
         render_func,
     ) -> Optional[str]:
+        # skip checking enforcement if this is a 'custom' constraint
```
mikealfare, line 1538: Put another way, are we saying that custom constraints are always supported and always enforced? Or do we still need to check that?

mikealfare: I just read the issue. So it seems as though custom constraints are always supported and _never_ enforced. But because of this, we wouldn't need to warn on not enforced as the user knows what they're signing up for. This makes sense then.

mikealfare (approved):

PR #93, opened 2024-02-16
Create separate base classes for shared tests

mikealfare:
blocks https://github.com/dbt-labs/dbt-postgres/pull/18

### Problem

There are some tests in the "adapter zone" which do not have base classes. When these tests are imported, they run automatically due to how `pytest` works. This makes it impossible to configure the test if changes are required for a specific platform.

### Solution

- Update the test class name to not start with "Test" (which is what triggers `pytest` to execute it)
- Create a new test class with the with same name that inherits the base class and does nothing for backwards compatibility

For example, this:
```python
class TestFoo:
    def test_bar(self):
        assert baz == "baz"
```
becomes:
```python
class BaseFoo:
    def test_bar(self):
        assert baz == "baz"


class TestFoo(BaseFoo):
    pass
```

### Checklist


dbt/tests/adapter/caching/test_caching.py
```diff
@@ -116,3 +116,7 @@ class TestCachingUppercaseModel(BaseCachingUppercaseModel):
 
 class TestCachingSelectedSchemaOnly(BaseCachingSelectedSchemaOnly):
     pass
+
+
+class TestNoPopulateCache(BaseNoPopulateCache):
```
MichelleArk, line 121: thinking out loud: do we actually need any of the `Test`-prefixed integration tests in dbt-adapters? as in, will they actually be run against a concrete adapter as part of dbt-adapters development? If not, I wonder if we should just remove them and only define `Base` integration tests in this repo.

mikealfare: I agree that we don't need them. In cases where they already exist, I kept them in case folks were importing them as-is (causing them to run as-is). But we don't need to add new ones. I did that because that was the pattern that we had. I can remove the net new ones if that's preferable.

dbt/tests/adapter/concurrency/test_concurrency.py
```diff
@@ -303,8 +303,6 @@ def models(self):
             "skip.sql": models__skip_sql,
         }
 
-
-class TestConcurenncy(BaseConcurrency):
```
VersusFacit, line 307: Did you cut this one off.

dbt/tests/adapter/dbt_debug/test_dbt_debug.py
```diff
@@ -85,11 +85,11 @@ def test_empty_target(self, project):
         self.assertGotValue(re.compile(r"\s+output 'none_target'"), "misconfigured")
 
 
-class TestDebugProfileVariablePostgres(BaseDebugProfileVariable):
+class TestDebugPostgres(BaseDebugPostgres):
```
VersusFacit, line 88: Whats up with this rename? I'm missing something

VersusFacit (approved):
Couple nits

dbt/tests/adapter/concurrency/test_concurrency.py
mikealfare: Yeah, I effectively moved the test method from the test class up into the base class.

dbt/tests/adapter/dbt_debug/test_dbt_debug.py
mikealfare: The diffs on this PR are confusing.

`BaseDebugProfileVariable` has no test methods in it, which means that `TestDebugProfileVariablePostgres` did nothing. So I removed it.

Also, I broke out `TestDebugPostgres` into a base class and a test class, which got caught up in this.

So it went from this:
```python
class TestDebugPostgres(BaseDebug):
    def test_ok(self, project):
        ...


class TestDebugProfileVariablePostgres(BaseDebugProfileVariable):
    pass  # this class contains no test methods


class TestDebugInvalidProjectPostgres(BaseDebug):
    def test_empty_project(self, project):
        ...
```

to this:
```python
class BaseDebugPostgres(BaseDebug):
    def test_ok(self, project):
        ...


# break base class out from test class
class TestDebugPostgres(BaseDebugPostgres):
    pass


# delete empty test class TestDebugProfileVariablePostgres


class BaseDebugInvalidProjectPostgres(BaseDebug):
    def test_empty_project(self, project):
        ...


# break base class out from test class
class TestDebugInvalidProjectPostgres(BaseDebugInvalidProjectPostgres):
    pass
```

mikealfare:
> Couple nits

Good questions. I spent a good 15-30 minutes figuring out why the rename looked like it did. It turns out to be correct, but very misleading. So now we have the why captured in this PR.

dbt/tests/adapter/dbt_debug/test_dbt_debug.py
VersusFacit: Yep that makes sense. Just wanted to double check.

PR #94, opened 2024-02-19
Readd removed logic for deferral fail-over

jtcohen6:
The logic added in #60 (to support https://github.com/dbt-labs/dbt-core/pull/9199) was removed in #55.

I'm not sure if there's a good way to test this within `dbt-adapters` by itself...

### Problem

Logic added to failover to `defer_relation` if `this` is not present in the target schema. (I realize that the logic I added in #60 wasn't quite right, either!)

### Solution

Check to see if `this` exists (cache lookup), then fail over to using `defer_relation` instead for column type detection.

### Checklist

### DIFF
--- a/dbt/include/global_project/macros/unit_test_sql/get_fixture_sql.sql
+++ b/dbt/include/global_project/macros/unit_test_sql/get_fixture_sql.sql
@@ -3,7 +3,7 @@
 {% set default_row = {} %}
 
 {%- if not column_name_to_data_types -%}
-{%-   set columns_in_relation = adapter.get_columns_in_relation(this) -%}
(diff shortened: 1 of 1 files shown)


dbt/include/global_project/macros/unit_test_sql/get_fixture_sql.sql
```diff
@@ -44,7 +44,7 @@ union all
 {% macro get_expected_sql(rows, column_name_to_data_types) %}
 
 {%- if (rows | length) == 0 -%}
-    select * FROM dbt_internal_unit_test_actual
+    select * from dbt_internal_unit_test_actual
```
jtcohen6, line 47: sorry couldn't help myself

mikealfare (approved):

PR #118, opened 2024-02-29
Move the entirety of dbt-tests-adapter into the dbt-tests-adapter directory

mikealfare:
### Problem

`hatch` makes a lot of assumptions about where files are relative to each other. In particular, it prefers one package per repo.

### Solution

Instead of nesting the namespace packages together, move the `dbt-tests-adapter` package into its own directory, isolated from everything else.

### Checklist


.github/workflows/release.yml
```diff
@@ -72,20 +72,55 @@ defaults:
...
+        run: |
+          echo package        : ${{ inputs.package }}
+          echo working-dir    : ${{ steps.release-inputs.outputs.working-dir }}
+          echo run-unit-tests : ${{ steps.release-inputs.outputs.run-unit-tests }}
+          echo archive-name   : ${{ steps.release-inputs.outputs.archive-name }}
+
   bump-version-generate-changelog:
     name: "Bump package version, Generate changelog"
-    uses: dbt-labs/dbt-adapters/.github/workflows/release_prep_hatch.yml@main
+    uses: dbt-labs/dbt-adapters/.github/workflows/release_prep_hatch.yml@isolate-dbt-tests-adapter
```
mikealfare, line 109: Point to `main` before merging

.github/workflows/release.yml
```diff
@@ -108,38 +143,30 @@ jobs:
...
+          archive-name: ${{ needs.release-inputs.outputs.archive-name }}
 
   github-release:
     name: "GitHub Release"
     # ToDo: update GH release to handle adding dbt-tests-adapter and dbt-adapter assets to the same release
     if: ${{ !failure() && !cancelled() && inputs.package == 'dbt-adapters' }}
-    needs: [build-and-test, bump-version-generate-changelog]
-    uses: dbt-labs/dbt-adapters/.github/workflows/github-release.yml@main
+    needs: [release-inputs, build-and-test, bump-version-generate-changelog]
+    uses: dbt-labs/dbt-adapters/.github/workflows/github-release.yml@isolate-dbt-tests-adapter
```
mikealfare, line 158: Point to `main` before merging

dbt/adapters/__about__.py
```diff
@@ -1 +1 @@
-version = "1.8.0a2"
+version = "1.0.0a2"
```
mikealfare, line 1: Revert version back to the correct one. This must have been inadvertently bumped to the `dbt-tests-adapter` version.

pyproject.toml
```diff
@@ -142,6 +142,7 @@ files = [
 ]
 exclude = [
     "dbt/adapters/events/adapter_types_pb2.py",
+    "dbt-tests-adapter/dbt/__init__.py",  # overlaps with `dbt/__init__.py` as expected for namespaces
```
mikealfare, line 145: Since we split the namespace package, this shows up twice for `mypy`, which means the same path goes to two files. This is expected and works itself out once the namespace packages are combined.

dbt-tests-adapter/pyproject.toml
mikealfare, line 1: Combining everything in the same directory means nothing needs to be renamed. The version file got moved up because aligns with the `dbt-adapters` `pyproject.toml` file.

.github/workflows/release.yml
```diff
@@ -72,20 +72,55 @@ defaults:
     shell: bash
 
 jobs:
+  release-inputs:
```
mikealfare, line 75: Determine all package-specific parameters up front in one place. This allows for one action per "function" instead of two with conditionals on the package, which makes job dependencies easier.

.github/workflows/release.yml
```diff
@@ -108,38 +143,30 @@ jobs:
       - name: "Setup `hatch`"
         uses: dbt-labs/dbt-adapters/.github/actions/setup-hatch@main
 
-      - name: "Build `dbt-adapters`"
-        if: ${{ inputs.package == 'dbt-adapters' }}
+      - name: "Build ${{ inputs.package }}"
```
mikealfare, line 146: Combines the two build steps into one.

.github/workflows/release_prep_hatch.yml
```diff
@@ -53,6 +53,9 @@ on:
         type: string
         required: false
         default: ''
+      run-unit-tests:
```
mikealfare, line 56: `dbt-tests-adapter` has no tests, so unit tests also need to be optional.

.github/workflows/release_prep_hatch.yml
```diff
@@ -433,7 +447,7 @@ jobs:
           python-version: ${{ env.PYTHON_TARGET_VERSION }}
 
       - name: Run tests
-        run: hatch run test:integration
+        run: hatch run integration-tests:all
```
mikealfare, line 450: This never gets run because there are no integration tests in this repo, but we should probably fix the command anyway.

dbt-tests-adapter/dbt/__init__.py
mikealfare, line 1: We need to create a copy of this file now since the namespace package is being split out.

.github/actions/build-hatch/action.yml
```diff
@@ -35,8 +32,6 @@ runs:
     - name: Upload artifacts
       uses: actions/upload-artifact@v3
       with:
-        name: ${{ inputs.archive-name}}
-        path: |
-          ${{ inputs.changelog-path }}
```
mikealfare, line 40: We don't need to push the changelogs up to the archive

.github/actions/publish-pypi/action.yml
```diff
@@ -17,7 +17,11 @@ runs:
       uses: actions/download-artifact@v3
       with:
         name: ${{ inputs.archive-name }}
-        path: .
```
mikealfare, line 20: We should leave the distributions in the expected directory. Also added a log statement to see what's going on in this step.

colin-rogers-dbt (approved):

PR #127, opened 2024-03-11
Introduce TableLastModifiedMetadataBatch and implement BaseAdapter.calculate_freshness_from_metadata_batch

MichelleArk:
resolves https://github.com/dbt-labs/dbt-adapters/issues/138

### Problem

No base abstraction / implementation exists for calculating metadata-based freshness in batch

### Solution

1. Introduce `TableLastModifiedMetadataBatch` capability, for adapter implementations to indicate whether a batch version of metadata-based freshness calculation is available
2.Implement `BaseAdapter.calculate_freshness_from_metadata_batch`
  * groups sources into information schemas and runs 1 query / information schema under the hood, returning a response per query and a mapping of relations -> freshness results
  * I also considered updating `calculate_freshness_from_metadata` but the interfaces were different enough that it would have been difficult to do in a backward-compatible way. However, `calculate_freshness_from_metadata` was rewritten in terms of `calculate_freshness_from_metadata_batch` so there are some wins in terms of code reuse here. 
3. Added `needs_conn` to `BaseAdapter.execute_macro` for callers to signal whether a macro will require an open connection or not. By default, this is false (preserving existing behaviour), but `calculate_freshness_from_metadata_batch` sets it to true.

### Checklist

- [ ] This PR includes tests, or tests are not required/relevant for this PR. Tests in supported adapters
   - [ ] dbt-snowflake: https://github.com/dbt-labs/dbt-snowflake/pull/928
   - [ ] dbt-redshift: https://github.com/dbt-labs/dbt-redshift/pull/744


dbt/adapters/base/impl.py
```diff
@@ -1285,46 +1287,84 @@ def calculate_freshness(
         }
         return adapter_response, freshness
 
+    def calculate_freshness_from_metadata_batch(
+        self,
+        sources: List[BaseRelation],
+        macro_resolver: Optional[MacroResolverProtocol] = None,
+    ) -> Tuple[Optional[AdapterResponse], Dict[BaseRelation, FreshnessResponse]]:
+        # Track schema, identifiers of sources for lookup from batch query
```
MichelleArk, line 1295: There's probably a better way to do this. I'm trying to create a mapping ahead of time to use to map the individual rows from the result table back to a source relation, since the result from the warehouse is not guaranteed to be in any particular order (& I've validated that the order is not preserved, even coincidentally).

Main issues I see with this current approach are:
  * source and identifier are technically optional which this (as implemented) doesn't handle
  * there is probably something more reusable that already exists for this purpose, e.g. BaseRelation.matches

I thought about using `BaseRelation.matches` with the per-row `schema` and `identifier` values we get back from the batch query, but my concern there is performance: we'd need to iterate over all inputs to find a matching one, for each result row. Additionally, the casing we get back from the warehouse is pretty out of our control. For example, it seems like snowflake always returns these capitalized, which probably won't match count as an `exact` match using `BaseRelation.matches`. We'd probably need to extend `matches` to optionally return `True` if an approximate match is found. 

Another option is to also expect adapters to implement the `get_relation_last_modified` to include `source.render()` as an `id`, and use that to map from the table result to the. That would require updates to the macro implementations, but only for the adapters that support the MetadataBatch capability, and wouldn't break previous behaviour if such a key was not provided. This is the approach I'm leaning toward currently.

cc @mikealfare -- open to any thoughts or suggestions here!

dbt/adapters/base/impl.py
```diff
@@ -1285,46 +1287,92 @@ def calculate_freshness(
...
-        try:
-            row = table[0]
-            last_modified_val = get_column_value_uncased("last_modified", row)
-            snapshotted_at_val = get_column_value_uncased("snapshotted_at", row)
-        except Exception:
-            raise MacroResultError(GET_RELATION_LAST_MODIFIED_MACRO_NAME, table)
+        return adapter_response, list(freshness_responses.values())[0]
 
-        if last_modified_val is None:
+    def _create_freshness_response(
```
MichelleArk, line 1358: I chose to carve out this helper out of `calculate_freshness_from_metadata_batch` as I can imagine adapter implementations that need to overwrite `calculate_freshness_from_metadata_batch` can still make use of this method for consistent FreshnessResponse creation. The other pieces of `calculate_freshness_from_metadata_batch` were pretty specific to the approach taken in the base

dbt/adapters/base/impl.py
```diff
@@ -1285,46 +1287,92 @@ def calculate_freshness(
...
+                GET_RELATION_LAST_MODIFIED_MACRO_NAME,
+                kwargs={
+                    "information_schema": information_schema,
+                    "relations": sources_for_information_schema,
+                },
+                macro_resolver=macro_resolver,
+            )
+            adapter_response, table = result.response, result.table  # type: ignore[attr-defined]
+
+            for row in table:
```
mikealfare, line 1328: Have you considered putting the contents of this `for` loop in a helper function? There are a few reasons I'd consider it here. This is a nested for loop and there's a decent amount of processing happening within the loop. Separating the logic may make it easier to read. I would also consider doing the relation mapping outside of the loop. For example:

```python
def calculate_freshness_from_metadata_batch(...):

    ...

    raw_freshness_responses: Dict[Tuple[str, str], FreshnessResponse] = {}
    for (
        information_schema,
        sources_for_information_schema,
    ) in information_schema_to_metadata_sources.items():
        result = self.execute_macro(
            GET_RELATION_LAST_MODIFIED_MACRO_NAME,
            kwargs={
                "information_schema": information_schema,
                "relations": sources_for_information_schema,
            },
            macro_resolver=macro_resolver,
        )
        adapter_response, table = result.response, result.table  # type: ignore[attr-defined]
        raw_freshness_responses.update(
            self._parse_freshness_row(row)
            for row in table
        )
    
    # handle the schema/identifier -> BaseRelation here
    
    return adapter_response, freshness_response

...

def _parse_freshness_row(self, row: agate.Row) -> Tuple[Any, FreshnessResponse]:
    try:
        last_modified_val = get_column_value_uncased("last_modified", row)
        snapshotted_at_val = get_column_value_uncased("snapshotted_at", row)
        identifier = get_column_value_uncased("identifier", row)
        schema = get_column_value_uncased("schema", row)
    except Exception:
        raise MacroResultError(GET_RELATION_LAST_MODIFIED_MACRO_NAME, table)
    
    freshness_response = self._create_freshness_response(
        last_modified_val, snapshotted_at_val
    )
    raw_relation = schema.lower(), identifier.lower()
    return raw_relation, freshness_response
```

mikealfare (commented):
I left a few comments/questions, some might even be useful.

dbt/adapters/base/impl.py
```diff
@@ -1285,46 +1287,92 @@ def calculate_freshness(
...
+        ) in information_schema_to_metadata_sources.items():
+            result = self.execute_macro(
+                GET_RELATION_LAST_MODIFIED_MACRO_NAME,
+                kwargs={
+                    "information_schema": information_schema,
+                    "relations": sources_for_information_schema,
+                },
+                macro_resolver=macro_resolver,
+            )
+            adapter_response, table = result.response, result.table  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
MichelleArk: I've ended up returning all of them, so that the caller can determine how/whether it should handle partial failures here.

dbt/adapters/base/impl.py
MichelleArk: Incorporated this suggestion, thank you!

dbt/adapters/base/impl.py
```diff
@@ -1285,46 +1287,92 @@ def calculate_freshness(
...
+        # Track schema, identifiers of sources for lookup from batch query
+        schema_identifier_to_source = {
+            (
+                source.path.get_lowered_part(ComponentName.Schema),
+                source.path.get_lowered_part(ComponentName.Identifier),
+            ): source
+            for source in sources
+        }
+
+        # Group metadata sources by information schema -- one query per information schema will be necessary
```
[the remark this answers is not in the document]
MichelleArk: Ended up finding `self._get_catalog_relations_by_info_schema` -- which did exactly what had been implemented here previously so used that instead. + renamed 👍

dbt/adapters/base/impl.py
```diff
@@ -1285,46 +1287,92 @@ def calculate_freshness(
         }
         return adapter_response, freshness
 
+    def calculate_freshness_from_metadata_batch(
+        self,
+        sources: List[BaseRelation],
+        macro_resolver: Optional[MacroResolverProtocol] = None,
+    ) -> Tuple[Optional[AdapterResponse], Dict[BaseRelation, FreshnessResponse]]:
+        # Track schema, identifiers of sources for lookup from batch query
+        schema_identifier_to_source = {
```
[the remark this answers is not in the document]
MichelleArk: [It does!](https://github.com/dbt-labs/dbt-adapters/blob/8cddfd35a842ec2ee279b021c5e2ef7a70c11e9d/dbt/adapters/contracts/relation.py#L137-L139)

This means that the key can contain `None` values, which is fine and would be handled gracefully as below. I'll add some of this nuance (all sources from the provided `sources` input that are possible to obtain from a batch result will be returned)

dbt/adapters/base/impl.py
```diff
@@ -1100,6 +1100,12 @@ def execute_macro(
 
         macro_function = CallableMacroGenerator(macro, macro_context)
 
+        # A connection may or may not be required for a given macro execution
```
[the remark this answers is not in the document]
MichelleArk: sounds good! will plumb this through the `calculate_freshness_from_metadata_batch` method as well 👍

dbt/adapters/base/impl.py
```diff
@@ -1297,48 +1306,110 @@ def calculate_freshness(
...
+        adapter_responses, freshness_responses = self.calculate_freshness_from_metadata_batch(
+            sources=[source],
             macro_resolver=macro_resolver,
         )
-        adapter_response, table = result.response, result.table  # type: ignore[attr-defined]
-
-        try:
-            from dbt_common.clients.agate_helper import get_column_value_uncased
+        adapter_response = adapter_responses[0] if adapter_responses else None
+        return adapter_response, list(freshness_responses.values())[0]
```
colin-rogers-dbt, line 1375: probably missing something but instead of coercing to a list and getting the first value could we pass the `source` as the key to retrieve the value?
i.e 
```suggestion
        return adapter_response, freshness_responses[source]
```

MichelleArk: oh good call! I think we could, will give it a shot

colin-rogers-dbt (approved):

PR #135, opened 2024-03-24
Debug log when `type_code` fails to convert to a `data_type`

### LINKED ISSUE #134: [Feature] Debug log when `type_code` fails to convert to a `data_type`

### Describe the feature

https://github.com/dbt-labs/dbt-postgres/issues/38

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

dbeatty10:
resolves #134

### Problem

For any given adapter, it's possible that dbt doesn't recognize a data type specified in a [dbt model contract](https://docs.getdbt.com/docs/collaborate/govern/model-contracts#how-to-define-a-contract). An example is https://github.com/dbt-labs/dbt-postgres/issues/54

In those cases, we want there to be a mechanism for the dbt adapter to emit a log message of what's going on.

### Solution

Emit a debug-level log message any time a data type is not recognized within dbt-postgres. That way, users can at least see in the logs what is going on.

Log messages will look like the following:

> The `type_code` 790 was not recognized

An example of how to utilize this log message is within https://github.com/dbt-labs/dbt-postgres/pull/39/

### Checklist

- [x] This PR includes tests
- [x] This PR has already received feedback and approval from Product or DX


dbt/adapters/events/types.py
```diff
@@ -422,3 +422,12 @@ def message(self) -> str:
             "be ignored. Set 'warn_unsupported: false' on this constraint to ignore this warning."
         )
         return line_wrap_message(warning_tag(msg))
+
+
+class TypeCodeNotFound(DebugLevel):
+    def code(self) -> str:
+        return "E050"
+
+    def message(self) -> str:
```
mikealfare, line 431: I noticed on the longer messages (e.g. the two above this) we use `line_wrap_message`. Should we do that here as well? The message also appears to be longer than the linter would allow, and should probably be two or three lines.

mikealfare (dismissed):
I only noticed one thing, which is aesthetic. I'm not super familiar with `protobuf`, but this change looks consistent with the others.

dbt/adapters/events/types.py
```diff
@@ -422,3 +422,12 @@ def message(self) -> str:
...
+        msg = f"The `type_code` {self.type_code} was not recognized, which may affect error messages for enforced contracts that fail as well as `Column.data_type` values returned by `get_column_schema_from_query`"
+        return msg
```
dbeatty10, line 433:
```suggestion
        msg = (
            f"The `type_code` {self.type_code} was not recognized, which may affect error "
            "messages for enforced contracts that fail as well as `Column.data_type` values "
            "returned by `get_column_schema_from_query`"
        )
        return line_wrap_message(warning_tag(msg))
```

dbt/adapters/events/types.py
dbeatty10: Thx for the suggestion! Added in e398d8b676e38c4f354ce4dd2ff5066e6f493d79

mikealfare (approved):

PR #141, opened 2024-03-29
Change AdapterRegistered event to dynamic log level

emmyoop:
resolves https://dbtlabs.atlassian.net/browse/MNTL-191

### Problem

The `AdapterRegistered` event is currently INFO level but some would prefer it to be DEBUG level.

### Solution

 Let the log level of `AdapterRegistered` be dynamic but default it to INFO to preserve current behavior and prevent breaking changes.

### Checklist


mikealfare (approved):

mikealfare (approved):

dbt/adapters/factory.py
```diff
@@ -186,8 +198,8 @@ def get_adapter_constraint_support(self, name: Optional[str]) -> List[str]:
 FACTORY: AdapterContainer = AdapterContainer()
 
 
-def register_adapter(config: AdapterRequiredConfig, mp_context: SpawnContext) -> None:
-    FACTORY.register_adapter(config, mp_context)
+def register_adapter(config: AdapterRequiredConfig, mp_context: SpawnContext, AdapterRegistered_log_level: EventLevel) -> None:
```
mikealfare, line 201: Should `AdapterRegistered_log_level` be optional to preserve the signature? Also, why is the name initially camel-case? Couldn't it just be `log_level`?

emmyoop: Yes it should be optional.  I've been testing this against `dbt-mantle` so should probably have marked the PR as draft.

The event is `AdapterRegistered` and the log_level we set is specific to that one event.  Calling it just `log_level` felt like it would be confusing.  Happy to rename if you come up with something better.

mikealfare: Since the function is already called `register_adapter`, I don't know that you need to be specific for an argument that's local to that function. But if that's desired, I would still stick to snake case as that's a python standard, so `adapter_registered_log_level`.

colin-rogers-dbt (changes requested):
Blocking on internal conversation unrelated to the code

colin-rogers-dbt (approved):

PR #142, opened 2024-04-01
download archive to the expected directory

mikealfare:
### Problem

We downloaded the archive to the working directory, but we expect that the artifacts are in `./dist/`.

### Solution

Update the download step to reference `./dist/`

### Checklist

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -211,7 +211,7 @@ jobs:
         uses: actions/download-artifact@v3
         with:
           name: ${{ inputs.archive_name }}
-          path: .
+          path: dist/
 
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog-existence.yml
```diff
@@ -19,7 +19,6 @@ name: Check Changelog Entry
 on:
   pull_request_target:
     types: [opened, reopened, labeled, unlabeled, synchronize]
-    paths-ignore: ['.changes/**', '.github/**', 'tests/**', 'third-party-stubs/**', '**.md', '**.yml']
```
mikealfare, line 22: Apparently if all changes are in skipped paths and the test is required, it never runs, hence never passes.

VersusFacit (approved):

PR #161, opened 2024-04-10
Update dependabot config to cover GHA

mikealfare:
### Problem

We have inconsistent and under-configured dependabot config files.

### Solution

Consistify them.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-postgres/blob/main/CONTRIBUTING.md) and understand what's expected of me

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,42 @@
 version: 2
 updates:
-  # python dependencies
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
mikealfare, line 1: We need one entry in this config per directory, until we can supply a list of directories (in beta currently). So this is just `pip` and `GHA`, but for many directories.

emmyoop (approved):

PR #165, opened 2024-04-12
[fix] use cast macro in default__get_empty_schema_sql

MichelleArk:
resolves https://github.com/dbt-labs/dbt-adapters/issues/166
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#

### Problem

adapter-specific `cast` macro not being used in `default__get_empty_schema_sql` which leads to adapter-specific cases being mishandled and erroring. 

### Solution

Use the newly available adapter-specific cast macro! By default, this is just a refactor since the default implementation just does a `cast <field> as <type>`. However, adapters that override the cast macro (e.g. snowflake) would benefit from broader type support in constraint enforcement if the base implementation of `get_empty_schema_sql` made use of the adapter-specific override.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-adapter/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development, and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
   - [x] Snowflake tests: https://github.com/dbt-labs/dbt-snowflake/pull/969
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX

:tophat:

```sql
-- geography.sql
SELECT 
TO_GEOMETRY('POINT(1820.12 890.56)') as geometry_field, 
TO_GEOGRAPHY('POINT(-122.35 37.55)') as geography_field
```

✅ correct & passing contract: 
```
❯ dbt run --select geography
02:05:28  Running dbt...
02:05:28  target not specified in profile 'snowflake', using 'default'
02:05:29  Found 14 models, 6 seeds, 20 data tests, 6 sources, 13 metrics, 709 macros, 6 semantic models, 3 unit tests
02:05:29  
02:05:31  Concurrency: 4 threads (target='default')
02:05:31  
02:05:31  1 of 1 START sql view model dbt_marky.geography ................................ [RUN]
02:05:31  1 of 1 OK created sql view model dbt_marky.geography ........................... [SUCCESS 1 in 0.72s]
02:05:31  
02:05:31  Finished running 1 view model in 0 hours 0 minutes and 1.84 seconds (1.84s).
02:05:31  
02:05:31  Completed successfully
02:05:31  
[...]

PR #173, opened 2024-04-13
Add tests for cross-database `cast` macro

### LINKED ISSUE #84: [CT-1229] [Feature] Cross-database `cast` macro

### Describe the feature

Per conversation with @jtcohen6 we are interested in creating a `cast()` macro meant to behave like the `CAST` operator described in the SQL standard.

Usage likely to be something like:
```sql
{{ cast("some expression", type_string()) }}
```

Which some databases might render as:
```sql
cast(some expression as TEXT)
```

- Not providing this abstraction at all
- Providing a GitHub gist containing this abstraction and end users can copy-paste into their own dbt packages/projects

### Who will this benefit?

Like all cross-database macros, we anticipate this being most relevant to dbt package maintainers.

It would also be useful in implementing this:
- https://github.com/dbt-labs/dbt-core/issues/5445#issuecomment-1210875104

Yep!

### Anything else?

N/A

dbeatty10:
resolves #84
docs: https://github.com/dbt-labs/docs.getdbt.com/pull/5278

### Problem

Having a cross-database `cast` macro has been suggested as a way to address a variety of things including, but not limited to:
- https://github.com/dbt-labs/dbt-adapters/pull/55
- https://github.com/dbt-labs/dbt-adapters/pull/165
- https://github.com/dbt-labs/dbt-snowflake/pull/969
- https://github.com/dbt-labs/dbt-core/issues/5445#issuecomment-1210875104

### Solution

We actually already added it in https://github.com/dbt-labs/dbt-adapters/pull/55/commits/5a50be73163514dd7055af046c5223b092ee9b61 within https://github.com/dbt-labs/dbt-adapters/pull/55 !

But this PR adds some tests (just duplicated from the ones for `safe_cast`) and a changelog entry.

### Checklist

- [x] This PR includes tests
- [x] This PR has already received feedback and approval from Product or DX


dbt-tests-adapter/dbt/tests/adapter/utils/test_cast.py
```diff
@@ -0,0 +1,25 @@
...
+            "test_cast.sql": self.interpolate_macro_namespace(
+                self.interpolate_macro_namespace(
+                    fixture_cast.models__test_cast_sql, "cast"
+                ),
+                "type_string",
+            ),
+        }
+
+
+class TestCast(BaseCast):
```
mikealfare, line 24: I don't think we need this class, do we? This would be created in the calling test suite. If you import it, there's no way to tell `pytest` not to run it. We used to add it when `dbt-postgres` was in `dbt-core` because we could just run this whole suite to run the tests for `dbt-postgres`, but that's no longer the case.

dbeatty10: My preference is for consistency.

And then if we want to take some kind of action, then this would be included along with that.

But I'd be happy to defer to your judgement if you prefer I peel it out.

dbeatty10: I just saw that you approved and gave me the option here:

> Your choice as to removing the test implementation or keeping it.

Since CI is passing, I'm going to merge now. But just let me know if you want me to peel it out in a follow-up PR!

PR #174, opened 2024-04-16
Add and configure `pre-commit`

mikealfare:
### Problem / Solution

We need to setup `pre-commit`. This PR also addresses the resulting violations.

```[tasklist]
### Acceptance Criteria
- [ ] `pre-commit` is configured with built-in hooks that are used in other adapter repos
- [ ] `pre-commit` is configured to run `black`, `flake8`, and `mypy` using the configurations in `pyproject.toml`
- [ ] `pre-commit` can be installed via `hatch`
- [ ] GHA workflows that run code quality checks, or use `pre-commit` in other ways (e.g. to tidy up the changelog on release) are updated so they work properly
```

```[tasklist]
### Tasks
- [x] Update `.pre-commit-config.yaml` to use the `local` repo so that the config in `pyproject.toml` gets used everywhere
- [x] Add a command `hatch run dev` that will instantiate the dev environment and install `pre-commit` hooks
- [x] Combine `default`, `lint`, and `typecheck` `hatch` environments into the `hatch` `default` environment
- [x] Install `pre-commit` as part of the CI envrionment setup (`setup-hatch`) since it's needed for most CI anyway
- [x] Update `code-quality` workflow to run all code quality at once instead of as separate steps
- [x] Exclude `dbt/adapters/events/adapter_types_pb2.py` from linting; this is a meta-coded file
- [x] Exclude `dbt/adapters/protocol.py` from `flake8` since `flake8` does not correctly handle the syntax for `Protocol` classes
- [x] Run `pre-commit` (by running `hatch run code-quality`) to trigger updates so that CI passes
```

### Checklist


pyproject.toml
```diff
@@ -54,32 +54,21 @@ include = ["dbt/adapters", "dbt/include", "dbt/__init__.py"]
 [tool.hatch.envs.default]
 dependencies = [
     "dbt_common @ git+https://github.com/dbt-labs/dbt-common.git",
-]
-
-[tool.hatch.envs.lint]
```
mikealfare, line 59: There's no point in separate lint and typecheck environments. We want the default to be able to run precommit, which requires all of this in default.

.github/workflows/code-quality.yml
```diff
@@ -33,8 +33,5 @@ jobs:
       - name: Setup `hatch`
         uses: ./.github/actions/setup-hatch
 
-      - name: Run linters
-        run: hatch run lint:all
-
-      - name: Run typechecks
-        run: hatch run typecheck:all
+      - name: Run code quality
+        run: hatch run code-quality
```
VersusFacit, line 37: Where does this point to? I spent way too long trying to discover where.

dbt/adapters/base/impl.py
```diff
@@ -1328,14 +1328,16 @@ def calculate_freshness_from_metadata_batch(
         # Track schema, identifiers of sources for lookup from batch query
         schema_identifier_to_source = {
             (
-                source.path.get_lowered_part(ComponentName.Schema),
-                source.path.get_lowered_part(ComponentName.Identifier),
+                source.path.get_lowered_part(ComponentName.Schema),  # type: ignore
```
mikealfare, line 1331: I'm not sure why `mypy` doesn't like this, but it doesn't. It think it's getting a `str` and wants a `ComponentName`, of which this is clearly an instance. Maybe it has something to do with the fact that we rolled our own `StrEnum`.

pyproject.toml
mikealfare, line 1: All of the config that's missing here should be in `.pre-commit-config.yaml` now, unless it was found to be unnecessary (not actually used/didn't have an effect).

.pre-commit-config.yaml
mikealfare, line 1: There did not seem to be much benefit to splitting out between manual runs and CI runs, so I collapsed the configs in those scenarios.

colin-rogers-dbt (approved):

dbeatty10:
We should probably include install and usage instructions, so I opened this issue:
- https://github.com/dbt-labs/dbt-adapters/issues/180

Issue #124, opened 2024-03-09
[Bug] --empty doesn't work with relation that are already named and might conflict


github-christophe-oudar:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

This a follow up from [my comment](https://github.com/dbt-labs/dbt-core/pull/8971#issuecomment-1881062823) on the original PR.

If any is using code like `FROM {{ source('schema', 'table') }} my_table_alias` while using the --empty alias.
The resulting SQL is the following: ```FROM  (select * from `db`.`schema`.`table` where false limit 0) _dbt_limit_subq my_table_alias```

The problem is linked to the alias that's already set up here: https://github.com/dbt-labs/dbt-adapters/blob/2a99e36a7479ec23f40456e89377419aae7388e7/dbt/adapters/base/relation.py#L213-L214
which prevents from aliasing by ourselves. Also if there are multiple references, I guess the tables will all be aliased to `_dbt_limit_subq` resulting in an error.

### Expected Behavior

The simplest expected would be that it results in ```FROM  (select * from `db`.`schema`.`table` where false limit 0) my_table_alias``` (practically that no table name is suffixed automatically). However if the user didn't name its table, it will fail... which is another problem. Yet we could look into alternatives:
- finding a way to detect the table is suffixed
     - it looks like complex based on how dbt works
- enabling setting the alias through the kwargs from the source/ref macros
     - it might be seems reasonable but requires modifications on the end users

### Steps To Reproduce

- Create a simple model `my_model.sql` with an SQL like:
`SELECT * FROM {{ source('schema', 'table') }} my_table_alias`
- run dbt command `dbt run -s my_model --empty`

### Environment

```markdown
- OS: MacOS 13.2.1
- Python: 3.11
- dbt-adapter: 1.8.0b1
```

dwreeves:
> set no alias (but then I think some DBs requires an alias and then it would fail for them with default setup)

Yes, some do (e.g. Postgres) but some don't (e.g. Snowflake).

This could be an attribute of the dialect's adapter, e.g. `requires_aliases_for_subqueries: bool`.

For those that do not required, aliasing shouldn't be used at all. Then the problem is solved arbitrarily, no other fixes required.

[...]

PR #179, opened 2024-04-18
Make aliasing subquery configurable for render_limited

colin-rogers-dbt:
partially addresses #124 

Currently only a subset of adapters require adding an alias to the subquery generated by `render_limited` (i.e. when a user passes the `--empty` flag). Currently if a user has already aliased a referenced model or source we duplicate the aliasing and generate invalid SQL. 

Unfortunately at the moment we have no way of knowing if a user has already aliased the reference so quickest mitigation is to simply not add the alias if the platform does not require it. This should allow us to fix this bug for many platforms (snowflake, bigquery) but doesn't fix this for others (postgres, mysql, oracle). 

dbt-snowflake update: https://github.com/dbt-labs/dbt-snowflake/pull/980

### Checklist


dbt/adapters/base/relation.py
```diff
@@ -205,14 +206,22 @@ def render(self) -> str:
         # if there is nothing set, this will return the empty string.
         return ".".join(part for _, part in self._render_iterator() if part is not None)
 
+    def _render_limited_alias(self) -> str:
+        """ Some databases require an alias for subqueries (postgres, mysql) for all others we want to avoid adding
+        an alias as it has the potential to introduce issues with the query if the user also defines an alias.
+        """
+        if self.require_alias:
+            return f" _dbt_limit_subq_{self.table}"
```
mikealfare, line 214: Why are we going from `_dbt_limit_subq` to `_dbt_limit_subq_{self.table}`? I thought we just needed to deal with the fact that we don't want to render the alias unless we have to.

colin-rogers-dbt: This addresses a tertiary problem I identified where if you have multiple refs in a query you end up with duplicate aliases

mikealfare (approved):

mikealfare (approved):

PR #182, opened 2024-04-22
Allows snapshots to have a list as a unique key

### LINKED ISSUE #181: [Feature] Allow unique_key for snapshots to take a list

### Describe the feature

Right now the 'unique_key' for a snapshot must be a single column or an expression. If a user needs a combination of columns to be unique then they can use the expression to concatenate multiple columns together how they like.

Ideally users would be able to enter a list of columns, just like they can do for incremental models.

I considered adding additional parameters instead of changing the existing `unique_key` parameter. I think using the existing parameter is best because it's the most consistent with how incremental tables work.

### Who will this benefit?

Snapshot users who work in Postgres or Redshift can take advantage of performance optimizations in their databases with this change.

There are performance benefits from using multiple columns rather than an expression. In Postgres users would be able to pass in partition keys or indexed columns to speed up their snapshots, without needing to make an additional expression index. In Redshift users can pass in a separate distribution key to also get performance benefits.

These benefits can't be realized unless dbt allows the unique key to natively be a combination of columns, rather than a concatenation/combination expression.

Yes, I have a PR here and in dbt-core coming shortly

### Anything else?

_No response_

AGPapa:
resolves #181

### Problem

Currently the unique_key must be a string. It can be a column name or an expression. This PR changes it so it can be a list of column names or expressions.

### Solution

This is modeled off of this PR: https://github.com/dbt-labs/dbt-core/pull/4618

It checks if the unique_key is a list and modifies the code based on that. I am open to other approaches if we want something different (maybe turning a string unique_key into an array one element long?).

I couldn't find tests for the existing snapshot macros, so I'm also open to feedback on the best way to add tests for this feature. I have tested this by copying the macros to one of my dbt projects and running it with a modified version of dbt-core that allows a list for the unique_key.

### Checklist

- [s] This PR includes tests, or tests are not required/relevant for this PR
- [s] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX


AGPapa:
Thank you @gshank - let me know

PR #195, opened 2024-05-03
Add query recording for adapters which use SQLConnectionManager

peterallenwebb:
Resolves, in part, core issue https://github.com/dbt-labs/dbt-core/pull/9669

The overall feature is complex, so see the above issue for details. This PR adds support for the record/replay mechanism for queries executed using the execute() function on SQLConnectionManager. This is just a starting point for more complete record/replay of DWH interaction.

### Checklist


docs/guides/record_replay.md
```diff
@@ -0,0 +1,29 @@
...
+
+In practice, this means that any request sent to the warehouse must be recorded, along with the corresponding response. If this is done correctly, as described in the document linked in the intro, the Record portion of the Record/Replay subsystem should work as expected.
+
+At the time of this writing, there is only an incomplete implementation of this goal, which can be found in `dbt-adapters/dbt/adapters/record.py`.
+
+There are some iportant things to notice about this implementation. First, the QueryRecordResult class provides custom serialization methods `to_dict()` and `from_dict()`. This is necessary because the AdapterResponse and Agate.Table types cannot be automatically converted to and from JSON by the dataclass library, and JSON is the format used to persist recordings to disk and reload them for replay.
```
emmyoop, line 13:
```suggestion
There are some important things to notice about this implementation. First, the QueryRecordResult class provides custom serialization methods `to_dict()` and `from_dict()`. This is necessary because the `AdapterResponse` and `Agate.Table` types cannot be automatically converted to and from JSON by the dataclass library, and JSON is the format used to persist recordings to disk and reload them for replay.
```

docs/guides/record_replay.md
```diff
@@ -0,0 +1,29 @@
...
+
+Another important feature is that QueryRecordParams, implements the `_matches()` method. This method allows dbt-adapters to customize the way that the Record/Replay determines whether a query issued by dbt matches a previously recorded query. In this case, the method performs a comparison which attempts to ignore comments and whitespace which would not affect query behavior.
```
emmyoop, line 15:
```suggestion
Another important feature is that `QueryRecordParams` implements the `_matches()` method. This method allows `dbt-adapters` to customize the way that the Record/Replay determines whether a query issued by dbt matches a previously recorded query. In this case, the method performs a comparison which attempts to ignore comments and whitespace which would not affect query behavior.
```

docs/guides/record_replay.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Misc. Notes and Suggestsions
```
emmyoop, line 17:
```suggestion
## Misc. Notes and Suggestions
```
irony?

docs/guides/record_replay.md
```diff
@@ -0,0 +1,29 @@
...
+
+Currently, support for recording data warehouse interaction is very rudimentary, however, even rudimentary support is valuable and we should concentrating on extending it in a way that adds the most value with the least work. Usefulness, rather than perfection, is the initial goal.
```
emmyoop, line 19:
```suggestion
Currently, support for recording data warehouse interaction is very rudimentary, however, even rudimentary support is valuable and we should be concentrating on extending it in a way that adds the most value with the least work. Usefulness, rather than perfection, is the initial goal.
```

docs/guides/record_replay.md
```diff
@@ -0,0 +1,29 @@
...
+
+Picking the right functions to record, at the right level of abstraction, will probably be the most important part of carrying this work forward.
+
+Not every interaction with an external system has to be recorded in full detail, and authentication might prove to be a place where exclude sensitive secrets from the recording. For example, since replay will not actually be communicating with the warehouse, it may be possible to exclude passwords and auth keys from the parameters recorded, and to exclude auth tokens from the results.
```
emmyoop, line 23:
```suggestion
Not every interaction with an external system has to be recorded in full detail, and authentication might prove to be a place where we exclude sensitive secrets from the recording. For example, since replay will not actually be communicating with the warehouse, it may be possible to exclude passwords and auth keys from the parameters recorded, and to exclude auth tokens from the results.
```

docs/guides/record_replay.md
```diff
@@ -0,0 +1,29 @@
...
+
+In addition to adding an appropriate decorator to functions which communicate with external systems, you should check those functions for side-effects. Since the function's calls will be mocked out in replay mode, those side-effects will not be carried out during replay. At present, we are focusing on support for recording and comparing recordings, but this is worth keeping in mind.
+
+The current implementation records which dbt node issues a query, and uses that information to ensure a match during replay. The same node should issue the same query. A better model might be to monitor which connection issued which query, and associate the same connection with open/close operations, transaction starts/stops and so forth.
+
+
```
emmyoop, line 29:
```suggestion
The current implementation records which dbt node issues a query, and uses that information to ensure a match during replay. The same node should issue the same query. A better model might be to monitor which connection issued which query, and associate the same connection with open/close operations, transaction starts/stops and so forth.

```

emmyoop (commented):
Spell checked.  Otherwise looks good but you'll need a codeowner to approve.

peterallenwebb:
@colin-rogers-dbt asked a question about performance in a separate channel, which I'll answer here:

This will not have any impact on runtime performance when neither record or replay mode is in use. In those cases, the function annotations simply return the function unmodified, so it as if there was no annotation at all, and runtime behavior should be unchanged.

colin-rogers-dbt (approved):

Issue #210, opened 2024-05-14
[Bug] Test for `date_spine` can give a false positives


dbeatty10:
### Housekeeping

- [X] I am a maintainer of `dbt-adapters`

### Short description

The test case can pass even if the expected and actual data sets are not perfectly equal.

### Context

History: this test was migrated effectively as-is from dbt-utils to dbt-core to dbt-adapters.

This issue was discovered while working on https://github.com/dbt-labs/dbt-core/issues/10075.

### Acceptance criteria

```[tasklist]
- [ ] Replace the `left join` with a `full outer join`
```

### Testing

Manually alter the implementation of `date_spine` in order to check that the test case fails as expected in the following scenarios.

```[tasklist]
- [ ] Manually check that the test fails when the actual range returned by `date_spine` has too few days
- [ ] Manually check that the test fails when the actual range returned by `date_spine` has too many days
```

### Security

No security concerns.

### Docs

No doc updates needed.

### Consequences

This change impacts any adapter that imports this test.

We'd hope that no adapters are currently getting a false positive, and this change is a no-op for them.

For any adapters that were getting a false positive, this test would start failing for them, and they need to make updates accordingly.

Issue #208, opened 2024-05-14
[Feature] Test for `date_spine` should use a minimal date range


dbeatty10:
### Housekeeping

- [X] I am a maintainer of `dbt-adapters`

### Short description

The date range within [`fixture_date_spine.py`](https://github.com/dbt-labs/dbt-adapters/blob/d831caa33bc1a31d1162f1d1b3ccd6caaec08fd5/dbt-tests-adapter/dbt/tests/adapter/utils/fixture_date_spine.py) is larger than needed.

### Context

This doesn't affect the correctness of the test, but it does affect it's readability.

The test would be more readable with a minimal date range.

History: this test was migrated effectively as-is from dbt-utils to dbt-core to dbt-adapters. But now we'll probably want to clean it up a bit.

### Acceptance criteria

```[tasklist]
- [ ] The date range is no more than 3 days
- [ ] Optional: the date range spans a month (or year) boundary (e.g. `1999-12-31` to `2000-01-03`)
- [ ] Optional: the date range spans includes a leap day (e.g. `2024-02-28` to `2024-03-02`)
```

### Testing

See acceptance criteria above.

### Security

No security concerns.

### Docs

No doc updates needed.

### Consequences

This change impacts any adapter that imports this test. But it doesn't change the essence of the test and will effectively be a no-op.

PR #211, opened 2024-05-14
Avoid false positives in `date_spine` macro

dbeatty10:
resolves https://github.com/dbt-labs/dbt-adapters/issues/210
resolves https://github.com/dbt-labs/dbt-adapters/issues/208

### Problem

While working on https://github.com/dbt-labs/dbt-core/issues/10075, I noticed a few things about [this](https://github.com/dbt-labs/dbt-adapters/blob/d831caa33bc1a31d1162f1d1b3ccd6caaec08fd5/dbt-tests-adapter/dbt/tests/adapter/utils/fixture_date_spine.py) test case:
1. The test can give a false positives
1. The date range is larger than needed
1. There are hard-coded adapter types

The first one is the most impactful. This PR solves the first two but leaves the last one unsolved.

### Solution

1. Use a `full outer join` rather than a `left join` to eliminate possibility of a false positive
1. Reduce the date range from 9 down to 3 (which gets rid of 36 lines of code)

### Checklist

- [x] New tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc.)

### DIFF
--- a/dbt-tests-adapter/dbt/tests/adapter/utils/fixture_date_spine.py
+++ b/dbt-tests-adapter/dbt/tests/adapter/utils/fixture_date_spine.py
@@ -3,29 +3,17 @@
 models__test_date_spine_sql = """
 with generated_dates as (
     {% if target.type == 'postgres' %}
-        {{ date_spine("day", "'2023-09-01'::date", "'2023-09-10'::date") }}
+        {{ date_spine("day", "'2023-09-07'::date", "'2023-09-10'::date") }}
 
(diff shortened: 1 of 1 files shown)


adrianburusdbt:
Hey @dbeatty10, is this PR still needed? It's in draft for a long time now.
If yes, a couple of questions:
1. could you please describe an example when the false positive can be reproduced?
2. what is the reason the current range needs to be changed (aside from the added code lines)?

Thanks!

PR #217, opened 2024-05-16
Improve memory efficiency of process_results by iterating.

### LINKED ISSUE #218: Reduce Memory Use in process_results()

### Describe the feature

The process_results() function creates a second copy of query results, even though it is not needed. The function can be modified to iterate over the original results instead of creating a copy.

peterallenwebb:
resolves #218 

### Problem

The process of returning query results from execute() is memory inefficient, as multiple intermediate copies of the result data are maintained simultaneously.

In the case of `docs generate`, we are sometimes querying for information about every column in a schema. This can mean that a million or more records are returned in more extreme cases, resulting in gigabytes of memory allocation. In this scenario, maintaining multiple copies of the results, even temporarily, is untenable.

### Solution

Yield data rows one by one from process_results() rather than returning every row as a list, to eliminate one full copy of the result table. We could still do more work in this direction, but I documented a 33% reduction in memory associated with the `get_catalog` query with this approach.

### Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20240516-105757.yaml
+++ b/.changes/unreleased/Under the Hood-20240516-105757.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Improve memory efficiency of process_results()
+time: 2024-05-16T10:57:57.480672-04:00
+custom:
+  Author: peterallenwebb
+  Issue: "217"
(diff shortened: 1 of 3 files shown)


ChenyuLInx (approved):
Thanks Peter!

colin-rogers-dbt (approved):

PR #220, opened 2024-05-21
SPIKE: external tables as model materialization

dataders:
alternative to #216


dbt/include/global_project/macros/materializations/external_table/external_table.sql
```diff
@@ -0,0 +1,58 @@
+{% materialization external_table, default %}
+
+  {%- set identifier = model['alias'] -%}
+  {%- set full_refresh_mode = (should_full_refresh()) -%}
+
+  {%- set existing_relation = load_cached_relation(this) -%}
```
dataders, line 6: I need help here understanding why `existing_relation.type` returns `None` when the existing relation is an External Table

amychen1776:
closing for now because we are revisiting our thoughts on external tables for now

PR #231, opened 2024-05-28
Add get_relation_metadata capability

aranke:
resolves #

### Checklist

### DIFF
--- a/dbt/adapters/base/impl.py
+++ b/dbt/adapters/base/impl.py
@@ -222,6 +222,7 @@ class BaseAdapter(metaclass=AdapterMeta):
         - truncate_relation
         - rename_relation
         - get_columns_in_relation
+        - get_relation_metadata
         - get_column_schema_from_query
         - expand_column_types
         - list_relations_without_caching
@@ -626,6 +627,12 @@ class BaseAdapter(metaclass=AdapterMeta):
     def get_columns_in_relation(self, relation: BaseRelation) -> List[BaseColumn]:
         """Get a list of the columns in the given Relation."""
         raise NotImplementedError("`get_columns_in_relation` is not implemented for this adapter!")
+
+    @abc.abstractmethod
+    @available.parse_list
+    def get_relation_metadata(self, relation: BaseRelation) -> List[BaseColumn]:
+        """Get metadata about a relation's materialization (e.g., materialization type, size, number of rows)."""
+        raise NotImplementedError("`get_relation_metadata` is not implemented for this adapter!")
 
     @available.deprecated("get_columns_in_relation", lambda *a, **k: [])
     def get_columns_in_table(self, schema: str, identifier: str) -> List[BaseColumn]:
--- a/dbt/adapters/sql/impl.py
+++ b/dbt/adapters/sql/impl.py
@@ -12,6 +12,7 @@ from dbt.adapters.sql.connections import SQLConnectionManager
 
 LIST_RELATIONS_MACRO_NAME = "list_relations_without_caching"
 GET_COLUMNS_IN_RELATION_MACRO_NAME = "get_columns_in_relation"
+GET_RELATION_METADATA_NAME = "get_relation_metadata"
 LIST_SCHEMAS_MACRO_NAME = "list_schemas"
 CHECK_SCHEMA_EXISTS_MACRO_NAME = "check_schema_exists"
 CREATE_SCHEMA_MACRO_NAME = "create_schema"
@@ -41,6 +42,7 @@ class SQLAdapter(BaseAdapter):
         - get_catalog
         - list_relations_without_caching
         - get_columns_in_relation
+        - get_relation_metadata
     """
 
     ConnectionManager: Type[SQLConnectionManager]
@@ -156,6 +158,11 @@ class SQLAdapter(BaseAdapter):
     def get_columns_in_relation(self, relation):
         return self.execute_macro(
             GET_COLUMNS_IN_RELATION_MACRO_NAME, kwargs={"relation": relation}
+        )
+
+    def get_relation_metadata(self, relation):
+        return self.execute_macro(
+            GET_RELATION_METADATA_NAME, kwargs={"relation": relation}
         )
 
     def create_schema(self, relation: BaseRelation) -> None:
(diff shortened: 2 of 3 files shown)


dbt/adapters/base/impl.py
```diff
@@ -627,6 +628,11 @@ def get_columns_in_relation(self, relation: BaseRelation) -> List[BaseColumn]:
         """Get a list of the columns in the given Relation."""
         raise NotImplementedError("`get_columns_in_relation` is not implemented for this adapter!")
 
+    @abc.abstractmethod
+    def get_relation_metadata(self, relation: BaseRelation) -> Tuple[Any, Any]:
```
aranke, line 632: I'm not a fan of the return type here, but the return type I want `Tuple[Optional[TableMetadata], StatsDict]` is buried in `dbt-core/artifacts` 🥲

dbt/adapters/sql/impl.py
```diff
@@ -158,6 +160,9 @@ def get_columns_in_relation(self, relation):
             GET_COLUMNS_IN_RELATION_MACRO_NAME, kwargs={"relation": relation}
         )
 
+    def get_relation_metadata(self, relation) -> Tuple[Any, Any]:
```
aranke, line 163: Same comment on return type as above

dbt-tests-adapter/dbt/tests/adapter/basic/test_get_catalog_for_single_relation.py
```diff
@@ -0,0 +1,123 @@
...
+                        value=True,
+                        include=False,
+                        description="Indicates whether there are statistics for this table",
+                    ),
+                    "row_count": StatsItem(
+                        id="row_count",
+                        label="Row Count",
+                        value=0,
+                        include=True,
+                        description="Number of rows in the table as reported by Snowflake",
```
aranke, line 98: Note: how do we want to test this across adapters?

pyproject.toml
mikealfare, line 1: Reminder to revert changes in this file prior to merging. Also, we'll need to provide a lower bound on `dbt-common` since we expect `CatalogTable` to exist now.

dbt-tests-adapter/dbt/tests/adapter/basic/test_get_catalog_for_single_relation.py
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "my_model.sql": models__my_model_sql,
+        }
+
+    @pytest.fixture(autouse=True)
+    def clean_up(self, project):
```
mikealfare, line 45: I thought we only needed this when we create a schema other than the default test schema on the project fixture. I don't think it hurts anything to keep it here, but I also don't think we need it. If you do keep it, I would add `scope="class"` to the fixture. It doesn't matter at the moment since there is only one test method in this test class. But if we add a second, we can save ourselves some time by accounting for it now.

dbt-tests-adapter/dbt/tests/adapter/basic/test_get_catalog_for_single_relation.py
```diff
@@ -0,0 +1,123 @@
...
+        yield
+        with project.adapter.connection_named("__test"):
+            relation = project.adapter.Relation.create(
+                database=project.database, schema=project.test_schema
+            )
+            project.adapter.drop_schema(relation)
+
+    pass
```
mikealfare, line 53: This looks like a left over from stubbing out the class. We should remove it.

dbt-tests-adapter/dbt/tests/adapter/basic/test_get_catalog_for_single_relation.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def test_get_catalog_for_single_relation(self, project):
+        results = run_dbt(["seed"])
+        assert len(results) == 1
+
+        results = run_dbt(["run"])
+        assert len(results) == 1
+
+        if project.adapter.supports(Capability.GetCatalogForSingleRelation):
```
mikealfare, line 62: I don't think I would account for this here. If the adapter does not support this macro, then the maintainer will probably just not implement the test.

dbt-tests-adapter/dbt/tests/adapter/basic/test_get_catalog_for_single_relation.py
```diff
@@ -0,0 +1,123 @@
...
+                    database=project.database,
+                    schema=project.test_schema,
+                    identifier="MY_MODEL",
+                )
+                actual = project.adapter.get_catalog_for_single_relation(my_model_relation)
+
+                assert actual == expected
+
+
+class TestGetCatalogForSingleRelation(BaseGetCatalogForSingleRelation):
```
mikealfare, line 122: We don't need to do this anymore. This was done to support integration tests for `dbt-postgres`. However this is in a separate repo now, hence it would be implemented like any other adapter.

dbt-tests-adapter/dbt/tests/adapter/basic/test_get_catalog_for_single_relation.py
```diff
@@ -0,0 +1,123 @@
...
+                unique_id=None,
+            )
+
+            with get_connection(project.adapter):
+                my_model_relation = project.adapter.get_relation(
+                    database=project.database,
+                    schema=project.test_schema,
+                    identifier="MY_MODEL",
+                )
+                actual = project.adapter.get_catalog_for_single_relation(my_model_relation)
```
mikealfare, line 117: Minor nit, but I would only wrap this line (`actual = project.adapter...`) in the `with` block. The call to `get_relation` and the `assert` statement do not need the connection. The argument feels similar to minimizing what goes into a `try`/`except` block.

dbt-tests-adapter/dbt/tests/adapter/basic/test_get_catalog_for_single_relation.py
mikealfare: To make sure I understand, I assume you're referring to how to configure this for each adapter since the expected values are platform specific (e.g. this one refers to Snowflake). I would create a fixture in this class that builds the expected catalog and then refer to it in this test case. In the base I would either return an empty `CatalogTable`, or better, raise a `NotImplementedError` that specifies exactly what the maintainer needs to add to make the test work.

mikealfare (commented):
This is mostly where it needs to be. I left a few minor comments and addressed the question you had.

colin-rogers-dbt (approved):

colin-rogers-dbt (approved):

colin-rogers-dbt (approved):

PR #236, opened 2024-06-10
Use model alias for the CTE identifier generated during ephemeral materialization

jeancochrane:
Partially resolves https://github.com/dbt-labs/dbt-core/issues/5273. Reimplementation of https://github.com/dbt-labs/dbt-core/pull/5488 to match the new split between adapter and core repos.

### Problem

Ephemeral model materializations do not currently support using special characters like dots in file names, even if these special characters are sanitized using model aliases.

### Solution

Prioritize using the model alias (if specified) via `model.identifier` when creating a node identifier for ephemeral models.

This change requires a corresponding change to `dbt-core` to fix the bug in the special case of unit testing, which will be implemented in a separate PR.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20240610-195300.yaml
+++ b/.changes/unreleased/Fixes-20240610-195300.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Use model alias for the CTE identifier generated during ephemeral materialization
+time: 2024-06-10T19:53:00.086488231Z
+custom:
+    Author: jeancochrane
+    Issue: "5273"
--- a/dbt/adapters/base/relation.py
+++ b/dbt/adapters/base/relation.py
@@ -1,4 +1,3 @@
-import re
 from collections.abc import Hashable
 from dataclasses import dataclass, field
 from typing import (
@@ -233,8 +232,8 @@ class BaseRelation(FakeAPIObject, Hashable):
         )
 
     @staticmethod
-    def add_ephemeral_prefix(alias: str):
-        return re.sub(r"\W+", "", f"__dbt__cte__{alias}")
+    def add_ephemeral_prefix(name: str):
+        return f"__dbt__cte__{name}"
 
     @classmethod
     def create_ephemeral_from(
@@ -242,9 +241,11 @@ class BaseRelation(FakeAPIObject, Hashable):
         relation_config: RelationConfig,
         limit: Optional[int] = None,
     ) -> Self:
-        # Note that ephemeral models are based on the alias to give the user
-        # more control over the way that the CTE name is constructed
-        identifier = cls.add_ephemeral_prefix(relation_config.alias)
+        # Note that ephemeral models are based on the identifier, which will
+        # point to the model's alias if one exists and otherwise fall back to
+        # the filename. This is intended to give the user more control over
+        # the way that the CTE name is constructed
+        identifier = cls.add_ephemeral_prefix(relation_config.identifier)
         return cls.create(
             type=cls.CTE,
             identifier=identifier,
--- a/dbt/adapters/base/relation.py
+++ b/dbt/adapters/base/relation.py
@@ -1,3 +1,4 @@
+import re
 from collections.abc import Hashable
 from dataclasses import dataclass, field
 from typing import (
@@ -232,8 +233,8 @@ class BaseRelation(FakeAPIObject, Hashable):
         )
 
     @staticmethod
-    def add_ephemeral_prefix(name: str):
-        return f"__dbt__cte__{name}"
+    def add_ephemeral_prefix(alias: str):
+        return re.sub(r"\W+", "", f"__dbt__cte__{alias}")
 
     @classmethod
     def create_ephemeral_from(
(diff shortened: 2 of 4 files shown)


.changes/unreleased/Fixes-20240610-195300.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Use model alias for the CTE identifier generated during ephemeral materialization
+time: 2024-06-10T19:53:00.086488231Z
+custom:
+    Author: jeancochrane
+    Issue: "5273"
```
jeancochrane, line 6: This is the issue number in dbt-core; I wasn't sure whether I should make another one in this repo as well. Happy to do so if that's the correct procedure.

dbt/adapters/contracts/relation.py
```diff
@@ -49,6 +49,7 @@ def __delitem__(self, key): ...
 class RelationConfig(Protocol):
     resource_type: str
     name: str
+    alias: str
```
jeancochrane, line 52: My understanding is that this protocol is representing manifest nodes, which all appear to require `alias` attributes according to the [v12 manifest docs](https://schemas.getdbt.com/dbt/manifest/v12/index.html). If that's the case, I assume it should be fine to expect an `alias` attribute here, but let me know if that's off base and I can adjust it.

QMalcolm:
@cla-bot check

dbt/adapters/base/relation.py
```diff
@@ -232,17 +233,18 @@ def quoted(self, identifier):
...
     def create_ephemeral_from(
         cls: Type[Self],
         relation_config: RelationConfig,
         limit: Optional[int] = None,
     ) -> Self:
-        # Note that ephemeral models are based on the name.
-        identifier = cls.add_ephemeral_prefix(relation_config.name)
+        # Note that ephemeral models are based on the alias to give the user
+        # more control over the way that the CTE name is constructed
+        identifier = cls.add_ephemeral_prefix(relation_config.alias)
```
dbeatty10, line 247: There is already an [`identifier` property within `RelationConfig`](https://github.com/dbt-labs/dbt-adapters/blob/782a32b530e25afe34af6896392d9fb081dce9d1/dbt/adapters/contracts/relation.py#L55) which I believe comes from the [`alias` model config](https://docs.getdbt.com/reference/resource-configs/alias).

So I'm wondering if this small change would do the trick instead?

```suggestion
        identifier = cls.add_ephemeral_prefix(relation_config.identifier)
```

<br />

If that works, then we might be able to bypass a lot of the other proposed changes. i.e., just leave `add_ephemeral_prefix` as-is and skip the proposed regex substitution that strips certain characters.

dbeatty10:
Two places in the product docs that we could update to explain the effect of the `alias` config for `ephemeral` models:
- https://docs.getdbt.com/reference/resource-configs/alias
- https://docs.getdbt.com/docs/build/custom-aliases

dbt/adapters/base/relation.py
```diff
@@ -232,17 +233,18 @@ def quoted(self, identifier):
         )
 
     @staticmethod
-    def add_ephemeral_prefix(name: str):
-        return f"__dbt__cte__{name}"
+    def add_ephemeral_prefix(alias: str):
+        return re.sub(r"\W+", "", f"__dbt__cte__{alias}")
```
dbeatty10, line 237: If we do a regex substitution like this, then we could run into issues with ephemeral models like these:
1. `test_#_table`
1. `test_@_table`

If a third model tried to reference both of them, then there would be a name collision since they would both resolve to the same CTE name: `test__table` 💥

So I think we'd be better off **not** altering the value and leaving it as-is.

dbt/adapters/base/relation.py
jeancochrane: Nice call! I switched to `identifier` in e5c1b46807ca939548f2d503a5e6effdcc91ba86 and confirms that it works the same for both ephemeral models and unit tests in our project.

dbt/adapters/base/relation.py
jeancochrane: Great point, I removed this logic in e5c1b46807ca939548f2d503a5e6effdcc91ba86.

jeancochrane:
Thanks for the thorough review @dbeatty10! I think the changeset is now a lot simpler after e5c1b46807ca939548f2d503a5e6effdcc91ba86. Would you like me to make a docs PR to tweak the sections you mentioned above? I'd be happy to do so, but am also down to leave it to your team if you'd prefer to have editorial control over this one.

dbeatty10:
> Thanks for the thorough review @dbeatty10! I think the changeset is now a lot simpler after [e5c1b46](https://github.com/dbt-labs/dbt-adapters/commit/e5c1b46807ca939548f2d503a5e6effdcc91ba86).

🎉 Nice work @jeancochrane !

> Would you like me to make a docs PR to tweak the sections you mentioned above? I'd be happy to do so, but am also down to leave it to your team if you'd prefer to have editorial control over this one.

That would be awesome if you are willing to raise a PR in [docs.getdbt.com](https://github.com/dbt-labs/docs.getdbt.com). If you tag me in it, I'll give a review along with one of my colleagues on the documentation team.

ChenyuLInx (approved):
Thanks @jeancochrane for the thoughtful PR!  
This change looks good to me!

.changes/unreleased/Fixes-20240610-195300.yaml
ChenyuLInx: This works for now!

ChenyuLInx (approved):

ChenyuLInx:
Closing and reopening PR to make sure all needed checks are running

colin-rogers-dbt (approved):

PR #241, opened 2024-06-20
Remove get_catalog_for_single_relation from sql/impl.py

aranke:
Make consistent with `calculate_freshness_from_metadata`

### DIFF
--- a/dbt/adapters/sql/impl.py
+++ b/dbt/adapters/sql/impl.py
@@ -11,7 +11,6 @@ from dbt.adapters.sql.connections import SQLConnectionManager
 
 LIST_RELATIONS_MACRO_NAME = "list_relations_without_caching"
 GET_COLUMNS_IN_RELATION_MACRO_NAME = "get_columns_in_relation"
-GET_CATALOG_FOR_SINGLE_RELATION_NAME = "get_catalog_for_single_relation"
 LIST_SCHEMAS_MACRO_NAME = "list_schemas"
 CHECK_SCHEMA_EXISTS_MACRO_NAME = "check_schema_exists"
 CREATE_SCHEMA_MACRO_NAME = "create_schema"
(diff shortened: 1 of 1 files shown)


mikealfare (dismissed):

mikealfare:
Just to add context, if we find ourselves coming back to this in the future, this method is already defined in `BaseAdapter` with the same definition:

https://github.com/dbt-labs/dbt-adapters/blob/46e470cbbe8cefc98b00a2f0cd7223c43c2be4d1/dbt/adapters/base/impl.py#L632-L636

mikealfare:
I think there's a decent number of adapters that would be able to just define the macro and go on from there, without needing to overwrite the method. In a sense, this is the equivalent of an abstract method for a `SQLAdapter`-based adapter. Is there a specific reason to remove this and fallback to the abstract method in `BaseAdapter`?

aranke:
@mikealfare Is there an example of such an adapter?

aranke:
@mikealfare Synced with @jtcohen6 offline, this macro needs to do some post-processing to return a `CatalogTable` object; I don't think a macro can directly return this.

Given this fact, I'd prefer to remove this stub here to imply that there's more work required for an adapter maintainer.

mikealfare:
> @mikealfare Synced with @jtcohen6 offline, this macro needs to do some post-processing to return a `CatalogTable` object; I don't think a macro can directly return this.
> 
> Given this fact, I'd prefer to remove this stub here to imply that there's more work required for an adapter maintainer.

Ah, yes. I just looked at the `dbt-snowflake` implementation in your PR and there's quite a bit more logic there than the stub suggests. In that case I agree for two reasons. The stub will never be used and the presence of the stub is misleading.

In that case, can we also remove the related module variable for the macro name? I don't think it's used anywhere else.

aranke:
@mikealfare good call-out, done!

mikealfare (approved):

PR #243, opened 2024-06-21
Snapshot time data type

### LINKED ISSUE #242: Warn if snapshot_get_time is different data type than snapshot 'updated_at' field
### Description

This is part of https://github.com/dbt-labs/dbt-core/issues/10234

### Solution

This updates the snapshot materialization code to perform a data type check on the sql select statements used in constructing snapshot tables.

gshank:
resolves #242

### Problem

The values of the timestamp columns in a snapshot are different than the source table, leading to unexpected results.

### Solution

Add comparison of 'snapshot_get_time' and snapshot config 'updated_at' columns and issue a warning.

Note: test case is in https://github.com/dbt-labs/dbt-core/pull/10352

### Checklist


dbt/include/global_project/macros/materializations/snapshots/helpers.sql
```diff
@@ -168,14 +168,25 @@
 {% endmacro %}
 
 
-{% macro build_snapshot_staging_table(strategy, sql, target_relation) %}
```
MichelleArk, line 171: from what I can tell no 1P adapters override this macro specifically but it is technically a breaking change and could cause issues in projects / adapters that have overridden it. 

I can think of a couple of less-satisfying implementation approaches that would make this change backward-compatible: 
1. pass an optional `snapshot_select_sql` to `build_snapshot_staging_table` and use it instead of calling `snapshot_staging_table` in this macro if availab.e 
2. make a redundant call to  `snapshot_staging_table` [above](https://github.com/dbt-labs/dbt-adapters/pull/243/files#diff-678f35d7828b12f8a6fb7f70c01683aadf4f3a1141b1b3d4ab89aa356938c2baR40-R41), just to get `dbt_updated_at_data_type`.
  * I don't think this should have real performance implications since its just a templating macro
  * It's less slick but does keep the macros more independent from one another's results / execution order, so I think I prefer this option slightly over (1)

What do you think?

dbt/include/global_project/macros/adapters/timestamps.sql
```diff
@@ -15,6 +15,14 @@
     {{ current_timestamp() }}
 {% endmacro %}
 
+{% macro snapshot_get_time_data_type() %}
```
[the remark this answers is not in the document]
gshank: The "snapshot_get_time" macro is in this file, so it feels like "snapshot_get_time_data_type" belongs with it.

dbt/include/global_project/macros/materializations/snapshots/helpers.sql
gshank: I've changed this to keep the original macro signature so we're getting the sql twice.

peterallenwebb (approved):

dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
```diff
@@ -25,16 +25,20 @@
 
   {% set strategy_macro = strategy_dispatch(strategy_name) %}
   {% set strategy = strategy_macro(model, "snapshotted_data", "source_data", config, target_relation_exists) %}
+  {% set dbt_updated_at_data_type = none %}
```
mikealfare, line 28: I think this line is unnecessary since it will get set below in the `else` clause (or `if` if the condition is met).

dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
```diff
@@ -25,16 +25,20 @@
...
 
   {% if not target_relation_exists %}
 
       {% set build_sql = build_snapshot_table(strategy, model['compiled_code']) %}
+      {% set dbt_updated_at_data_type = get_updated_at_column_data_type(build_sql) or none %}
```
mikealfare, line 33: I read this a few times before realizing that all of these changes are only here to support a warning and do not affect `final_sql`. At first glance it looks like these lines are building up to the materialization sql whereas it's sort of a parallel process for a user warning.

Would you consider breaking this out into a separate macro in this file to make it more clear? Or perhaps a separate block of code in this macro? For example, if `snapshot_get_time_data_type` returns `none`, we don't even need to do the rest of this; so it could all live behind an if block with a comment suggesting it's just for a warning.

dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
gshank: I removed this line.

dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
gshank: I'm not sure what parts you think should be moved into a macro?  Collecting the value of dbt_updated_at_data_type needs to be collected in the big if/else because there are different sources of sql. The actual warning is only a line in an if after the big if/else.

mikealfare: Ah, yeah, I missed that they are different variables. So what about moving the exception into `get_updated_at_column_data_type`? It looks like the exception only depends on the result of that macro and `snapshot_get_time_data_type`, which could be called in that macro instead. In that case, the `get_` prefix may no longer make sense since the return value would no longer be needed.

gshank: It would be possible to do it that way, but it's my opinion that it would muddy the purpose of the snapshot_get_time_data_type macro. It's quite possible that we might want to call get_snapshot_time_data_type from a different place in the code, where we don't actually want to throw an exception, just get the data type. It has become very difficult to change the signature or purpose of macros once they have been released, so I think it makes more sense to keep the purpose of macros single purpose and limited. 

Additionally -- and this is clearly a matter of opinion/taste -- I prefer at this point to keep "what is happening" inline and readable, with regard to the check that is performed, instead of hiding it away in a macro. Additional changes are going to be coming up in this area and considering the difficulty of changing a macro signature once it's out there, I'd rather leave that for later. It's always possible to insert a new macro that combines "snapshot_get_time_data_type" and throwing the exception, but if we move the exception into the macro, it's more difficult to change later.

dbt/include/global_project/macros/adapters/timestamps.sql
MichelleArk, line 18: nit: could we rename this to have a `get_` prefix similar to `get_updated_at_column_data_type`? I realize it's a bit wordy especially with `get` already being in the name but I think it would help to differentiate this method from the variable its assigned to later on when we do: 
```
{% set get_time_data_type = snapshot_get_time_data_type() %}
```
could instead be: 
```
{% set snapshot_get_time_data_type = get_snapshot_get_time_data_type() %}
```

dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
MichelleArk: One way I can imagine refactoring this so the 'detect condition for warning + raise warning' bit is in its own macro while leaving the two helper getter methods alone is: 
1. create a variable for the input to `get_updated_at_column_data_type` (one of build_sql or snapshot_select_sql)
2. pass that into the new macro which would use it as input to `get_updated_at_column_data_type`, and also call `snapshot_get_time_data_type` and determine whether to raise + raise warning
3. call the new macro here: https://github.com/dbt-labs/dbt-adapters/pull/243/files#diff-678f35d7828b12f8a6fb7f70c01683aadf4f3a1141b1b3d4ab89aa356938c2baR77

MichelleArk: ^ I might be missing something here that makes this suggestion not viable, and if it's not I do also agree that this state is more explainable than overloading the `get_updated_at_column_data_type` method.

MichelleArk (approved):

dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
gshank: I've refactored this.

mikealfare: I like this approach better. I think it's easier to read and makes the happy path a bit more concise.

mikealfare (approved):

PR #244, opened 2024-06-21
Rework record/replay to record at the database connection level.

peterallenwebb:
### Problem

Our previous approach to recording DWH interaction short-circuited a large amount of our own code upon replay. This caused problems with replay behavior, and was also conceptually difficult to reason about.

### Solution

Record and replay at a different level of abstraction, cleanly separating dbt code from 3rd party code. 

### Checklist


dbt/adapters/record.py
```diff
@@ -1,67 +1,212 @@
...
+
+    def cursor(self):
+        # The native handle could be None if we are in replay mode, because no
+        # actual database access should be performed in that mode.
+        cursor = None if self.native_handle is None else self.native_handle.cursor()
+        return RecordReplayCursor(cursor, self.connection)
+
+
+@dataclasses.dataclass
+class CursorExecuteParams:
```
colin-rogers-dbt, line 23: haven't fully groked this so this might be dumb but just to keep these straight should we decompose these to separate modules? i.e. 
```
record/cursor_params.py
record/cursor_record.py
record/cursor_result.py
```

dbt/adapters/record.py
```diff
@@ -1,67 +1,206 @@
...
+    connection_name: str
+    operation: str
+    parameters: Union[Iterable[Any], Mapping[str, Any]]
+
+
+@Recorder.register_record_type
+class CursorExecuteRecord(Record):
```
colin-rogers-dbt, line 30: one other thought: not all adapters actually use 'cursors' and I _think_ what we're actually talking about 'connections'. It may just make more sense to drop the `Cursor*` prefix entirely

colin-rogers-dbt (approved):

PR #249, opened 2024-06-25
[Adap-207] remove subshell around get show sql with limit predicate

### LINKED ISSUE #207: Change the default behavior of `get_show_sql` to avoid wrapping in subquery
**Context:**
- https://github.com/dbt-labs/dbt-core/issues/9600
- [internal slack thread](https://dbt-labs.slack.com/archives/C03SK9W08BC/p1713265789155649)

**Implementation:**
- By default, `get_show_sql` should not wrap in a subquery. Instead, it should simply append `limit {{ limit }}` to the end of the query.
- This does not work for queries that already contain a `limit` clause, but it preserves ordering in all other cases.
- Because this only changes the behavior for `dbt show`, which is a development (not production) workflow, I do not believe this needs to be gated behind a behavior change flag. Users can always opt back into the previous behavior by reimplementing `get_show_sql` within their projects.

```sql
{% macro get_show_sql(compiled_code, sql_header, limit) -%}
  {%- if sql_header -%}
  {{ sql_header }}
  {%- endif -%}
  {{ compiled_code }}
  {% if limit is not none %}
  limit {{ limit }}
  {%- endif -%}
{% endmacro %}
```

**Acceptance tests:**
- I expect all existing tests to pass, since [`BaseShowLimit.test_limit`](https://github.com/dbt-labs/dbt-adapters/blob/d831caa33bc1a31d1162f1d1b3ccd6caaec08fd5/dbt-tests-adapter/dbt/tests/adapter/dbt_show/test_dbt_show.py#L28) is validating only that the `limit` is templated into the query (important!)
- Ideally, we'd write a new test that reliably reproduces incorrect ordering when the query is wrapped in a subquery, and then prove that it is always correctly sorted. Unfortunately I haven't managed to come up with a simple reproduction case of this.

VersusFacit:
resolves #207  

### Problem

We want to undo the subshelling of `--limit` modified `dbt` commands, like `run` and `show`. This is particularly important for the user experience of `show`, where subshelling plus `--limit` results in non guaranteed to be ordered results for display. 

### Solution

Alter `get_show_sql` to wrap an macro dispatch call: By default we simply append a `limit` predicate. This leaves the door open to users who wish to alter their SQL.

#### Things done:
- [x] remove the subshell for `--limit` predicates on `dbt show`
- [x] add new test for the syntax error -- tested using this Redshift PR 
    * tested locally using Redshift, which now is a PR: https://github.com/dbt-labs/dbt-redshift/pull/851
- [x] all existing tests continue to work _without modification_

My local tests using Redshift as the adapter substrate: image

#### Things _not_ done: 
* add a test case which should trigger the out of order behavior
    * I agree the usual modus operandi is to get a reprex and show we've fixed that behavior, but it's truly difficult to predict when postgres or any db will start doing this out of order. I tried valiantly with a local postgres db I put through all manner of situations; table and view, 1000 rows and 1000000 rows. Alas, nothing reliable could be ascertained. In the interest of impact, I proceeded instead to test `--limit` and inline `limit` together will flag a syntax error.

### Checklist


dbt/include/global_project/macros/adapters/show.sql
```diff
@@ -1,22 +1,18 @@
...
-  {%- if limit is not none -%}
-  {{ get_limit_subquery_sql(compiled_code, limit) }}
-  {%- else -%}
   {{ compiled_code }}
+  {%- if limit is not none %}
+  limit {{ limit }}
   {%- endif -%}
 {% endmacro %}
-
-{% macro get_limit_subquery_sql(sql, limit) %}
```
mikealfare, line 12: Even though we're no longer using this directly, we might want to keep it. The fact that it's dispatched suggests that concrete adapters probably implemented this somewhere along the line. They would get an exception if they're referencing this macro.

Along those lines, what happens if an adapter has implemented this macro in order to configure `get_show_sql`? That adapter-level implementation is lost now. I don't have a good answer for how to deal with that (and I'm not sure we can), but are we comfortable with that risk?

VersusFacit: Oh great points. I'll stick to macro interfaces in the future for these very important callouts.

mikealfare (approved):

dbt/include/global_project/macros/adapters/show.sql
```diff
@@ -1,22 +1,26 @@
...
+{%- endmacro -%}
 
-{% macro default__get_limit_subquery_sql(sql, limit) %}
-    select *
-    from (
-        {{ sql }}
-    ) as model_limit_subq
-    limit {{ limit }}
+{% macro default__get_limit_sql(sql, limit) %}
+  {{ compiled_code }}
```
greghroberts, line 22: @VersusFacit and @mikealfare  Looks like this line should be using the function parameter `{{ sql }}`. It works with `{{ compiled_code }}` I assume because of the [context being set in the get_show_sql macro](https://github.com/dbt-labs/dbt-core/blob/1d3d315249c81dbca70452e5a14fbfd0edad6b2a/core/dbt/task/show.py#L34), but that's likely hiding a future bug.

I could be wrong, but ran into this when providing a custom version of get_show_sql in my project.

PR #254, opened 2024-06-28
Update type hints for `get_adapter_constraint_support`

### LINKED ISSUE #253: [Bug] Incorrect type hints for `get_adapter_constraint_support`

### Current Behavior

I was debugging some issues with a third party tool calling the dbt Core API and digging in the code, I think that the type hint for the return value for the function `get_adapter_constraint_support` is incorrect.

It currently says `List[str]`

https://github.com/dbt-labs/dbt-adapters/blob/8ff69435a753f0a94dd1274d0e8dad199805cd13/dbt/adapters/factory.py#L191-L192

but in another part of the code, we access it as a dict.

https://github.com/dbt-labs/dbt-core/blob/71a8a411047cb6e2bd73a75c8a3c0bb8f9c2dbbb/core/dbt/contracts/graph/nodes.py#L649

https://github.com/dbt-labs/dbt-core/blob/71a8a411047cb6e2bd73a75c8a3c0bb8f9c2dbbb/core/dbt/contracts/graph/nodes.py#L682-L682

Looking at how the Constraints are defined in one of the adapters (BQ here), it looks like it is indeed a dict: 

https://github.com/dbt-labs/dbt-bigquery/blob/e67848972cce8ab51d68c6d1b1ced0f75816108c/dbt/adapters/bigquery/impl.py#L134-L140

I just realized that the type hint is defined in dbt-adapter so feel free to move the issue there.

### Expected Behavior

The type hint is correct

### Steps To Reproduce

See example of code above

### Relevant log output

_No response_

### Environment

_No response_

_No response_

### Additional Context

_No response_

dbeatty10:
resolves #253

### Problem

[`get_adapter_constraint_support`](https://github.com/dbt-labs/dbt-adapters/blob/8ff69435a753f0a94dd1274d0e8dad199805cd13/dbt/adapters/factory.py#L191-L192) returns [`CONSTRAINT_SUPPORT`](https://github.com/dbt-labs/dbt-adapters/blob/0739850873de0af14bc5506accb597256f0cdad1/dbt/adapters/base/impl.py#L254-L260) which is a dictionary of string keys and string values. But the type hint says `List[str]`.

### Solution

Update the type hint to be `Dict[str, str]`.

### Checklist


dbeatty10:
After making the suggested change, getting [this](https://github.com/dbt-labs/dbt-adapters/actions/runs/9717768223/job/26824167666?pr=254#step:4:49) error from mypy in CI:

```
dbt/adapters/factory.py:255: error: Incompatible return value type (got
"dict[str, str]", expected "list[str]")  [return-value]
        return FACTORY.get_adapter_constraint_support(name)
```

dbt/adapters/factory.py
```diff
@@ -188,7 +188,7 @@ def get_include_paths(self, name: Optional[str]) -> List[Path]:
     def get_adapter_type_names(self, name: Optional[str]) -> List[str]:
         return [p.adapter.type() for p in self.get_adapter_plugins(name)]
 
-    def get_adapter_constraint_support(self, name: Optional[str]) -> List[str]:
+    def get_adapter_constraint_support(self, name: Optional[str]) -> Dict[str, str]:
```
mikealfare, line 191: Do we still need the `# type: ignore`? My guess is that we needed that precisely because this type annotation was incorrect.

dbeatty10: Removed in https://github.com/dbt-labs/dbt-adapters/pull/254/commits/734f141d659b698f12997db4ab2572b3d1ead9e7

dbeatty10: Here's the result:

```
dbt/adapters/factory.py:192: error: "AdapterProtocol[Any, Any, Any, Any]" has
no attribute "CONSTRAINT_SUPPORT"  [attr-defined]
            return self.lookup_adapter(name).CONSTRAINT_SUPPORT
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
dbt/adapters/factory.py:192: error: Argument 1 to "lookup_adapter" of
"AdapterContainer" has incompatible type "str | None"; expected "str" 
[arg-type]
            return self.lookup_adapter(name).CONSTRAINT_SUPPORT
                                       ^~~~
Found 2 errors in 1 file (checked 48 source files)
```

colin-rogers-dbt (approved):

dbt/adapters/factory.py
colin-rogers-dbt: Yeah we can leave it for now I think this is because the method's name argument is `Optional`

PR #256, opened 2024-07-01
[CT-2819] default__alter_relation_add_remove_columns macro does not use quoting with case sensitive Snowflake relation

### LINKED ISSUE #250: [CT-2819] [Bug] default__alter_relation_add_remove_columns macro does not use quoting with case sensitive Snowflake relation

### Current Behavior

While working with DBT incremental config: `on_schema_change='append_new_columns'`
The append new columns flag is not able to capture the correct case-sensitive column name and add it to the incremental table causing the run to fail.

### Expected Behavior

Be able to apply to add the new columns to an incremental table with the `'append_new_columns'` config.

### Steps To Reproduce

1. Environment should have an existing incremental materialization while on_schema_change='append_new_columns'. 
2. Add a case-sensitive column name like "Schema_Test" in the source table and rerun the incremental table. The error will be reproduced.

### Relevant log output

```shell
invalid identifier '"Schema_Test"'
```

### Environment

```markdown
- OS:
- Python:
- dbt: 1.5
```

snowflake

### Additional Context

suggested the following as a workaround: 
```
     {% for column in add_columns %}
        add column "{{ column.name }}" {{ column.data_type }}{{ ',' if not loop.last }}
      {% endfor %}{{ ',' if remove_columns | length > 0 }}
            
     {% for column in remove_columns %}
       drop column "{{ column.name }}" {{ ',' if not loop.last }}
     {% endfor %}
```
saved in project macro folder.

McKnight-42:
resolves #250 

### Problem
Try to fix issue around case sensitive column quoting in macro `default__alter_relation_add_remove_columns`

### Solution

update the dbt-adapters default implementation test against postgres and other adapted as needed to maintain compaitiablity, update dbt-snowflake implementation of macro

create test in dbt-tests-adapters and inherit in adapters as needed.

### Checklist


McKnight-42:
NOTES: usage across adapters

Adapters that use own implementation of alter_relation_add_remove_columns (defined in dot-adapters)
    - Dbt-snowflake
    - Dbt-redshift
    - Dbt-spark

Adapters that would use default if at all (not referenced in either adapter directly)
    - Dbt-postgres 
    - Dbt-bigquery

dbt-tests-adapter/dbt/tests/adapter/incremental/fixtures.py
```diff
@@ -303,3 +303,34 @@
...
+
+{% if var("version", 0) == 0 %}
+
+    select {{ dbt.current_timestamp() }} as inserted_at, 'john' as name
+
+{% else %}
+
+    -- add a non-zero version to the end of the command to get a different version:
+    -- --vars "{'version': 1}"
+    select {{ dbt.current_timestamp() }} as inserted_at, 'john' as Name, 'engineer' as "Job"
```
mikealfare, line 322: This may be intentional, but should `Name` here match `name` above (casing)?

dbt/include/global_project/macros/adapters/columns.sql
```diff
@@ -123,11 +123,11 @@
      alter {{ relation.type }} {{ relation.render() }}
 
             {% for column in add_columns %}
-               add column {{ column.name }} {{ column.data_type }}{{ ',' if not loop.last }}
+               add column {{ adapter.quote(column.name) }} {{ column.data_type }}{{ ',' if not loop.last }}
```
mikealfare, line 126: Do we always want to quote? Or is there a config that we need to condition on, similar to the quote policy for relation names?

dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_on_schema_change.py
```diff
@@ -84,3 +84,31 @@ def test_run_incremental_fail_on_schema_change(self, project):
...
+        assert "Job" not in logs
+        run_dbt(["run", "--vars", "{'version': 1}"])
+        run_dbt(["show", "--inline", "select * from {{ ref('dim_artists') }}"])
+        res, logs = run_dbt_and_capture(
+            ["show", "--inline", "select * from {{ ref('dim_artists') }}"]
+        )
+        assert "Job" in logs
+
+
+class TestIncrementalCaseSenstivityOnSchemaChange(BaseIncrementalCaseSenstivityOnSchemaChange):
```
mikealfare, line 113: We don't need this class. The reason `TestIncrementalOnSchemaChange` is there is to preserve backwards compatibility. These tests used to run when testing `dbt-postgres`. Now that that's in its own repo and explicitly implements tests from `dbt-tests-adapter`, we no longer need to create these classes here.

dbt-tests-adapter/dbt/tests/adapter/incremental/fixtures.py
McKnight-42: was intentional to see if they would be seen differently or cause a issue (see bigquery) happy to make it lower case

dbt/include/global_project/macros/adapters/columns.sql
colin-rogers-dbt: I think adapter.quote leverages the same  quote policy as relation names

dbeatty10: @mikealfare are you thinking something like [this](https://github.com/dbt-labs/dbt-adapters/blob/782a32b530e25afe34af6896392d9fb081dce9d1/dbt/include/global_project/macros/relations/rename.sql#L31) (for relations) or [this](https://github.com/dbt-labs/dbt-adapters/blob/782a32b530e25afe34af6896392d9fb081dce9d1/dbt/include/global_project/macros/relations/table/create.sql#L45) (for columns)?

In this particular case, I think we need to match the quoting from [here](https://github.com/dbt-labs/dbt-core/blob/5310d3715cdaa109eea03995731425a85cd6cd33/core/dbt/include/global_project/macros/materializations/models/incremental/column_helpers.sql#L11) (which would mean always quoting via `adapter.quote()`). I forget the exact details -- and I could be wrong! -- but there's some notes under the "**Root Cause**" toggle in https://github.com/dbt-labs/dbt-adapters/issues/250#issuecomment-2192228487.

dbeatty10: Side note: here are the primary resources I've been using when researching quoting or case-sensitive issues:
- https://github.com/dbt-labs/docs.getdbt.com/issues/3518
- https://github.com/dbt-labs/dbt-core/issues/2986

McKnight-42: @dbeatty10 oh these are really coll gonna dig in haven't seen these before.

mikealfare: @colin-rogers-dbt `adapter.quote()` will always [quote](https://github.com/dbt-labs/dbt-adapters/blob/43ea32d821ed041cdc03175308e1f411211847a3/dbt/adapters/sql/impl.py#L209-L210).

@dbeatty10 If the objective is to always quote the column in this case, that's fine. I just know we've run into issues before where we quote something that doesn't get quoted, or vice versa, and then it fails a condition in a filter where it's supposed to pass. I haven't seen it with columns, so maybe it doesn't apply here.

dbt-tests-adapter/dbt/tests/adapter/incremental/fixtures.py
dbeatty10, line 322: I wonder if using [`adapter.quote`](https://docs.getdbt.com/reference/dbt-jinja-functions/adapter#quote) and/or [`string_literal`](https://docs.getdbt.com/reference/dbt-jinja-functions/cross-database-macros#string_literal) would enable this test to work across more adapters without modification?
 
```suggestion
    select {{ dbt.current_timestamp() }} as inserted_at, {{ dbt.string_literal("john") }} as Name, {{ dbt.string_literal("engineer") }} as {{ adapter.quote("Job") }}
```

dbt-tests-adapter/dbt/tests/adapter/incremental/fixtures.py
```diff
@@ -303,3 +303,34 @@
 
 from source_data
 """
+
+_MODELS__SRC_ARTISTS = """
```
dbeatty10, line 307: Nit: The content of the model looks like "jobs" now instead of "artists". So ideally they'd match instead of being different.

dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_on_schema_change.py
```diff
@@ -84,3 +84,27 @@ def test_run_incremental_fail_on_schema_change(self, project):
...
+        res, logs = run_dbt_and_capture(
+            ["show", "--inline", "select * from {{ ref('dim_artists') }}"]
+        )
+        assert "Job" not in logs
+        run_dbt(["run", "--vars", "{'version': 1}"])
+        run_dbt(["show", "--inline", "select * from {{ ref('dim_artists') }}"])
+        res, logs = run_dbt_and_capture(
+            ["show", "--inline", "select * from {{ ref('dim_artists') }}"]
+        )
+        assert "Job" in logs
```
dbeatty10, line 110: The `run_dbt(["show" ...` portions probably aren't necessary. They were just so people could see the state of the data within this reprex: https://github.com/dbt-labs/dbt-adapters/issues/250#issuecomment-2192228487

Instead of the assertions here, you might be able to just rely upon `expect_pass=True` as the default for [`run_dbt`](https://github.com/dbt-labs/dbt-core/blob/014444dc18dd18ab75b81ad62530256ec09df22a/core/dbt/tests/util.py#L73-L75) and [`run_dbt_and_capture`](https://github.com/dbt-labs/dbt-core/blob/014444dc18dd18ab75b81ad62530256ec09df22a/core/dbt/tests/util.py#L111-L113).

From the original bug report in https://github.com/dbt-labs/dbt-adapters/issues/250, my understanding was that it actually produced an error.

cboethigtrellance:
Anything else required or need help with to get this over the line?

PR #257, opened 2024-07-02
[Bug] Fix scenarios where `--empty` flag pushes limit 0 into metadata queries

### LINKED ISSUE #213: [Bug] --empty flag generates invalid SQL when running `adapter.get_columns_in_relation`
### Is this a new bug?

### Current Behavior

When running the `--empty` flag introduced in DBT 1.8 `adapter.get_columns_in_relation` generates and runs an invalid DESCRIBE TABLE statement, halting the run. The statement looks roughly like:

`describe table (select * from TABLE_NAME_HERE where false limit 0)`

Which generates the error:
```
  001003 (42000): SQL compilation error:
  syntax error line 1 at position 15 unexpected '('.
```

### Expected Behavior

`adapter.get_columns_in_relation` should fallback to some reasonable default when running with the --empty flag, or alternatively throws a more intelligible error message. 

For our usages, I'm working around it by guarding calls to `adapter.get_columns_in_relation` behind a check on whether we're running in empty mode. However, this is cumbersome and a bit error prone. 

### Steps To Reproduce

1. In DBT 1.8 on Snowflake
2. Run `./dbt run --empty` on a model that uses `adapter.get_columns_in_relation`
3. The error will be thrown 

### Relevant log output

_No response_

### Environment

```markdown
- OS: Linux
- Python: 3.11
- dbt-adapter: dbt-snowflake
```

### Additional Context

_No response_

mikealfare:
resolves #213

### Problem

We fixed `dbt-snowflake` when using `--empty` and running metadata queries.

### Solution

Add the tests for that fix so other adapters can use it if necessary.

### Implementation

1. Merge this PR demonstrating that these tests work
2. Move the bulk of this PR into `dbt-tests-adapter` (we`re here)
3. Create a final PR that removes most of this and uses the `dbt-tests-adapter` tests

### Checklist


colin-rogers-dbt:
Do we need to update this one too? https://github.com/dbt-labs/dbt-adapters/blob/bug/213-fix-empty-flag/dbt/include/global_project/macros/materializations/tests/where_subquery.sql#L9

mikealfare:
> Do we need to update this one too? https://github.com/dbt-labs/dbt-adapters/blob/bug/213-fix-empty-flag/dbt/include/global_project/macros/materializations/tests/where_subquery.sql#L9

I'm honestly not sure. Based on where this macro is, I assume it's used in the test materialization. However, I don't see where this macro is called. I looked in both `dbt-adapters` and `dbt-snowflake` and there's no reference outside of defining the macro. Either way, I think we'd want to limit the results here just like we would in other materializations. It also doesn't create invalid sql like it does in `describe`/`show` queries. I can add it if you feel otherwise, but I'm inclined to leave this as is.

colin-rogers-dbt (approved):

colin-rogers-dbt (approved):

PR #260, opened 2024-07-05
render to and to_columns on column and model level fk constraints

### LINKED ISSUE #271: [Constraints] Render 'to' and 'to_columns' fields on foreign key constraints
### Housekeeping

- [X] I am a maintainer of `dbt-adapters`

### Short description

As part of https://github.com/dbt-labs/dbt-core/issues/8062. we are parsing and compiling new `to` and `to_columns` fields on ModelLevelConstraint and ColumnLevelConstraint. dbt-adapters should use these new fields, when populated, to render out the foreign key constraint instead of `expression`. `expression` support should remain as-is for backward-compatability.

### Context

_No response_

### Acceptance criteria

```[tasklist]
- [x] Use `to` and `to_column` fields to render foreign key constraints on model-level and column-level constraints
- [x] maintain `expression` support for backward compatability
```

### Testing

```[tasklist]
 - [x] new integration test for `to` and `to_columns` usage as this will be the recommended practice
 - [x] ensure existing test coverage (e.g. unit testing) exists for `expression` support
```

### Security

N/A

### Docs

https://github.com/dbt-labs/dbt-core/pull/10414#issuecomment-2243915257

### Consequences

This will update the base implementation of constraint rendering in a backward-compatible manner. Other adapters will be able to extend it as necessary to support `to` and `to_columns`, but will continue to be able to use the default implementation.

MichelleArk:
resolves #271 
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/# - part of dbt-core issue

Core PR: https://github.com/dbt-labs/dbt-core/pull/10414

### Problem

As part of https://github.com/dbt-labs/dbt-core/issues/8062. we are parsing and compiling new to and to_columns fields on ModelLevelConstraint and ColumnLevelConstraint.

### Solution

dbt-adapters uses these new fields, when populated, to render out the foreign key constraint instead of expression. expression support remains as-is for backward-compatability.

### Checklist


MichelleArk:
Adapter Tests: 
* Redshift: https://github.com/dbt-labs/dbt-redshift/pull/880/files
* BigQuery: https://github.com/dbt-labs/dbt-bigquery/pull/1273
* Snowflake: https://github.com/dbt-labs/dbt-snowflake/pull/1143

dbt/adapters/base/impl.py
```diff
@@ -1682,20 +1687,29 @@ def render_model_constraint(cls, constraint: ModelLevelConstraint) -> Optional[s
...
+                rendered_model_constraint = f"{constraint_prefix}foreign key ({column_list}) references {constraint.expression}"
+            elif constraint.to and constraint.to_columns:
+                rendered_model_constraint = f"{constraint_prefix}foreign key ({column_list}) references {constraint.to} ({', '.join(constraint.to_columns)})"
         elif constraint.type == ConstraintType.custom and constraint.expression:
-            return f"{constraint_prefix}{constraint.expression}"
-        else:
-            return None
+            rendered_model_constraint = f"{constraint_prefix}{constraint.expression}"
+
+        return rendered_model_constraint
```
MichelleArk, line 1712: refactored this for mypy to have a clear return statement

dbt/adapters/base/impl.py
```diff
@@ -1682,20 +1687,29 @@ def render_model_constraint(cls, constraint: ModelLevelConstraint) -> Optional[s
...
-        elif constraint.type == ConstraintType.foreign_key and constraint.expression:
-            return f"{constraint_prefix}foreign key ({column_list}) references {constraint.expression}"
+            rendered_model_constraint = (
+                f"{constraint_prefix}primary key{constraint_expression} ({column_list})"
+            )
+        elif constraint.type == ConstraintType.foreign_key:
+            if constraint.expression:
+                rendered_model_constraint = f"{constraint_prefix}foreign key ({column_list}) references {constraint.expression}"
+            elif constraint.to and constraint.to_columns:
+                rendered_model_constraint = f"{constraint_prefix}foreign key ({column_list}) references {constraint.to} ({', '.join(constraint.to_columns)})"
```
MichelleArk, line 1708: this is the only non-refactor change to this method

dbt-tests-adapter/dbt/tests/adapter/constraints/fixtures.py
```diff
@@ -363,7 +363,8 @@
           - type: check
             expression: id >= 1
           - type: foreign_key
-            expression: {schema}.foreign_key_model (id)
+            to: ref('foreign_key_model')
```
mikealfare, line 366: Are we removing `expression` then? If so, is this potentially breaking? If not, do we need a test that still captures scenarios where we use `expression`?

gshank: "expression" isn't being removed, the test was just changed to use "to:" instead.

gshank (approved):

dbt-tests-adapter/dbt/tests/adapter/constraints/fixtures.py
MichelleArk: "expression" backward compatibility is still preserved, and there is unit test coverage for that. Personally I don't think it's worth the lift to introduce an integration test across all adapters given that `to` and `to_columns` will be the recommended way to use foreign key constraints going forward, and that we already have more thorough edge case handling for `expression` in unit testing.

MichelleArk: here's where the back-compat is implemented: 
* model-level: https://github.com/dbt-labs/dbt-adapters/pull/260/files#diff-e2306b841793e5b07177010e8920c59163df1536306fc76e7f6a09d36dcae4d3R1705-R1706
* column-level: https://github.com/dbt-labs/dbt-adapters/pull/260/files#diff-e2306b841793e5b07177010e8920c59163df1536306fc76e7f6a09d36dcae4d3R1606-R1607

mikealfare: I found the unit tests. I think that's fine. I get hesitant when we remove tests for supported functionality, even for deprecated functionality, because that's where we wind up with regressions. But the rationale makes sense. Should we have a deprecation warning on the use of `expression` then?

MichelleArk: We're not deprecating or planning to deprecate `expression` in the future, so I don't think a deprecation warning makes sense to introduce. `expression` will still be useful for other constraint types, and can serve as a more fleible fallback for more complex rendering use cases, or perhaps for adapters where the default 'to' and 'to_columns' implementation is not compatible + has not been overwritten by the adapter maintainer yet. Likely the way we'll document this is that `to` and `to_columns` are the recommended way to define foreign key constraints, and also describe the expression option.

mikealfare: Ok, I think I'm convinced of this then. `expression` is still a plausible use case because `to` / `to_columns` only covers 90% of the use cases with `expression` as a backup. We also need to account for `expression` because we don't want to break third party adapters.

dbt/adapters/base/impl.py
```diff
@@ -1602,8 +1602,13 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> Optional
             rendered_column_constraint = f"unique {constraint_expression}"
         elif constraint.type == ConstraintType.primary_key:
             rendered_column_constraint = f"primary key {constraint_expression}"
-        elif constraint.type == ConstraintType.foreign_key and constraint_expression:
-            rendered_column_constraint = f"references {constraint_expression}"
+        elif constraint.type == ConstraintType.foreign_key:
```
mikealfare, line 1605: If the preferred method is `to` and `to_columns`, we should evaluate that condition first.

dbt/adapters/base/impl.py
```diff
@@ -1682,20 +1687,29 @@ def render_model_constraint(cls, constraint: ModelLevelConstraint) -> Optional[s
...
+            )
         elif constraint.type == ConstraintType.primary_key:
             constraint_expression = f" {constraint.expression}" if constraint.expression else ""
-            return f"{constraint_prefix}primary key{constraint_expression} ({column_list})"
-        elif constraint.type == ConstraintType.foreign_key and constraint.expression:
-            return f"{constraint_prefix}foreign key ({column_list}) references {constraint.expression}"
+            rendered_model_constraint = (
+                f"{constraint_prefix}primary key{constraint_expression} ({column_list})"
+            )
+        elif constraint.type == ConstraintType.foreign_key:
```
mikealfare, line 1704: If the preferred method is `to` and `to_columns`, we should evaluate that condition first.

mikealfare (approved):

nataliefiann:
Hiya @MichelleArk 

For the adapter tests, do you have any info as to if / when this will be available on other adapters?

Kind Regards
Natalie

PR #261, opened 2024-07-09
[Feature] Enable user to provide custom adapter via profiles.yml

ismailsimsek:
resolves https://github.com/dbt-labs/dbt-core/issues/3962

### Problem
Currently customizing dbt adapters and dbt functionalities is very hard to not possible. This PR enables end user to extend exsting adapters and add his/her own customization(special needs) to to it

### Solution
Enabling end user to provide his/her  custom adapter, using the  `adapter_class` setting in `profiles.yml` 

```yaml
snowflake:
  target: dev
  outputs:
    dev:
      type: snowflake
      adapter_class: my.customized.adapter.SnowflakeAdapterV2Customized
```
and user adapter could be something like
```python
class SnowflakeAdapterV2Customized(SnowflakeAdapter):
    def __init__(self, config) -> None:
        print(f"WARNING: Using User Provided DBT Adapter: {type(self).__module__}.{type(self).__name__}")
        super().__init__(config=config)

    @available
    def my_custom_method(self, arg1: str, arg2: str):
        print('do something here')
```
 

### Checklist

### DIFF
--- a/dbt/adapters/contracts/connection.py
+++ b/dbt/adapters/contracts/connection.py
@@ -127,6 +127,7 @@ class LazyHandle:
 class Credentials(ExtensibleDbtClassMixin, Replaceable, metaclass=abc.ABCMeta):
     database: str
     schema: str
+    adapter_class: str
     _ALIASES: ClassVar[Dict[str, str]] = field(default={}, init=False)
 
     @abc.abstractproperty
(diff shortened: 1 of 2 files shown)


ismailsimsek:
cc @jtcohen6 could you please have look? I believe the changes gives the idea. if it looks good then i will ad test as well

dbeatty10:
Like @jtcohen6 explained in https://github.com/dbt-labs/dbt-core/issues/3962#issuecomment-929394776, it's been a deliberate decision to not allow user-space Python code within your dbt project. We see this as serving as an important security and complexity safeguard, and we do not plan to change it. So I'm closing this PR accordingly.

PR #265, opened 2024-07-19
[CT-1661] Quotes usernames with special characters

### LINKED ISSUE #156: [CT-1661] [Bug] Grants fails to correctly quote username when revoking

### Current Behavior

As part of the new grants feature, https://docs.getdbt.com/reference/resource-configs/grants dbt will attempt to revoke users not part of the grant before reapplying users listed in the grant. 
DBT does this (at least in the Redshift adaptor ) by pulling the table permissions and explicitly looping through every user then using a revoke command like 
revoke select on "database"."schema"."table" from user1, user2 ;
The user is not quoted so a user like user.name will fail with 
syntax error at or near "."
  LINE 2: ...user1, user.name, user2...

I am fairly certain the macro in https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/include/global_project/macros/adapters/apply_grants.sql line 82 just needs some quotes

### Expected Behavior

the revoke command should quote users
revoke select on "database"."schema"."table" from "user1", "user2" ;

### Steps To Reproduce

Create a user with a period in the name for example in Redshift 
create user "first.last";

Create a simple dbt incremental model and user the grant syntax to give permission to that table within dbt.

Run DBT twice (on the first run the revoke macro is not run as the table is new)

### Relevant log output

_No response_

### Environment

```markdown
- OS:Ubuntu 20.04
- Python: 3.8.10 
- dbt: 1.3.1
```

redshift

### Additional Context

_No response_

qmg-karan:
resolves #156 

### Problem

This PR solves issues where dbt tries to apply grant-related SQL queries against users that have special characters in their names.  

### Solution

Quotes are added around users that hold special characters. So far, this just includes a few special characters that people have raised issues for across several dbt repos ("**.**" in #156, "**-**" in [#55](https://github.com/dbt-labs/dbt-postgres/issues/55), and "**@**" in my case.

Please suggest how this can be tested, I'm not entirely sure about this because the unit-tests seem to pass after I've made this change. Not sure where the tests should be added for this.

### Checklist


gastonsilva:
@cla-bot[bot] check

bbodensieck:
@qmg-karan could you verify by some tests that the implementation works?

somewhere here I guess:
dbt-tests-adapter/src/dbt/tests/adapter/grants

bbodensieck:
@dbeatty10 could you have a look?

PR #269, opened 2024-07-22
When using incremental tables, the column data types are invalidly updated

### LINKED ISSUE #276: [Bug] When using incremental tables, the column data types are invalidly updated
### Current Behavior

When an incremental table has contracted columns, the table has the column definition for varchar/character varying size changed to 256.

Original issue in core: https://github.com/dbt-labs/dbt-core/issues/10362

gshank:
resolves #276

Original issue in https://github.com/dbt-labs/dbt-core/issues/10362

### Problem

In dbt-postgres, contract changes were applied to temporary tables, causing original tables to be changed to match.

### Solution

If a contract exists, do not modify the columns of incremental tables. 

### Checklist


dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
```diff
@@ -39,9 +39,12 @@
       {% set need_swap = true %}
   {% else %}
     {% do run_query(get_create_table_as_sql(True, temp_relation, sql)) %}
-    {% do adapter.expand_target_column_types(
-             from_relation=temp_relation,
-             to_relation=target_relation) %}
+    {% set contract_config = config.get('contract') %}
+    {% if not contract_config.enforced %}
```
mikealfare, line 43: Just to confirm, since jinja can have a mind of its own, we want this behavior:

1. `contract` is in `config` and `contract.enforced == True` - don't execute
1. `contract` is in `config` and `contract.enforced == False` - execute
1. `contract` is not in `config` - execute

Is that correct? I think that's what happens here? The questionable case is 3. I'm pretty sure `{% if contract_config.enforced %}` evaluates to `false` if `contract_config` is `None`. What I don't know is if `{% if not contract_config.enforced %}` evaluates to `true` when `contract_config` is `None`.

gshank: I've changed it to `{% if not contract_config or not contract_config.enforced %}`, which should avoid that.

mikealfare:
@gshank I read through the original issue and I'm not clear on how this resolves that issue. If I understand this change correctly, it looks like we're doing this:

1. Remove the `__repr__` to use the default from dataclasses
2. Only add columns when there is no contract enforcement

I don't see how 1. would affect anything. And 2. removes adjusting columns entirely for contract-enforced models instead of correcting the type. From this [PR](https://github.com/dbt-labs/dbt-core/pull/10371), it seems like there's an assumption that models with a contract will never change. Is that true? What happens if the user deliberately adds a column, and updates the contract to reflect the new column? Is that something we don't support?

gshank:
I removed the __repr__ only because it caused a lot of confusion when debugging, because it wasn't adequately displaying the contents of the class and the default worked better. It's not related to the fix.

The problem that this fixes is that the intermediate tables have different data types than the actual incremental table, and that is causing the actual incremental table to have the data types updated from valid (contract enforced) data types to data types that do not match the contract.

It's not totally clear to me what the point of "expand_column_data_types" is supposed to be, so I tried to limit the change to just what was necessary to fix this bug.

mikealfare:
> I removed the repr only because it caused a lot of confusion when debugging, because it wasn't adequately displaying the contents of the class and the default worked better. It's not related to the fix.

That's what I suspected, and I'm fine with that.

> The problem that this fixes is that the intermediate tables have different data types than the actual incremental table, and that is causing the actual incremental table to have the data types updated from valid (contract enforced) data types to data types that do not match the contract.

I see how this is happening too, and agree that it shouldn't be.

> It's not totally clear to me what the point of "expand_column_data_types" is supposed to be, so I tried to limit the change to just what was necessary to fix this bug.

I could be totally off here, but I believe the point of that is to account for the scenario where the model is updated to include new columns. Say the old model is `select a, b, c from my_source`. The resulting incremental table would have three columns. But then we need to update the model to be `select a, b, c, d from my_source` because [reasons]. We need to run an alter statement on the incremental table before adding the new records so that it can accept column `d`. This scenario could happen whether or not the model is contract-enforced. The contract-enforced scenario just requires that the user update the contract as well when this happens. Assuming my interpretation is correct, I believe this change would inadvertently lock folks into a fixed incremental schema when using a contract, or require a full refresh of the table any time a column is added.

mikealfare (approved):

PR #282, opened 2024-08-13
Add Behavior Flag Framework

### LINKED ISSUE #281: [Behavior Flags] Add Behavior Flag Framework to `dbt-adapters`
### Housekeeping

- [X] I am a maintainer of `dbt-adapters`

## Short description

As an adapter maintainer, I want to use the Behavior Flag Framework in `dbt-common`, so that I can safely implement breaking changes while notifying users.

```[tasklist]
## Acceptance criteria
- [x] I can register behavior flags whenever an instance of BaseAdapter is created
- [x] Users can override the default setting of a behavior flag by setting the flag in the `flags` section of their `dbt_project.yml`
- [x] I can reference each flag as a property of the adapter, e.g. `adapter.behavior.my_behavior_flag`
- [ ] Concrete adapters (e.g. `dbt-postgres`) can implement their own behavior flags; these flags will get incorporated into the same behavior flags that are implemented in `dbt-adapters`
- [ ] Adapter maintainers have access to documentation that explains how to implement their own behavior flags
- [ ] End user documentation is updated with this change; in particular, there's a new source of tracking, and a new motivation to allow tracking (a "vote" to not have legacy behavior removed if you're still using it)
```

```[tasklist]
## Suggested testing
- [x] Confirm that adding a behavior flag with no user override results in the expected default setting for the flag
- [x] Confirm that specifying a behavior flag in `dbt_project.yml::flags` results in overriding that flag
```

## Security

This is effectively config, hence no more risky than current functionality.

## Docs

We should make sure the behavior flag docs in docs.getdbt.com are updated. We also need to add docs for adapter maintainers.

## Consequences

We should work closely with third party adapter maintainers to make sure they understand how to implement behavior flags in their own adapters as we will be asking them do so when appropriate.

mikealfare:
resolves #281

### Problem

Add the Behavior Flag Framework from `dbt-common`.

### Solution

Setup an empty collection of behavior flags and load them into the base adapter.

### Checklist


pyproject.toml
```diff
@@ -54,7 +54,7 @@ include = ["dbt/adapters", "dbt/include", "dbt/__init__.py"]
 
 [tool.hatch.envs.default]
 dependencies = [
-    "dbt_common @ git+https://github.com/dbt-labs/dbt-common.git",
+    "dbt_common @ git+https://github.com/dbt-labs/dbt-common.git@behavior-flags",
```
mikealfare, line 57: Reminder to revert this prior to merging

dbt/adapters/base/impl.py
```diff
@@ -291,6 +293,27 @@ def set_macro_context_generator(
...
+        flags.extend(self._behavior_extra)
+        try:
+            # we don't always get project flags, for example during `dbt debug`
+            self._behavior = Behavior(flags, self.config.flags)
+        except AttributeError:
+            # in that case, don't load any behavior to avoid unexpected defaults
+            self._behavior = Behavior([], {})
+
+    @property
+    def _behavior_extra(self) -> List[BehaviorFlag]:
```
colin-rogers-dbt, line 311: Not sure I have a great alternative here but instead of `_extra` could we be more specific? Maybe `_behavior_for_adapter` and then have a `_behavior_global`?

dbt/adapters/base/impl.py
```diff
@@ -291,6 +293,27 @@ def set_macro_context_generator(
     ) -> None:
         self._macro_context_generator = macro_context_generator
 
+    @property
+    def behavior(self) -> Behavior:
```
colin-rogers-dbt, line 297: Can this method be called before `_behavior` has been set?

colin-rogers-dbt (commented):
mostly fine just a couple questions

dbt/adapters/base/impl.py
mikealfare: I'll update this to `_behavior_flags`

tests/unit/conftest.py
mikealfare, line 1: Break this up into fixture files and then import those fixtures.

tests/unit/conftest.py
```diff
@@ -0,0 +1,220 @@
...
+    QueryComment,
+)
+
+
+@pytest.fixture
+def adapter(config, existing_relations, behavior_flags) -> BaseAdapter:
+
+    adapter = BaseAdapterStub(config, get_context("spawn"))
+
+    adapter.behavior = behavior_flags
```
mikealfare, line 29: Move this into BaseAdapterStub

colin-rogers-dbt (dismissed):

colin-rogers-dbt (approved):

PR #285, opened 2024-08-22
Add EventTimeFilter and BaseRelation.render_event_time_filtered

MichelleArk:
Resolves: https://github.com/dbt-labs/dbt-adapters/issues/294

* Introduces the `EventTimeFilter` class, which contains properties necessary to render a where filter on a given event time field
* `BaseRelation` has an optional `EventTimeFilter` attribute
* `BaseRelation.__str__` renders `EventTimeFilter`, if present to wrap the ref call in a `select * from ... where event_time >= lower_bound and event_time < upper_bound` 
   * if `limit` also provided, resolve limit first before event-time filter to prioritize performance of input data scans

Postgres testing: https://github.com/dbt-labs/dbt-postgres/actions/runs/10775130606


dbt/adapters/base/relation.py
```diff
@@ -36,6 +37,30 @@
 SerializableIterable = Union[Tuple, FrozenSet]
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class EventTimeFilter(FakeAPIObject, Hashable):
+    field_name: str
+    start: Optional[datetime] = None
+    end: Optional[datetime] = None
+
+    def render(self) -> str:
```
MichelleArk, line 46: Discussed w @mikealfare live today: we might want to move this method into the `BaseRelation` for ease of overwriting by adapter maintainers.

dbt/adapters/base/relation.py
```diff
@@ -36,6 +37,16 @@
...
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class EventTimeFilter(FakeAPIObject, Hashable):
+    field_name: str
+    start: Optional[datetime] = None
+    end: Optional[datetime] = None
+
+    def __hash__(self) -> int:
+        return hash(self.render())
```
MichelleArk, line 47: remove this?

dbt/adapters/base/relation.py
```diff
@@ -225,6 +239,31 @@ def render_limited(self) -> str:
...
+        """
+        Returns "" if start and end are both None
+        """
+        filter = ""
+        if event_time_filter.start and event_time_filter.end:
+            filter = f"{event_time_filter.field_name} >= '{event_time_filter.start}' and {event_time_filter.field_name} < '{event_time_filter.end}'"
+        elif event_time_filter.start:
+            filter = f"{event_time_filter.field_name} >= '{event_time_filter.start}'"
+        elif event_time_filter.end:
+            filter = f"{event_time_filter.field_name} < '{event_time_filter.end}'"
```
colin-rogers-dbt, line 263: Are we sure we want to mix inclusive/exclusive comparisons?

dbt/adapters/base/relation.py
```diff
@@ -36,6 +37,13 @@
 SerializableIterable = Union[Tuple, FrozenSet]
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class EventTimeFilter(FakeAPIObject, Hashable):
+    field_name: str
+    start: Optional[datetime] = None
```
colin-rogers-dbt, line 43: do we need to do any validation on start/end time?

dbt/adapters/base/relation.py
MichelleArk: I think so. The way we're going to be chunking up microbatch runs will be by partitions. For example, we'll be building daily partition `2020-01-01 ( 2020-01-01 > event_time <= 2020-01-02)`, followed by `2020-01-02 ( 2020-01-02 > event_time <= 2020-01-03)`, so these inclusive/exclusive semantics work very intuitively with that. 

https://github.com/dbt-labs/dbt-core/pull/10677 (check out the preliminary screenshots there!)

dbt/adapters/base/relation.py
MichelleArk: Beyond them being timestamps, I think we're good 👍 The core part of the framework considers the configured partition grain and plumbs the appropriate start/end bounds while constructing an `EventTimeFilter` to initialize a `Relation` with.

dbt/adapters/base/relation.py
```diff
@@ -240,6 +279,7 @@ def create_ephemeral_from(
         cls: Type[Self],
         relation_config: RelationConfig,
         limit: Optional[int] = None,
+        event_time_filter: Optional[EventTimeFilter] = None,
```
mikealfare, line 282: We need to make sure that whatever will be calling this will pin to this version of `dbt-adapters` so that we don't try to pass an extra kwarg to a prior version of `dbt-adapters`.

mikealfare (approved):

dbt/adapters/base/relation.py
MichelleArk: Yes! It will be in this PR: https://github.com/dbt-labs/dbt-core/pull/10594

Will do so once this is released in dbt-adapters.

cc @QMalcolm

PR #290, opened 2024-09-03
Snapshot column names

### LINKED ISSUE #289: Allow configuring snapshot column names
### Description

This is the dbt-adapters part of https://github.com/dbt-labs/dbt-core/issues/10185. We currently don't allow configuring the column names of snapshot tables.

### Acceptance Criteria

We allow configuring snapshot table column names.

gshank:
resolves #289

### Problem

We don't allow configuring snapshot column names.

### Solution

Add configuration of dbt_valid_to, dbt_valid_from, dbt_updated_at, and dbt_scd_id column names.

### Checklist


dbt/adapters/exceptions/compilation.py
```diff
@@ -150,7 +150,7 @@ def __init__(self, missing: List):
         super().__init__(msg=self.get_message())
 
     def get_message(self) -> str:
-        msg = 'Snapshot target is not a snapshot table (missing "{}")'.format(
+        msg = 'Snapshot target is not a snapshot table or is missing configured columns (missing "{}")'.format(
```
graciegoheen, line 153:
```suggestion
        msg = 'Snapshot target is missing meta-fields (missing "{}")'.format(
```
Maybe we also add a link to the docs? https://docs.getdbt.com/docs/build/snapshots#snapshot-meta-fields

dbt/adapters/base/impl.py
```diff
@@ -782,21 +783,15 @@ def valid_snapshot_target(self, relation: BaseRelation) -> None:
 
         columns = self.get_columns_in_relation(relation)
         names = set(c.name.lower() for c in columns)
-        expanded_keys = ("scd_id", "valid_from", "valid_to")
-        extra = []
+        columns = ("dbt_scd_id", "dbt_valid_from", "dbt_valid_to")
```
mikealfare, line 786: Redeclaring a variable can be confusing. Could we change the name of one of these? Or, given how short this is, maybe we just do it inline on 788:
```python
for column in ("dbt_scd_id", "dbt_valid_from", "dbt_valid_to"):
```

dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
```diff
@@ -24,7 +23,7 @@
   {{ run_hooks(pre_hooks, inside_transaction=True) }}
 
   {% set strategy_macro = strategy_dispatch(strategy_name) %}
-  {% set strategy = strategy_macro(model, "snapshotted_data", "source_data", config, target_relation_exists) %}
+  {% set strategy = strategy_macro(model, "snapshotted_data", "source_data", model['config'], target_relation_exists) %}
```
mikealfare, line 26: I commented on this in the `dbt-spark` PR as well; same question, what prompted the change?

dbt/include/global_project/macros/materializations/snapshots/strategies.sql
```diff
@@ -133,11 +135,11 @@
 {%- endmacro %}
 
 
-{% macro snapshot_check_strategy(node, snapshotted_rel, current_rel, config, target_exists) %}
-    {% set check_cols_config = config['check_cols'] %}
-    {% set primary_key = config['unique_key'] %}
-    {% set invalidate_hard_deletes = config.get('invalidate_hard_deletes', false) %}
-    {% set updated_at = config.get('updated_at', snapshot_get_time()) %}
+{% macro snapshot_check_strategy(node, snapshotted_rel, current_rel, model_config, target_exists) %}
```
mikealfare, line 138: It looks liked the kwarg was updated but the original name is still referenced below. Is that intentional?

dbt/adapters/base/impl.py
```diff
@@ -764,7 +763,9 @@ def get_missing_columns(
         return [col for (col_name, col) in from_columns.items() if col_name in missing_columns]
 
     @available.parse_none
-    def valid_snapshot_target(self, relation: BaseRelation) -> None:
+    def valid_snapshot_target(
```
mikealfare, line 766: Anything that calls this method with the new kwarg will need to bump its lower bound pin on `dbt-adapters`. That's fine, just making a note of it.

dbt/include/global_project/macros/materializations/snapshots/helpers.sql
```diff
@@ -34,7 +34,12 @@
   {{ adapter.dispatch('snapshot_staging_table', 'dbt')(strategy, source_sql, target_relation) }}
 {% endmacro %}
 
+{% macro get_snapshot_table_column_names() %}
```
mikealfare, line 37: Anything that calls this new macro will need to bump its lower bound pin on `dbt-adapters`. That's fine, just making a note of it.

mikealfare (commented):
One suggestion, a few points of clarification, and a few new features that will require a minor version bump. Otherwise it looks good.

dbt/include/global_project/macros/materializations/snapshots/helpers.sql
```diff
@@ -34,7 +34,12 @@
...
+    {{ return({'dbt_valid_to': 'dbt_valid_to', 'dbt_valid_from': 'dbt_valid_from', 'dbt_scd_id': 'dbt_scd_id', 'dbt_updated_at': 'dbt_updated_at'}) }}
+{% endmacro %}
+
 {% macro default__snapshot_staging_table(strategy, source_sql, target_relation) -%}
+    {% set stcn = config.get('snapshot_table_column_names') or get_snapshot_table_column_names() %}
```
colin-rogers-dbt, line 43: Instead of passing around `stcn`, which is tricky to grok it's semantic meaning as it's an acronym, would it be simpler to just assign them to variables immediately? 
```suggestion
{% set dbt_updated_at = stcn.dbt_updated_at %}
```

colin-rogers-dbt: Alternative we could just rename `stcn` to something like `columns`, I think it's reasonable to for the reader to infer these are snapshot columns from the context  of being in the `snapshot_staging_table` macro

dbt/include/global_project/macros/materializations/snapshots/strategies.sql
gshank: Do you mean the "model_config"?  Yeah, I left the param there for cases where somebody might have been using it, but renamed it to make it clear that it's not the same config that's used below.

dbt/include/global_project/macros/materializations/snapshots/snapshot.sql
gshank: I wanted to use a "property" from the config object, and the dictionary form that was being used had no access to it.

dbt/include/global_project/macros/materializations/snapshots/helpers.sql
gshank: There is code that uses all 4 of the column names (and there's another one coming...) so it feels more succinct to use the dictionary. The places where there's only one used are more arguable, but I'd rather do it the same in every place.

I've renamed stcn to "column".

dbt/adapters/base/impl.py
gshank: This has been changed.

dbt/adapters/base/impl.py
gshank: So the lower bound pin of dbt-spark and dbt-postgres can be updated in the appropriate pull requests. What about the adapters without a pr?  Like dbt-snowflake and dbt-redshift?  Their code doesn't call the new macro directly, but will call it via the dbt-adapters code.

dbt/adapters/exceptions/compilation.py
gshank: Added.

dbt/include/global_project/macros/materializations/snapshots/strategies.sql
VersusFacit: I can foresee a future in which the significance of this move is lost on whoever's reading this. Mind adding a comment that we left the model config to keep same interface or something to that effect?

gshank: I've added some comments where it's passed in and in the functions.

colin-rogers-dbt (approved):

PR #311, opened 2024-09-18
Better handle nested Array types

### LINKED ISSUE #310: [Bug] Support complex Array data types for macro get_columns_in_relation
### Is this a new bug?

### Current Behavior

If a given column is a complex ARRAY type such as `ARRAY(VARCHAR(16777216))`, the macro `get_columns_in_relation` raises an error.  

### Expected Behavior

We can better handle this case to not raise an error.

### Steps To Reproduce

Do a `dbt run` on a model which calls `adapter.get_columns_in_relation()` on a table with a column of type `ARRAY(VARCHAR(16777216))`.

### Relevant log output

```shell
Could not interpret data_type "ARRAY(VARCHAR(16777216))": could not convert "VARCHAR(16777216" to an integer
```

### Environment

```markdown
- OS:
- Python: 3.11.4
- dbt-adapters: 1.6.1
```

### Additional Context

Relevant code that can more gracefully handle this case -  

https://github.com/dbt-labs/dbt-adapters/blob/v1.6.1/dbt/adapters/base/column.py#L142-L146

jdistler:
resolves #310

### Problem

When a model uses `get_columns_in_relation` on a table which includes columns of complex array types like `ARRAY(VARCHAR(16777216))`, we raise an error because the pre-existing logic only expects there to be an int inside the parentheses. In this case we have nested types so `VARCHAR(16777216)` cannot be converted to an `int` and processing fails.

### Solution

Before we try to convert to an int, we run some new regex to check if this is a nested type, as in some alphabet chars followed by an open parentheses. If so, we do not try to extract a `char_size`, `numeric_precision`, or `numeric_scale` and instead allow the code to return with those variables left as `None`. 

In the case of ``VARCHAR(16777216)``, we would have  `data_type = ARRAY,  char_size = None, numeric_precision = None, numeric_scale = None`.

### Checklist

### DIFF
--- a/dbt/adapters/base/column.py
+++ b/dbt/adapters/base/column.py
@@ -135,29 +135,31 @@ class Column:
         if size_info is not None:
             # strip out the parentheses
             size_info = size_info[1:-1]
-            parts = size_info.split(",")
-            if len(parts) == 1:
-                try:
-                    char_size = int(parts[0])
-                except ValueError:
-                    raise DbtRuntimeError(
-                        f'Could not interpret data_type "{raw_data_type}": '
-                        f'could not convert "{parts[0]}" to an integer'
-                    )
-            elif len(parts) == 2:
-                try:
-                    numeric_precision = int(parts[0])
-                except ValueError:
-                    raise DbtRuntimeError(
-                        f'Could not interpret data_type "{raw_data_type}": '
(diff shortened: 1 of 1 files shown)


amychen1776:
Cloing this because we have completed this work

PR #323, opened 2024-10-01
Switch from environment variable to behavior flag for gating microbatch functionality

### LINKED ISSUE #327: [Microbatch] Use behavior flag to gate microbatch functionality
In #308 we began gating the microbatch functionality behind an environment variable. However, for a better user experience, we're moving to using a project flag in dbt-core (https://github.com/dbt-labs/dbt-core/pull/10799). In dbt-adapters, project flags can be interpreted by behavior flags. Thus we need to update dbt-adapters to use a behavior flag for gating the microbatch functionality instead of using an environment variable.

QMalcolm:
resolves #327

### Problem

We're currently gating microbatch functionality via environment variables, and we should instead use a behavior flag

### Solution

Switch to behavior flag for gating microbatch functionality

### Checklist

### DIFF
--- a/dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_microbatch.py
+++ b/dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_microbatch.py
@@ -61,14 +61,6 @@ class BaseMicrobatch:
 
         assert len(result) == expected_row_count, f"{relation_name}:{pformat(result)}"
 
-    @pytest.fixture(scope="class")
-    def project_config_update(self):
-        return {
-            "flags": {
-                "require_batched_execution_for_custom_microbatch_strategy": True,
-            }
-        }
-
     def test_run_with_event_time(self, project, insert_two_rows_sql):
         # initial run -- backfills all data
         with patch_microbatch_end_time("2020-01-03 13:57:00"):
--- a/dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_microbatch.py
+++ b/dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_microbatch.py
@@ -1,6 +1,4 @@
-import os
 from pprint import pformat
-from unittest import mock
 
 import pytest
 
@@ -63,7 +61,14 @@ class BaseMicrobatch:
 
         assert len(result) == expected_row_count, f"{relation_name}:{pformat(result)}"
 
-    @mock.patch.dict(os.environ, {"DBT_EXPERIMENTAL_MICROBATCH": "True"})
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            "flags": {
+                "require_batched_execution_for_custom_microbatch_strategy": True,
+            }
+        }
+
     def test_run_with_event_time(self, project, insert_two_rows_sql):
         # initial run -- backfills all data
         with patch_microbatch_end_time("2020-01-03 13:57:00"):
--- a/dbt/adapters/base/impl.py
+++ b/dbt/adapters/base/impl.py
@@ -23,7 +23,6 @@ from typing import (
     Union,
     TYPE_CHECKING,
 )
-import os
 import pytz
 from dbt_common.behavior_flags import Behavior, BehaviorFlag
 from dbt_common.clients.jinja import CallableMacroGenerator
@@ -316,7 +315,13 @@ class BaseAdapter(metaclass=AdapterMeta):
         """
         This method should be overwritten by adapter maintainers to provide platform-specific flags
         """
-        return []
+        return [
+            {
+                "name": "require_batched_execution_for_custom_microbatch_strategy",
+                "default": False,
+                "docs_url": "https://docs.getdbt.com/docs/build/incremental-microbatch",
+            }
+        ]
 
     ###
     # Methods that pass through to the connection manager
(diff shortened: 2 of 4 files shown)


dbt/adapters/base/impl.py
```diff
@@ -312,7 +311,13 @@ def _behavior_flags(self) -> List[BehaviorFlag]:
         """
         This method should be overwritten by adapter maintainers to provide platform-specific flags
         """
-        return []
+        return [
+            {
+                "name": "require_builtin_microbatch_strategy",
+                "default": False,
+                "description": "If True, then a builtin `microbatch` materialization strategy can be expected to exist.",
```
mikealfare, line 318: The description is something that winds up in the user-facing warning; it's intended to be an explanation for the user of why they're seeing the warning. This sounds more like a dev-facing docstring. I think it could be more useful to provide a `docs_url` to the microbatch docs on docs.getdbt.com. You could then either update this or leave it with the default, which would be "The behavior controlled by `require_builtin_microbatch_strategy` is currently turned off."

dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_microbatch.py
```diff
@@ -63,7 +61,14 @@ def assert_row_count(self, project, relation_name: str, expected_row_count: int)
 
         assert len(result) == expected_row_count, f"{relation_name}:{pformat(result)}"
 
-    @mock.patch.dict(os.environ, {"DBT_EXPERIMENTAL_MICROBATCH": "True"})
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
```
mikealfare, line 65: Is it worth testing what happens if the flag is turned off?

dbt/adapters/base/impl.py
QMalcolm: Goooootcha! Okay I'll update this to link to the documentation 🙂

dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_microbatch.py
QMalcolm: We can do that. We have tests in core I believe that check this, but it makes sense to have a test here as well 🙂

QMalcolm:
So it turns out, we actually need to do MUCH less. I'm going to drop a bunch of commits... Like half of them. And then also squash some of them. And then make some changes to simplify the tests further.

dbt/adapters/base/impl.py
```diff
@@ -1574,13 +1579,24 @@ def valid_incremental_strategies(self):
 
     def builtin_incremental_strategies(self):
         builtin_strategies = ["append", "delete+insert", "merge", "insert_overwrite"]
-        if os.environ.get("DBT_EXPERIMENTAL_MICROBATCH"):
+        if self.behavior.require_batched_execution_for_custom_microbatch_strategy:
```
mikealfare, line 1582: Adding `.no_warn` here would make it so that the warning doesn't fire on every incremental model. But then we would need to evaluate this deeper in the code.

dbt/adapters/base/impl.py
```diff
@@ -1574,13 +1579,24 @@ def valid_incremental_strategies(self):
...
             builtin_strategies.append("microbatch")
 
         return builtin_strategies
 
     @available.parse_none
     def get_incremental_strategy_macro(self, model_context, strategy: str):
+        """Gets the macro for the given incremental strategy.
```
mikealfare, line 1589: First, nice docs. We need to add these more often.

Per @jtcohen6's observation, at this point, we know what the user is trying to do and we know what the value of the behavior flag is. What if instead of adding `microbatch` in `builtin_incremental_strategies`, we add it conditionally here:
```python
    # validate strategies for this adapter
    valid_strategies = self.valid_incremental_strategies()
    valid_strategies.append("default")
    builtin_strategies = self.builtin_incremental_strategies()
    if strategy == "microbatch" and self.behavior.require_batched_execution_for_custom_microbatch_strategy:
        builtin_strategies.append("microbatch")
    if strategy in builtin_strategies and strategy not in valid_strategies:
        ...
```
Then the warning should only fire when the user has specified `microbatch`, but has not enabled the built-in microbatch strategy.

mikealfare (commented):
One recommendation that I think solves the "always warning" issue.

mikealfare (approved):

dbt-tests-adapter/dbt/tests/adapter/incremental/test_incremental_microbatch.py
```diff
@@ -63,7 +61,6 @@ def assert_row_count(self, project, relation_name: str, expected_row_count: int)
 
         assert len(result) == expected_row_count, f"{relation_name}:{pformat(result)}"
 
-    @mock.patch.dict(os.environ, {"DBT_EXPERIMENTAL_MICROBATCH": "True"})
```
MichelleArk, line 66: Nice! Let's make sure to apply this change to concrete adapters, I think it's just BigQuery. The existing tests shouldn't break if we left it in, so it's just mostly for cleanliness: 
* https://github.com/dbt-labs/dbt-bigquery/pull/1358/files#diff-933dcacdedf65d9337c2fd85863166cc878c4c8351b07a16d778bb99abaaaa95L33

(feel free to start a fresh dbt-bigquery PR if you'd like!)

MichelleArk:
Release notes: we'll need (patch?) releases of dbt-tests-adapter as well as `dbt-adapters` for these changes. Given microbatch is in beta status, it feels alright to just do patches here. What do you think?

QMalcolm:
> Release planning: we'll need (patch?) releases of dbt-tests-adapter as well as `dbt-adapters` for these changes. Given microbatch is in beta status, it feels alright to just do patches here. What do you think?

That feels right to me. @mikealfare Does that sound right?

PR #326, opened 2024-10-01
Allow use of multiple column unique keys in snapshots

### LINKED ISSUE #181: [Feature] Allow unique_key for snapshots to take a list

gshank:
resolves #181

### Problem

Implement use of a list for a unique_key in snapshots

### Solution

Check whether the unique_key is a list and modify the SQL.

### Checklist


dbt/include/global_project/macros/materializations/snapshots/strategies.sql
```diff
@@ -70,7 +70,16 @@
         ({{ snapshotted_rel }}.{{ columns.dbt_valid_from }} < {{ current_rel }}.{{ updated_at }})
     {%- endset %}
 
-    {% set scd_id_expr = snapshot_hash_arguments([primary_key, updated_at]) %}
+    {% if primary_key is iterable and primary_key is not string and primary_key is not mapping %}
```
mikealfare, line 73: I would make this its own macro as well. Are `iterable` and `sequence` different in jinja? Could we check both in a shared macro if they are?

dbt/include/global_project/macros/materializations/snapshots/strategies.sql
```diff
@@ -70,7 +70,16 @@
...
+        {% set scd_args = [] %}
+        {% for key in primary_key %}
+            {{ scd_args.append(key) }}
+        {% endfor %}
+        {{ scd_args.append(updated_at) }}
+        {% set scd_id_expr = snapshot_hash_arguments(scd_args) %}
+    {% else %}
+        {% set scd_id_expr = snapshot_hash_arguments([primary_key, updated_at]) %}
```
mikealfare, line 81: I would only set the variable parts of `scd_args` in the if block and then move the call to `snapshot_hash_arguments` and the inclusion of `updated_at` outside of the if block:

```sql
    {% set scd_args = [updated_at] %}
    {% if primary_key is iterable and primary_key is not string and primary_key is not mapping %}
        {% for key in primary_key %}
            {{ scd_args.append(key) }}
        {% endfor %}
    {% else %}
        {{ scd_args.append(primary_key) }}
    {% endif %}
    {% set scd_id_expr = snapshot_hash_arguments(scd_args) %}
```

Or even better, I would make this a python method since there is no templating:

```sql
    {% set scd_args = api.Relation.scd_args(primary_key, updated_at) %}
    {% set scd_id_expr = snapshot_hash_arguments(scd_args) %}
```

mikealfare (commented):
I added a few suggestions focusing on minimizing the amount of jinja that we have and making that jinja more readable. Let me know what you think.

dbt/include/global_project/macros/materializations/snapshots/helpers.sql
```diff
@@ -50,7 +50,14 @@
     snapshotted_data as (
 
         select *,
+        {% if strategy.unique_key | is_list %}
```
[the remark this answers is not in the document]
gshank: Done.

dbt/include/global_project/macros/materializations/snapshots/helpers.sql
```diff
@@ -100,13 +126,28 @@
...
         from insertions_source_data as source_data
-        left outer join snapshotted_data on snapshotted_data.dbt_unique_key = source_data.dbt_unique_key
-        where snapshotted_data.dbt_unique_key is null
-           or (
-                snapshotted_data.dbt_unique_key is not null
-            and (
-                {{ strategy.row_changed }}
-            )
+        left outer join snapshotted_data on
+            {% if strategy.unique_key | is_list %}
```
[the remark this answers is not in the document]
gshank: Much better. Done.

dbt/include/global_project/macros/materializations/snapshots/strategies.sql
gshank: This just wasn't updated to use the new 'is_list' filter that I added in dbt-common.

dbt/include/global_project/macros/materializations/snapshots/strategies.sql
gshank: Much better. Done.

mikealfare (approved):
Every time we remove jinja, an angel gets its wings.

PR #328, opened 2024-10-16
Add adapter telemetry method for core model run snowplow event.

### LINKED ISSUE #301: [Feature] Adapters Model Configuration Snowplow Tracking

- [X] I am requesting a straightforward extension of existing dbt-adapters functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Since it's inception dbt Core has not given a great deal of thought to adapter specific telemetry. This poses a challenge for us in tracking feature adoption and assessing the impact/risk of making changes. 

What we propose is to expose a new method in the base adapter that accepts a `RelationConfig` and returns an object that conforms to`RelationTrackingInfo` protocol (tentative name) which details the standard set of fields we want to know about _all_ relations as well as the adapter specific information relevant to that relation. By default this method would return just the basic fields and individual adapters could override to return more specific information. 

This method would be invoked [here](https://github.com/dbt-labs/dbt-core/blob/1d3d315249c81dbca70452e5a14fbfd0edad6b2a/core/dbt/task/run.py#L118C5-L118C30) by dbt-core so this does not require direct integration with snowplow itself. 

Next step is to flesh the design of this out and implement a rough proof-of-concept. 

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

VersusFacit:
resolves #301 

### Problem

We need a way to expose adapter telemetry for our internal product adoption analytics.

### Solution

The base adapter provides a simple method which is used in dbt core and its tracking system. This method will be overwritten for each adapter that we wish to collect telemetry for.

### Checklist


VersusFacit:
Proof of operation for fallback method (i.e. adapter does not override)

image

Note adapter_details is `{}`. It'll be non-empty in, say, snowflake ([here](https://github.com/dbt-labs/dbt-snowflake/pull/1209#issuecomment-2421531778)) which provides an override with snowflake-specific info. This field is _optional_ in our JSON schema.

dbt/adapters/base/relation.py
```diff
@@ -531,3 +531,10 @@ def flatten(self, allow_multiple_databases: bool = False) -> "SchemaSearchMap":
             )
 
         return new
+
+
+@dataclass(frozen=True, eq=False, repr=False)
+class AdapterTrackingRelationInfo(FakeAPIObject, Hashable):
```
colin-rogers-dbt, line 537: This is an interesting case to me: should this actually be considered a "contract" like `AdapterResponse` or is it tied to the concept of a relation (i.e. the thing it's tracking)? If the former we should probably put it in it's own module in `dbt/adapters/contracts`. I think there's an argument for both so curious what others think

colin-rogers-dbt:
Is this unit test-able?

dbt/adapters/base/relation.py
VersusFacit: So just putting this forward -- I originally had it as a contract. Then it ran into "can't instantiate protocol" errors. So I'm not sure it should be a contract on that alone :p

dbt/adapters/base/impl.py
```diff
@@ -1743,6 +1746,15 @@ def capabilities(cls) -> CapabilityDict:
     def supports(cls, capability: Capability) -> bool:
         return bool(cls.capabilities()[capability])
 
+    @available
```
mikealfare, line 1749: Does this really need to be available in the jinja context? I feel like we only need this for `dbt-core` to read.

dbt/adapters/base/impl.py
```diff
@@ -1743,6 +1746,15 @@ def capabilities(cls) -> CapabilityDict:
...
+    @classmethod
+    def get_adapter_run_info(cls, config: RelationConfig) -> AdapterTrackingRelationInfo:
```
mikealfare, line 1751: `RelationConfig` doesn't feel like the correct type. I would have guessed this is some version of `NodeConfig`.

dbt/adapters/base/impl.py
```diff
@@ -1743,6 +1746,15 @@ def capabilities(cls) -> CapabilityDict:
...
+        return AdapterTrackingRelationInfo(
```
mikealfare, line 1752: I could see wanting to return both the `BaseAdapter` version as well as the concrete adapter version. Can we add that in? It's also possible to predict both `adapter_name` and the path to the adapter version (`dbt.adapters.<adapter_name>.__version__.version`) as we do in `dbt.adapters.factory`. Could we fill that in for concrete adapters? And how would we extend this to combine the two? For example, if we had this method plus an adapter specific one, then we could merge them together:

```python
@classmethod
def get_adapter_run_info(cls, config) -> AdapterTrackingRelationInfo:
    """
    Interface method for collecting adapter metadata for tracking. This method should not be overwritten; instead, `self._get_adapter_run_info()` should be overwritten.
    """
    return AdapterTrackingRelationInfo(
        adapter_name=self.type,
        adapter_version=import_module(f"dbt.adapters.dbt_{self.type}.__version__).version,
        base_adapter_version=import_module(f"dbt.adapters.__about__).version,
        adapter_details=self._get_adapter_run_info(config),
    )

@classmethod
def _get_adapter_run_info(cls, config) -> Dict[str, Any]:
    """
    Adapter maintainers should overwrite this method to return any run metadata that should be captured during a run.
    """
    return {}
```

dbt/adapters/base/impl.py
VersusFacit: Ooop, nope, auto copy :p

dbt/adapters/base/impl.py
VersusFacit: I don't have a good handle on the differences here. When I did introspection, we are passing in a `RelationConfig`. I bet I'm missing some appreciation of all these different Config objects and perhaps I was "lead on" by the suggestion for `RelationConfig` in the original ticket. If you feel strongly on this type, please do offer your thoughts here!

colin-rogers-dbt: Since this will only be called in the context of a model it should always conform to `RelationConfig`

colin-rogers-dbt (approved):

mikealfare (dismissed):

PR #331, opened 2024-10-16
Always ensure valid incremental strategy

### LINKED ISSUE #330: [Bug] On first-run/full-refresh the incremental strategy of a model isn't validated
Presently, if the relation for an incremental model is being built for the first time (because the relation in the data warehouse doesn't exist) or if the incremental model is being run as a full refresh, then validation of the incremental strategy doesn't occur. Technically, for most incremental strategies, this is "okay". That is because the incremental strategy isn't used when creating the relation for the first time or during a full refresh. The exception to that is for microbatch incremental models. Additionally, this leads to an experience where if you write a new incremental model (with an invalid strategy) and then run it to ensure it works, everything will seem fine. You won't know anything is wrong until it is run for the second time. This can be a jarring experience, and is a bug.

QMalcolm:
resolves #330

### Problem

Previously if you set an invalid `incremental_strategy` for an incremental model, it'll succeed on the first run, and then fail on the second run
I.e. if you have
```
{{config(materialized="incremental", incremental_strategy="bad_strategy")}}

select * from {{ ref("input_model") }}

{% if is_incremental() %}
where event_time >= (select coalesce(max(event_time),'1900-01-01') from {{ this }} )
{% endif %}
```

That was always succeeding when the model was either built for the first time or rebuilt via --full-refresh. You only got an error about the strategy the next time the model was run.

### Solution

Change the order of operations in `incremental.sql` to ensure that a incremental model's `incremental_strategy` is _always_ validated

### Checklist


dbt-tests-adapter/dbt/tests/adapter/basic/test_incremental.py
```diff
@@ -86,9 +86,37 @@ def test_incremental_not_schema_change(self, project):
...
+
 class Testincremental(BaseIncremental):
     pass
 
 
 class TestBaseIncrementalNotSchemaChange(BaseIncrementalNotSchemaChange):
     pass
+
+
+class TestBaseIncrementalBadStrategy(BaseIncrementalBadStrategy):
```
mikealfare, line 121: We don't need this anymore. This used to be here because this package used to also serve as the integration tests for `dbt-postgres`, which is now it's own repo.

dbt-tests-adapter/dbt/tests/adapter/basic/test_incremental.py
```diff
@@ -86,9 +86,37 @@ def test_incremental_not_schema_change(self, project):
...
+    def seeds(self):
+        return {"base.csv": files.seeds_base_csv, "added.csv": files.seeds_added_csv}
+
+    def test_incremental_invalid_strategy(self, project):
+        # seed command
+        results = run_dbt(["seed"])
+        assert len(results) == 2
+
+        # try to run the incremental model, it should fail on the first attempt
+        results = run_dbt(["run"], expect_pass=False)
```
mikealfare, line 110: Should we make sure that it raises the correct exception?

PR #338, opened 2024-10-24
[Bug] Behavior flags should take the default value when the project file is not loaded

mikealfare:
### Problem

We don't always get the `flags` property on the `config` object that gets used to create a `BaseAdapter` instance. This happens whenever `project.yml` is not loaded. In particular, this happens during `dbt debug`, which is meant to be a quick check of whether we can connect to a database; this command only requires `profile.yml`.

The decision was initially made to not load any behavior flags in the scenarios where `project.yml` was not loaded. The rationale there was that we did not want to return one value for a behavior flag when `project.yml` is not loaded (hence the default) and another value for the behavior flag when `project.yml` is loaded (the default, potentially overridden by the user). However, in the former scenario, referencing the behavior flag at all results in a `CompilationError` since the flag does not exist on `BaseAdapter.behavior`. This is not desirable behavior either.

### Solution

While the behavior flags are loaded onto `BaseAdapter.behavior` dynamically, the object should always have the same flags available. Since the user's overrides are not available when `project.yml` is not loaded, we should act as if `project.yml` was loaded, but without a `flags` node. In other words, we should simply assume the default values when we don't get a `flags` attribute.

### Checklist


tests/unit/behavior_flag_tests/test_empty_project.py
```diff
@@ -0,0 +1,87 @@
...
+        },
+        {
+            "name": "default_true_user_skip_flag",
+            "default": True,
+            "description": "This is a true flag.",
+        },
+    ]
+
+
+def test_register_behavior_flags(adapter):
```
mikealfare, line 76: This is the same test as in `./test_behavior_flags.py`, but with `flags` removed from the `config` fixture above, and swapping the expected values for the two cases where the user override changes the result.

dbt/adapters/base/impl.py
```diff
@@ -300,12 +300,13 @@ def behavior(self) -> Behavior:
...
-            # we don't always get project flags, for example during `dbt debug`
-            self._behavior = Behavior(flags, self.config.flags)
-        except AttributeError:
-            # in that case, don't load any behavior to avoid unexpected defaults
-            self._behavior = Behavior([], {})
+
+        # we don't always get project flags, for example, the project file is not loaded during `dbt debug`
+        # in that case, load the default values for behavior flags to avoid compilation errors
+        # this mimics not loading a project file, or not specifying flags in a project file
+        user_overrides = getattr(self.config, "flags", {})
```
mikealfare, line 307: This is effectively the same as swapping `[]` for `flags` on line `308`, but doing this reads much more cleanly.

colin-rogers-dbt (approved):

PR #339, opened 2024-10-29
feat(python): allows to use prj. vars

### LINKED ISSUE #5617: [Feature] Support dbt.var + dbt.env_var in Python models

### Describe the feature

Copying from [thread in #beta-feedback-python-models](https://getdbt.slack.com/archives/C03QUA7DWCW/p1659470042461869):

> A quick thought, why not expose a `dbt.var` function (and similar for env var) rather than overloading config? If you apply the same limitations as for config function (ie. Only literal args) should be pretty straightforward to statically analyse which vars need to be injected?

It's true! The mechanism for `dbt.var` and `dbt.env_var` could be just the same as `dbt.config.get`. We detect the literal arguments passed in:
https://github.com/dbt-labs/dbt-core/blob/eb72dbf32aa96394d599c28db439e0b2c6fcc135/core/dbt/parser/models.py#L113-L117

https://github.com/dbt-labs/dbt-core/blob/eb72dbf32aa96394d599c28db439e0b2c6fcc135/core/dbt/parser/models.py#L182-L184

And then provide those values at "compile" time (which we're currently doing via Jinja... gross), as new methods (`var` + `env_var`) on the `dbtObj` class:

https://github.com/dbt-labs/dbt-core/blob/eb72dbf32aa96394d599c28db439e0b2c6fcc135/core/dbt/include/global_project/macros/python_model/python.sql#L33-L40

https://github.com/dbt-labs/dbt-core/blob/eb72dbf32aa96394d599c28db439e0b2c6fcc135/core/dbt/include/global_project/macros/python_model/python.sql#L71-L77

Not doing this, and continuing to ask users (as we are currently) to pass those vars + env_vars in as configs in `.yml` files:
```yml
version: 2

models:
  - name: my_python_model
    config:
      materialized: table
      target_name: "{{ target.name }}"
      specific_var: "{{ var('SPECIFIC_VAR') }}"
      specific_env_var: "{{ env_var('SPECIFIC_ENV_VAR') }}"
```

### Who will this benefit?

Usability and ergonomics for people developing and configuring Python models, especially ones with more advanced conditional logic or env-specific behavior

_No response_

### Anything else?

To be clear, the environment variable values would need to be the ones evaluated during _dbt parsing_. (They are NOT the environment variables set in the runtime environment, e.g. Snowpark.)

That's slightly different from how environment variables work in SQL models today, since the `{{ env_var() }}` Jinja is re-rendered at compile/execute runtime.

But I think it probably _should_ work this way across the board, where all inputs (including env vars) are resolved during parsing and pre-execution.

### LINKED ISSUE #10914: Create dbt.var.get() for python models to retrieve project variables without a yaml file

### Describe the feature

Create dbt.var.get() for python models to retrieve project variables without requiring yaml file.

I am narrowly requesting a var feature. However, this can be a small part of a larger initiative to get python at parity with SQL. 

Here is a [loom](https://www.loom.com/share/95f800d2d2a34541b0a62f77a35bf55a?sid=d79d9226-b76e-41c2-aedb-4c83d8fbd412) walking through the today solution.

The tl;dr is that python models don't have an argument such as dbt.var.get(). We have a dbt.config.get() but that argument is specific to each model (not global). 
**What we need to do today:** 
1. Give the model a passthrough variable for example, `run_frequency`. 
2. Make a model specific yaml for your python model. This is really we need to do all python model configs if we want SQL parity. 
3. Define the default you want, in this case `'Daily'  `in the `dbt_project.yml`. 
4. Pass the variable override in the command line as usual, in our example, `'Weekly'`. So `dbt run -s model_name  --vars '{"run_frequency": "Weekly"}'`

### Who will this benefit?

Our user base that is using python. I realize this is not the majority of users, but it would boost the python experience up to reduce the user experience and ergonomics divide between python and SQL in dbt. 

Yes, I can be a tester. 

### Anything else?

_No response_

devmessias:
resolves #
Resolves https://github.com/dbt-labs/dbt-core/issues/5617, Resolves https://github.com/dbt-labs/dbt-core/issues/10914

### Problem
Python models are not friendly when we need to use project variables. This will allow to use project variables without to define that inside the model.yaml 

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20241031-160641.yaml
+++ b/.changes/unreleased/Features-20241031-160641.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: allows to use prj. vars in python models
+time: 2024-10-31T16:06:41.214334214-03:00
+custom:
+    Author: devmessias
+    Issue: 5617 10914
--- a/dbt/include/global_project/macros/python_model/python.sql
+++ b/dbt/include/global_project/macros/python_model/python.sql
@@ -58,6 +58,21 @@ def source(*args, dbt_load_df_function):
 config_dict = {{ config_dict }}
 {% endmacro %}
 
+
+{% macro build_var_dict(model, context) %}
+    {%- set var_dict = {} -%}
+    {% set var_dbt_used = zip(model.config.var_keys_used, model.config.var_keys_defaults) | list %}
+    {%- for key, default in var_dbt_used -%}
+        {# weird type testing with enum, would be much easier to write this logic in Python! #}
+        {%- if key == "language" -%}
+          {%- set value = "python" -%}
+        {%- endif -%}
+        {%- set value = context.var(key, default) -%}
+        {%- do var_dict.update({key: value}) -%}
+    {%- endfor -%}
+var_dict = {{ var_dict }}
+{% endmacro %}
+
 {% macro py_script_postfix(model) %}
 # This part is user provided model code
 # you will need to copy the next section to run the code
@@ -67,6 +82,7 @@ config_dict = {{ config_dict }}
 {{ build_ref_function(model ) }}
 {{ build_source_function(model ) }}
 {{ build_config_dict(model) }}
+{{ build_var_dict(model, context) }}
 
 class config:
     def __init__(self, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


Gilwe:
More than a year past and there's still no way to get variable in a python model in bigquery

Gilwe:
@devmessias are you still maintaining this repo? if so, can you check this one out please?

PR #348, opened 2024-11-12
Fix setting of global behavior flags

### LINKED ISSUE #351: [Bug] Global behavior flags are getting overwritten
### Is this a new bug?

### Current Behavior

Currently, we set the baseline expected behavior flags in the `_behavior_flags` property method of `BaseAdapter`. However, this property method is _expected_ to be overwritten by adapter implementations. For adapters that currently do this (currently `dbt-snowflake` and `dbt-redshift`), things break when those flags are used. This is exasperated by the fact that some of the inherited methods for implementations like `builtin_incremental_strategies` try to lookup behavior flags. This results in an error message like
```
20:24:14  Encountered an error while running operation: Compilation Error
  The flag require_batched_execution_for_custom_microbatch_strategy has not be registered.
  
  > in macro materialization_incremental_snowflake (macros/materializations/incremental.sql)
  > called by <Unknown>
```

### Expected Behavior

The baseline expected behavior flags shouldn't be so easily blown away.

### Steps To Reproduce

1. Grab `dbt-snowflake` or `dbt-databricks`
2. Create an incremental model in your project
3. run `dbt run`

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-adapters:
```

### Additional Context

_No response_

MichelleArk:
resolves #351 

### Problem

As part of https://github.com/dbt-labs/dbt-adapters/pull/323, we had introduced the `require_batched_execution_for_custom_microbatch_strategy` behaviour flag as part of the `BaseAdapter._behavior_flags` method. However, any concrete adapters that overwrote this property effectively removed the new behaviour flag from the collection, which would lead to downstream CompilationErrors (behaviour flag not registered).

### Solution

Set up `require_batched_execution_for_custom_microbatch_strategy` as a global behaviour flag.

Also added a test to ensure BaseAdapter._behaviour_flags is always empty to avoid this issue in the future.

### Checklist

### DIFF
--- a/dbt/adapters/base/impl.py
+++ b/dbt/adapters/base/impl.py
@@ -262,6 +262,13 @@ class BaseAdapter(metaclass=AdapterMeta):
     }
 
     MAX_SCHEMA_METADATA_RELATIONS = 100
+    GLOBAL_BEHAVIOR_FLAGS = [
+        {
+            "name": "require_batched_execution_for_custom_microbatch_strategy",
+            "default": False,
+            "docs_url": "https://docs.getdbt.com/docs/build/incremental-microbatch",
+        }
+    ]
 
     # This static member variable can be overridden in concrete adapter
     # implementations to indicate adapter support for optional capabilities.
@@ -274,7 +281,7 @@ class BaseAdapter(metaclass=AdapterMeta):
         self._macro_resolver: Optional[MacroResolverProtocol] = None
         self._macro_context_generator: Optional[MacroContextGeneratorCallable] = None
         # this will be updated to include global behavior flags once they exist
-        self.behavior = []  # type: ignore
+        self.behavior = self.GLOBAL_BEHAVIOR_FLAGS  # type: ignore
 
     ###
     # Methods to set / access a macro resolver
--- a/dbt/adapters/base/impl.py
+++ b/dbt/adapters/base/impl.py
@@ -321,6 +321,8 @@ class BaseAdapter(metaclass=AdapterMeta):
     def _behavior_flags(self) -> List[BehaviorFlag]:
         """
         This method should be overwritten by adapter maintainers to provide platform-specific flags
+
+        The BaseAdapter should NOT include any global flags here as should be defined in BaseAdapter.GLOBAL_BEHAVIOR_FLAGS
         """
         return []
 
--- a/dbt/adapters/base/impl.py
+++ b/dbt/adapters/base/impl.py
@@ -322,13 +322,7 @@ class BaseAdapter(metaclass=AdapterMeta):
         """
         This method should be overwritten by adapter maintainers to provide platform-specific flags
         """
-        return [
-            {
-                "name": "require_batched_execution_for_custom_microbatch_strategy",
-                "default": False,
-                "docs_url": "https://docs.getdbt.com/docs/build/incremental-microbatch",
-            }
-        ]
+        return []
 
     ###
     # Methods that pass through to the connection manager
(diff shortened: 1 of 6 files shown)


MichelleArk:
Testing against dbt-snowflake: https://github.com/dbt-labs/dbt-snowflake/actions/runs/11806247075

MichelleArk:
dbt-postgres testing: https://github.com/dbt-labs/dbt-postgres/actions/runs/11806285678

dbt/adapters/base/impl.py
```diff
@@ -314,14 +321,10 @@ def behavior(self, flags: List[BehaviorFlag]) -> None:
     def _behavior_flags(self) -> List[BehaviorFlag]:
         """
         This method should be overwritten by adapter maintainers to provide platform-specific flags
+
+        The BaseAdapter should NOT include any global flags here as should be defined in BaseAdapter.GLOBAL_BEHAVIOR_FLAGS
```
mikealfare, line 325: We should update this based on the chosen approach (module variable or directly in the init.

dbt/adapters/base/impl.py
```diff
@@ -274,7 +281,7 @@ def __init__(self, config, mp_context: SpawnContext) -> None:
         self._macro_resolver: Optional[MacroResolverProtocol] = None
         self._macro_context_generator: Optional[MacroContextGeneratorCallable] = None
         # this will be updated to include global behavior flags once they exist
-        self.behavior = []  # type: ignore
+        self.behavior = self.GLOBAL_BEHAVIOR_FLAGS  # type: ignore
```
mikealfare, line 284: The flags should just be put directly here, or as a module level attribute and included here. We should avoid adding them  as a class variable; this would avoid confusion of `self.behavior` versus `self.GLOBAL_BEHAVIOR_FLAGS`.

MichelleArk: > or as a module level attribute and included here

Sorry I'm not following this suggestion, could you please elaborate?

mikealfare: I would either move `GLOBAL_BEHAVIOR_FLAGS` up to line 101 or I would define the behavior flags directly:
```python
def __init__(self, config, mp_context: SpawnContext) -> None:
    ...
    self.behavior = [
        {
            "name": "require_batched_execution_for_custom_microbatch_strategy",
            "default": False,
            "docs_url": "https://docs.getdbt.com/docs/build/incremental-microbatch",
        }
    ]
```
This reduces the number of class variables on `BaseAdapter` and avoids folks accidentally using `self.GLOBAL_BEHAVIOR_FLAGS`, which would be incorrect (they don't account for user settings, they won't fire warnings, etc.).

QMalcolm: I'm picking up what you're putting down. I think we should keep it as a variable, but move it outside the class as you suggested. My worry with it being an inline specification, that is that if anyone is overriding the `__init__` in their implementation, they might need to pull in what the default builtins behavior flags are.

I also might rename it to something like `DEFAULT_BUILTIN_BEHAVIOR_FLAGS` as well to be more clear. Someone might misunderstand the `GLOBAL_BEHAVIOR_FLAGS` name and think it's a global accessor or something.

QMalcolm: Updated!

dbt/adapters/base/impl.py
QMalcolm: Updated!

mikealfare (approved):

PR #356, opened 2024-11-16
ADAP-1135: Add `src` directory structure for `dbt-tests-adapter`

mikealfare:
As part of moving to a monorepo, we have decided to adopt the `src` namespace approach for storing functional code.

While updating `pyproject.toml` with these changes, I noticed that we use `__about__.py` to store the package version. However, other packages use `__version__.py`, and we are locked into this because this version is read by other packages. We should update `dbt-tests-adapter` to use `__version__.py` to standardize.


dbt-tests-adapter/pyproject.toml
mikealfare, line 1: This is the only file with actual changes; the rest are moves. These changes reflect those file movements.

Issue #360, opened 2024-11-20
[Microbatch] Update default `make_temp_relation` macro to incorporate a batch specific identifier if available


MichelleArk:
## Desired Improvement

Microbatch batches have been improved in core such that they _may_ be run concurrently. When a batch is executed, the necessary data is first moved into a `temp_relation` in the data warehouse. The path for this temp_relation is `<resource_identifier>__dbt_tmp`. As it currently stands, the `<resource_identifier>__dbt_tmp` will be the same for each batch of a given microbatch model. If the batches are run concurrently, they may end up clobbering the temp_relation destination, which may lead to some wonkiness. As such, we need a way to ensure that each batch for a microbatch model gets a unique `temp_relation` path.

## Helpful Prior Art

Similar to dbt-postgres: https://github.com/dbt-labs/dbt-postgres/blob/ae48e67dae6c1b00cda37ee9bdc61d3330506638/dbt/include/postgres/macros/adapters.sql#L149-L152

vanAkim:
Hello,

what a shame that I didn't find this issue before learning how the source code works, and found that my problem could be solved by https://github.com/dbt-labs/dbt-adapters/pull/361. I was currently starting the same PR.
At least, I've learned a lot and waiting to this release.

However, let me give you a little more context, because this solves my issue, but it's not related to microbatches. I'm currently implementing dbt with Airflow, and we are used to having multiples runs to ingest data in different partitions of BigQuery table.
Thus, having a dbt model `model_A`, materialized as `incremental` with `insert_overwrite` strategy, launching multiples runs, let's say D-2 and D-1, for different partitions, will create a conflict with `model_A__dbt_tmp` table. And the `dbt-bigquery` don't leverage this situation and relies on the `make_temp_relation` on `dbt-adapters`. So, it's the exact same situation as microbatch but on a less granular scope.

You'll probably know and understand better than me, but I wanted to share my situation.

PR #361, opened 2024-11-20
Append `model.batch.id` to suffix in `make_temp_relation` when present

QMalcolm:
resolves https://github.com/dbt-labs/dbt-adapters/issues/360

### Problem

Batches for a microbatch model were all getting the same temp relation name. This is fine if batches are running sequentially. However, we're enabling concurrent batch execution support. In which case, if the batches have the same temp relation name, they may clobber each other and result in "bad things"™️ 

### Solution

Append the `model.batch.id` to the suffix of the temporary relation name when present

### Checklist

### DIFF
--- a/dbt/include/global_project/macros/adapters/relation.sql
+++ b/dbt/include/global_project/macros/adapters/relation.sql
@@ -7,6 +7,11 @@
 {% endmacro %}
 
 {% macro make_temp_relation(base_relation, suffix='__dbt_tmp') %}
+  {#-- This ensures microbatch batches get unique temp relations to avoid clobbering --#}
+  {% if suffix == '__dbt_tmp' and model.batch %}
+    {% set suffix = suffix ~ '_' ~ model.batch.id %}
+  {% endif %}
+
   {{ return(adapter.dispatch('make_temp_relation', 'dbt')(base_relation, suffix)) }}
 {% endmacro %}
 


dbt/include/global_project/macros/adapters/relation.sql
```diff
@@ -7,6 +7,11 @@
 {% endmacro %}
 
 {% macro make_temp_relation(base_relation, suffix='__dbt_tmp') %}
+  {#-- This ensures microbatch batches get unique temp relations to avoid clobbering --#}
+  {% if suffix == '__dbt_tmp' and model.batch %}
```
mikealfare, line 11: Is there a reason to limit this to a particular suffix? Or can we just look for the existence of a batch configuration?

QMalcolm: My thought process was that if a specific suffix is being passed in, we probably don't want to mess with it, as we don't know if doing so is safe or not.

MichelleArk: If we don't limit to the `__dbt_tmp` suffix, it won't be possible for users to provide their own custom suffix for microbatch models. It's unclear whether that's actually desirable or not. If we're saying that microbatch temp tables should always be suffixed with a batch-specific id, then we shouldn't restrict to a particular suffix. If we want to give users the flexibility to fully manage their own temp table names, even in microbatch, then Quigley's current solution is more appropriate.

mikealfare: > If we want to give users the flexibility to fully manage their own temp table names, even in microbatch, then Quigley's current solution is more appropriate.

This is the distinction I'm looking for. If that's the intention, and we don't need to do this for other built-in suffixes (e.g. `__dbt_backup`), then this is fine. I suspect we may want to broaden this in the future to some type of "dbt-managed" versus "non-dbt-managed", but that's well outside of the scope of this PR.

mikealfare (approved):

PR #368, opened 2024-12-04
support retryable exceptions during query execution

colin-rogers-dbt:
- support retryable exceptions during query execution

resolves #

### Checklist


dbt/adapters/sql/connections.py
```diff
@@ -199,3 +211,46 @@ def commit(self):
...
         return connection
+
+    def _retryable_cursor_execute(self,
+                                  execute_fn: Callable,
+                                  sql: str,
+                                  bindings: Optional[Any] = None,
+                                  retryable_exceptions: Iterable[Type[Exception]] = [],
+                                  retry_limit: int = 1,
+                                  retry_timeout: Union[Callable[[int], SleepTime], SleepTime] = 1,
+                                  _attempts: int = 0,
```
mikealfare: This feels like an early abstraction. Can we start with a timeout that's just an integer and go from there?

dbt/adapters/sql/connections.py
```diff
@@ -199,3 +211,46 @@ def commit(self):
...
+                                  ) -> None:
+        timeout = retry_timeout(_attempts) if callable(retry_timeout) else retry_timeout
+        if timeout < 0:
```
mikealfare, line 238: This feels like java style programming, aka "look before you leap". If we remove the callable variant of the timeout, can we omit this as well? In that scenario, I think it's either our default, or the user setting, both of which can be reasonably assumed to be non-negative.

dbt/adapters/sql/connections.py
```diff
@@ -199,3 +211,46 @@ def commit(self):
...
+            raise DbtRuntimeError(
+                "retry_timeout cannot be negative or return a negative time."
+            )
+
+        try:
+            execute_fn(sql, bindings)
+
+        except tuple(retryable_exceptions) as e:
```
mikealfare, line 244: This also feels like early abstraction. Can we update the type to be `Tuple` from start since the user of this is an adapter maintainer and not an end user?

dbt/adapters/sql/connections.py
```diff
@@ -199,3 +211,46 @@ def commit(self):
...
+                            f"{retry_limit} attempts remaining. Retrying in {timeout} seconds.\n"
+                            f"Error:\n{e}"
+                )
+            )
+
+            time.sleep(timeout)
+            return self._retryable_cursor_execute(
+                execute_fn=execute_fn,
+                sql=sql,
+                retry_limit=retry_limit - 1,
```
mikealfare, line 260: We already decremented the retry limit on line 234. This does it twice.

dbt/adapters/sql/connections.py
```diff
@@ -199,3 +211,46 @@ def commit(self):
...
+                retry_timeout=retry_timeout,
+                retryable_exceptions=retryable_exceptions,
+                _attempts=_attempts + 1,
```
mikealfare, line 263: I see why we would pass in `_attempts` now, but this could be a separate concept, similar to the way BQ manages retries. I still think it's an early abstraction and we could update this with a more intelligent method in a future update.

mikealfare (commented):
A few style nits and one actual blocking change (the double decrement for retries).

dbt/adapters/sql/connections.py
VersusFacit: What does removing the callable variant of the timeout mean here? We're passing in a function that the downstream user implements and I personally jive with that

dbt/adapters/sql/connections.py
```diff
@@ -61,7 +72,50 @@ def add_query(
...
+                )
+                time.sleep(1)
+
+                return _execute_query_with_retry(
+                    cursor=cursor,
+                    sql=sql,
+                    bindings=bindings,
+                    retryable_exceptions=retryable_exceptions,
+                    retry_limit=retry_limit,
+                    attempt=1,
```
mikealfare, line 116: I think this is supposed to be `attempt=attempt + 1`, otherwise this will never stop retrying as long as `retry_limit>1`.

mikealfare (dismissed):

mikealfare (dismissed):

mikealfare (approved):

PR #374, opened 2024-12-06
ADAP-1124: Establish shared workflows for the monorepo migration

mikealfare:
### Problem

We're moving to a monorepo for adapters. This will require workflows that can be reused across adapters.

### Solution

Migrate workflows from `dbt-athena`, which is already technically a monorepo (two packages) and make updates as needed to generalize. Update the workflows to support `dbt-adapters` and `dbt-tests-adapter` in the short term until their folder structures are updated, and then remove any existing workflows which are no longer necessary. Use this refactor as an opportunity to also clean up existing workflows which will live on in the monorepo.

### Checklist


mikealfare:
This was intended to show renamed files next to each other, but GH doesn't appear to be smart enough to interpret both a `git mv` and file updates and instead shows a deleted file and a new file instead.

PR #376, opened 2024-12-06
Respect the `limit` config when storing test failures

### LINKED ISSUE #373: [Bug] All test failures are stored even when a limit config is set

### Current Behavior

_Hi there, firstly thanks for the amazing job you're doing, I am huge fan of dbt !_

I use dbt core with dbt-redshift.

I wrote a test on a simple seed to check the behavior of the _store_failures_ test config when used in addition to the _limit_ config.

The seed I test (`seeds/customers.csv`) looks like this and has 100 rows (from ID 1 to 100):
```csv
ID,FIRST_NAME,LAST_NAME
1,Michael,P.
2,Shawn,M.
3,Kathleen,P.
4,Jimmy,C.
5,Katherine,R.
6,Sarah,R.
7,Martin,M.
8,Frank,R.
9,Jennifer,F.
10,Henry,W.
```
Download file here : [customers.csv](https://github.com/dbt-labs/dbt-core/files/14063939/customers.csv)

The test definition (`seeds/_seeds.yml`) looks like this:
_TLDR : The "ID" column must be strictly lower than 50. Store the failures. But stop if more than 20 failures._

```yaml
version: 2

seeds:
  - name: customers
    +quote_columns: false
    columns:
      - name: ID
        tests:
          - dbt_utils.accepted_range:
              name: my_test_with_limit_and_store_failures
              max_value: 50
              inclusive: true
              config:
                store_failures: true
                warn_if: "<5"
                error_if: ">=5"
                limit: 20
```

What I experience when running `dbt test -s customers` :
- The test log seems OK
image

- But my test audit table generated in my warehouse contains all the failed row (50) and not 20 (the limit specified).
image

### Expected Behavior

I guess only the 20 first row must be stored in the audit table (based on the [store_failure config documentation](https://docs.getdbt.com/reference/resource-configs/store_failures) pictured here)

image

### Steps To Reproduce

```bash 
python3 -m venv .venv. 
.venv/bin/activate; pip install -r requirements.txt
dbt deps
dbt seed
dbt test -s customers
```

Where requirements.txt is 

```txt
dbt-core==1.7.6
dbt-redshift==1.7.2
```

### Relevant log output

_No response_

### Environment

```markdown
- OS:Ubuntu 22.04.3 LTS
- Python: 3.9.7
- dbt: 1.7.6
```

redshift

### Additional Context

I noticed this SQL file generated in the target folder if it can help 

image

dbeatty10:
resolves #373

### Problem

Currently, the `limit` config is applying when _counting_ the number of failures, but it is not applying the limit when _storing_ the errors.

So behavior in our docs for the `limit` config for data tests doesn't match the current behavior.

As explained in https://github.com/dbt-labs/dbt-core/issues/11543:

> ... the limit is being added when running the test against the table that stores the failures _not_ when creating the table that stores the failures.
> 
> This seems to defeat the point of a test failures limit. 
> 
> According to the dbt docs:
> The [Limit](https://docs.getdbt.com/reference/resource-configs/limit) exists to...
> > Limit the number of failures that will be returned by a test query. We recommend using this config when working with large datasets and [storing failures in the database](https://docs.getdbt.com/reference/resource-configs/store_failures).
> 
> And [Store Failures](https://docs.getdbt.com/reference/resource-configs/store_failures) says...
> > When true, `store_failures` saves all records (up to [limit](https://docs.getdbt.com/reference/resource-configs/limit)) that failed the test. Failures are saved in a new table with the name of the test.

### Solution

When data test failures are stored, dbt executes two different queries:
1. Storing the failures
2. Retrieving the stored failures so they can be counted and compared against [severity](https://docs.getdbt.com/reference/resource-configs/severity) config

When test failures are **not** stored, dbt only executes the 2nd query.

So this PR opts to **move** the `limit` from being embedded within the 2nd query into a common source for **both** queries.

To preserve backwards compatibility for [`get_test_sql`](https://github.com/dbt-labs/dbt-adapters/blob/6255adf01aa4e3004d0267047c02813ee10d6b5e/dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/helpers.sql#L5), `limit` is converted to being an optional argument with a default of `none`.

### Future refactor

https://github.com/dbt-labs/dbt-adapters/pull/376 contains a minimal fix (this PR!) to [make it work](https://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast) (by fixing the bug).
https://github.com/dbt-labs/dbt-adapters/pull/375 contains a full suggested refactor to [make it "right"](https://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast) (by splitting the code into two composable pieces + making it more readable).

<details>
<summary>

### Testing
</summary>

#### Functional testing

✅ Succeeded [here](https://github.com/dbt-labs/dbt-adapters/actions/runs/14696289513/job/41238355194?pr=376#step:7:713) (as hoped).

✅ Failed [here](https://github.com/dbt-labs/dbt-adapters/actions/runs/14696512068/job/41238868223?pr=376#step:7:2355) (as anticipated) when the `limit` was was changed without changing the expected value to match.

#### Failing test in CI

This test is failing in CI, but appears completely unrelated to the changes in this PR, so assuming it flaky or something:
```
pytest tests/functional/adapter/materialized_view_tests/test_materialized_view.py::TestBigqueryMaterializedViewsBasic::test_materialized_view_create_idempotent
```

</details>

### Checklist


dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql
```diff
@@ -1,6 +1,12 @@
...
 
   {% set relations = [] %}
+  {% set limit = config.get('limit') %}
+
+  {% set main_sql %}
+    {{ sql }}
+    {{ "limit " ~ limit if limit != none }}
+  {% endset %}
 
   {% if should_store_failures() %}
```
dbernett-spring, line 11:
```suggestion
    {% if should_store_failures() %}

        {% set limit = config.get('limit') %}

        {% set test_sql %}
            {{ sql }}
            {{ "limit " ~ limit if limit != none }}
        {% endset %}

```

dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql
```diff
@@ -26,32 +32,29 @@
...
 
     {% do relations.append(target_relation) %}
 
+    {# Since the test failures have already been saved to the database, reuse that result rather than querying again #}
     {% set main_sql %}
         select *
         from {{ target_relation }}
     {% endset %}
 
     {{ adapter.commit() }}
```
dbernett-spring, line 47: {% else %}

      {% set main_sql = sql %}

dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql
```diff
@@ -26,32 +32,29 @@
...
```
dbernett-spring, line 46:
```suggestion
    {{ adapter.commit() }}
    
      {% else %}

          {% set main_sql = sql %}
```

dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql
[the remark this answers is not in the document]
dbeatty10: Good call-out @dbernett-spring !

### Key insight

💡 Key insight: the `limit` is supposed to apply regardless the if `store_failures` is true or not. And there are surely people relying on the latter.

### Aims of this fix

The main aims of the original implementation in this PR were two-fold:
1. Apply the limit in _one_ place and apply it _one_ time during the processing (versus applying it in multiple places or multiple times)
2. Make the minimal changes possible.

And the original implementation in this PR looks correct and robust to me and accomplishes both of those aims. But maybe there are aims that are more desirable!

Agreed that it could be confusing to follow the logic due to the reassignment. So it's just a matter of taste from here (and how deeply to refactor for readability).

### Larger proposed refactor in separate PR 

(https://github.com/dbt-labs/dbt-adapters/pull/375 is my proposed deeper refactor for readability. The [test materialization logic](https://github.com/dbt-labs/dbt-adapters/blob/aa6c7933d370b0b80ee9e4d189ab34d9e63bc7e7/dbt/include/global_project/macros/materializations/tests/test.sql) is slimmed down drastically, for example.)

I'd like to see this refactor happen. I'd also like to be a slim as possible in this fix and not combine fix + refactor within the same PR -- would prefer to do them separately.

### Your suggestions

The angle you proposed is another option (depending on taste), but it's not necessarily more robust / less brittle IMO.

For example, maybe it's a legitimate option to apply the limit in _two_ places: once when (optionally) materializing the failures as a table (or a view!), and again when fetching the results. Trade-offs: while it might be "okay" to apply the limit twice, it seems better to me to not, if possible.

If we adopted your suggestions, we'd also need to make sure `{% set limit = config.get('limit') %}` was executed in both scenarios.

### Slightly more verbose option

Here's a slight variant on your suggestion that I think could work -- what do you think?

```sql
  {% set limit = config.get('limit') %}

  {% set sql_with_limit %}
    {{ sql }}
    {{ "limit " ~ limit if limit != none }}
  {% endset %}

  {% if should_store_failures() %}

    {% set target_relation = store_failures(sql_with_limit) %}

    {# Since the test failures have already been saved to the database, reuse that result rather than querying again #}
    {% set main_sql %}
        select *
        from {{ target_relation }}
    {% endset %}

  {% else %}

     {% set main_sql = sql_with_limit %}

  {% endif %}
  ```

File: dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql (hunk 1 above)
Comment (dbeatty10) [reply dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql]: Partially integrated this suggestion into 0296d2debe3fc1432fcdaf17374536d64b6a2f50 and 8ca809a644ff5e3a866e107bd929764eb7585753

I prefer to include any deeper refactors into https://github.com/dbt-labs/dbt-adapters/pull/375 instead.

File: dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql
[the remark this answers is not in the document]
Comment (dbeatty10) [reply dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql:1]: @dbernett-spring Thanks again for your suggestions and feedback! 🧠 Partially integrated them into 0296d2debe3fc1432fcdaf17374536d64b6a2f50 and 8ca809a644ff5e3a866e107bd929764eb7585753

Keeping this to the essential changes to fix the bug though as I prefer to include any deeper refactors into https://github.com/dbt-labs/dbt-adapters/pull/375 instead. Further suggestions / feedback welcome there! Will update that PR after this PR is approved and merged.

**Comment by dbernett-spring** [comment]
@dbeatty10 thank you for taking this on! Much appreciated!

**Comment by dbeatty10** [comment]
> @dbeatty10 thank you for taking this on! Much appreciated!

@dbernett-spring Thank you for helping elevate the profile of the problem and analyzing the relative merits of proposed solutions! 🫶

File: dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql [hunk 4]
```diff
@@ -1,6 +1,12 @@
 {%- materialization test, default -%}
 
   {% set relations = [] %}
+  {% set limit = config.get('limit') %}
+
+  {% set sql_with_limit %}
+    {{ sql }}
+    {{ "limit " ~ limit if limit != none }}
```
colin-rogers-dbt, line 8: won't this break on platforms that don't support limit?

colin-rogers-dbt: Can you use the `get_limit_sql` macro here?

colin-rogers-dbt: example where fabric overrides to achieve correct limit https://github.com/microsoft/dbt-fabric/blob/772d896d366137ff1ed1a364110d9eee67a727e6/dbt/include/fabric/macros/adapters/show.sql#L4

dbeatty10: Good call! 🧠 I'll make that change in this PR.

I'll save it for a separate PR to update `default__get_test_sql` to use [`get_limit_sql`](https://github.com/dbt-labs/dbt-adapters/blob/6255adf01aa4e3004d0267047c02813ee10d6b5e/dbt-adapters/src/dbt/include/global_project/macros/adapters/show.sql#L21):

https://github.com/dbt-labs/dbt-adapters/blob/6255adf01aa4e3004d0267047c02813ee10d6b5e/dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/helpers.sql#L12

dbeatty10: Addressed in 421970712483177c49442bc9dba6f1c53dac4c51.

Felt awkward to use a macro named `get_limit_subquery_sql`, so opened https://github.com/dbt-labs/dbt-adapters/issues/1038 as a result.

colin-rogers-dbt (approved):

PR #382, opened 2024-12-13
Fix concurrency test

mikealfare:
### Problem

We're verifying the expected run results for concurrency tests by parsing the logs. This creates a brittle test as the log structure could change over time. This happened recently, breaking this test and causing a false positive.

### Solution

Replace the verification via logs with a verification via run results.

### Checklist

### DIFF
--- a/dbt-tests-adapter/dbt/tests/adapter/concurrency/test_concurrency.py
+++ b/dbt-tests-adapter/dbt/tests/adapter/concurrency/test_concurrency.py
@@ -1,11 +1,13 @@
+from collections import Counter
+
 import pytest
 
+from dbt.artifacts.schemas.results import RunStatus
 from dbt.tests.util import (
     check_relations_equal,
     check_table_does_not_exist,
     rm_file,
     run_dbt,
-    run_dbt_and_capture,
     write_file,
 )
 
@@ -317,8 +319,8 @@ class BaseConcurrency:
         rm_file(project.project_root, "seeds", "seed.csv")
         write_file(seeds__update_csv, project.project_root, "seeds", "seed.csv")
 
-        results, output = run_dbt_and_capture(["run"], expect_pass=False)
-        assert len(results) == 7
+        results = run_dbt(["run"], expect_pass=False)
+
         check_relations_equal(project.adapter, ["seed", "view_model"])
         check_relations_equal(project.adapter, ["seed", "dep"])
         check_relations_equal(project.adapter, ["seed", "table_a"])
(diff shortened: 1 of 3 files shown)


mikealfare:
I tested this against `dbt-postgres` and it resolves the issue.

dbt-tests-adapter/pyproject.toml
```diff
@@ -52,6 +52,16 @@ include = ["dbt/tests", "dbt/__init__.py"]
 [tool.hatch.build.targets.wheel]
 include = ["dbt/tests", "dbt/__init__.py"]
 
+[tool.hatch.envs.default]
```
mikealfare, line 55: I'm surprised we didn't have this before. This allows us to develop within `dbt-tests-adapter` using a hatch environment (to get things like autocomplete in the IDE and whatnot).

As an aside, the `@ {root:uri}/..` bit is what local dev will look like in the monorepo.  Branches will no longer need to be set for packages in the `dbt-adapters` repo since `git` will pull all of them down at the same time. We'll still need to specify branches for `dbt-common` and `dbt-core` since they'll be outside of the repo.

dbt-tests-adapter/dbt/tests/adapter/concurrency/test_concurrency.py
```diff
@@ -317,16 +319,22 @@ def test_concurrency(self, project):
         rm_file(project.project_root, "seeds", "seed.csv")
         write_file(seeds__update_csv, project.project_root, "seeds", "seed.csv")
 
-        results, output = run_dbt_and_capture(["run"], expect_pass=False)
-        assert len(results) == 7
+        results = run_dbt(["run"], expect_pass=False)
+
```
mikealfare, line 323: We can remove `len(results) == 7` here since that's implicitly checked via the `dict` comparison below. For example, if there is another status that's not included in `expected_statuses` that has non-zero occurrence, then `assert result_statuses == expected_statuses` would fail.

dbt-tests-adapter/pyproject.toml
colin-rogers-dbt: Would this impact individual adapters installing a particular version of core/other deps?

mikealfare: This affects the development environment of `dbt-tests-adapter` itself. `hatch` environments are akin to `dev-requirements.txt`; they do not affect what gets packaged and shipped.

colin-rogers-dbt (approved):

PR #385, opened 2024-12-17
Fix Snapshot Merge Query

peterallenwebb:
Snowflake Bug Report: https://github.com/dbt-labs/dbt-snowflake/issues/1275
BigQuery But Report: https://github.com/dbt-labs/dbt-snowflake/issues/1275

### Problem

The default merge query which implements snapshot logic was missing a filter for inserts, which caused it to miss inserting new records on delete.

### Solution

Add a new filter for 'insert' records.

### Checklist


dbt/include/global_project/macros/materializations/snapshots/snapshot_merge.sql
```diff
@@ -11,7 +11,7 @@
 
     merge into {{ target.render() }} as DBT_INTERNAL_DEST
     using {{ source }} as DBT_INTERNAL_SOURCE
-    on DBT_INTERNAL_SOURCE.{{ columns.dbt_scd_id }} = DBT_INTERNAL_DEST.{{ columns.dbt_scd_id }}
+    on DBT_INTERNAL_SOURCE.{{ columns.dbt_scd_id }} = DBT_INTERNAL_DEST.{{ columns.dbt_scd_id }} and DBT_INTERNAL_SOURCE.dbt_change_type != 'insert'
```
adrianburusdbt, line 14: Changing a condition like this _should_ fail some tests, but as far as I can tell there are none for snapshot_merge. Can you please add them?

peterallenwebb: Hi @adrianburusdbt,

This query is used whenever any snapshot is executed, so there are a variety of tests which cover it, including [this new one](https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-tests-adapter/dbt/tests/adapter/simple_snapshot/new_record_mode.py) for the new_record mode. Many of the other tests are currently in the dbt-core test suite and therefore only execute against dbt-postgres, but I believe @gshank is working on bringing them into dbt-adapters, where they belong.

The logic used for updating snapshots is fairly complex, and I could certainly be overlooking something, but I believe this change is correct, or at least brings the query into alignment with the behavior of the [dbt-postgres implementation](https://github.com/dbt-labs/dbt-postgres/blob/main/dbt/include/postgres/macros/materializations/snapshot_merge.sql), which is executed as two separate queries, since PostgreSQL historically had no merge statement.

peterallenwebb: Having thought about it further, I'm going to take steps to insure the 'insert' records created by the new_record mode regenerate the dbt_scd_id column. After that, there could still be some unusual corner cases where this behavior might be preferable (i.e. hash collisions) but that may not be a good enough reason to retain it.

adrianburusdbt (approved):

colin-rogers-dbt (approved):

PR #388, opened 2024-12-18
Fix release branch conditional

mikealfare:


.github/workflows/_generate-changelog.yml
```diff
@@ -222,8 +222,7 @@ jobs:
         steps:
         -   id: branch
             run: |
-                branch=""
-                if [ ${{ needs.temp-branch.outputs.name == '' || inputs.merge }}]
```
mikealfare, line 226: This was evaluating to `False` even when `inputs.merge==True`. I verified that `inputs.merge==True` based on paths this workflow took in previous steps. But when it got here, it always took the `else` path. I think it's just because of the single brackets instead of double brackets.

colin-rogers-dbt (approved):

PR #383, opened 2024-12-17
Added equals macro that handles null value comparison

adrianburusdbt:
resolves #
- https://github.com/dbt-labs/dbt-core/issues/7873
- https://github.com/dbt-labs/dbt-adapters/issues/159

### Problem
Null comparison was not handled properly in the base adapter. While some value comparisons were missing this case altogether, others were fixed but not in a repeatable, organized way.

### Solution

Based loosely on https://github.com/dbt-labs/dbt-adapters/pull/110, but ultimately it's following the suggestions made here: https://github.com/dbt-labs/dbt-core/issues/6997#issuecomment-1832992367
Extracted the logic that was in the test fixture into an 'equals.sql' macro that can be reused and replaced all other usages with this utility macro.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-adapters/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development, and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX

PR #390, opened 2024-12-19
Revert "Added equals macro that handles null value comparison"

VersusFacit:
Reverts dbt-labs/dbt-adapters#383

Merging this broke Postgres, which we may not have tested against

https://github.com/dbt-labs/dbt-postgres/actions/runs/12403895291/job/34628184143


VersusFacit:
reverting this fixes PG: https://github.com/dbt-labs/dbt-postgres/actions/runs/12404097734/job/34628729119

colin-rogers-dbt (approved):

PR #399, opened 2025-01-08
ADAP-1125: Migrate dbt-athena into dbt-adapters

mikealfare:
Migrate `dbt-athena` into this repo using the following shell script and call:

```shell
repo=$1
source_branch=$2
target_branch=$3

git remote remove old || true
git remote add old https://github.com/dbt-labs/$repo.git
git fetch old

# merge the updated branch from the legacy repo into the dbt-adapters repo
git checkout $target_branch
git merge old/$source_branch --allow-unrelated-histories

# remove the remote that was created by this process
git remote remove old || true
```

```shell
source ./scripts/migrate-adapter.sh dbt-athena main monorepo/migrate-dbt-athena
```

Update workflows to include `dbt-athena` and `dbt-athena-community`.

Make updates related with code quality checks due to slightly different config.

Merge this PR as a regular merge instead of a squash merge in order to retain git blame history.


mikealfare:
The CLA is failing because this PR contains all commits ever made to `dbt-athena`, which includes those prior to dbt Labs assuming ownership.

I ran all of the PR checks by adding the pull_request trigger to `pull_request_checks.yml`. The [results](https://github.com/dbt-labs/dbt-adapters/actions/runs/12700698374?pr=399) passed. I then removed the trigger since we only want pull_request_target as a trigger.

.github/workflows/publish.yml
```diff
@@ -39,17 +41,27 @@ jobs:
...
+        uses: ./.github/workflows/_integration-tests.yml
+        with:
+            package: ${{ inputs.package }}
+            branch: ${{ inputs.branch }}
+            repository: ${{ github.repository }}
+        secrets: inherit
+
     generate-changelog:
-        needs: unit-tests
+        needs: [unit-tests, integration-tests]
```
VersusFacit, line 54: will this pass on skip by if or fail. I can't remember the boolean logic from me doing similar a year ago. You may need an "Always" clause or something like that.

dbt-athena/hatch.toml
```diff
@@ -0,0 +1,57 @@
...
+[build.targets.wheel]
+packages = ["src/dbt/adapters", "src/dbt/include"]
+sources = ["src"]
+
+[envs.default]
+dependencies = [
+    "dbt-adapters @ git+https://github.com/dbt-labs/dbt-adapters.git",
+    "dbt-common @ git+https://github.com/dbt-labs/dbt-common.git",
+    "dbt-tests-adapter @ git+https://github.com/dbt-labs/dbt-adapters.git#subdirectory=dbt-tests-adapter",
+    "dbt-core @ git+https://github.com/dbt-labs/dbt-core.git#subdirectory=core",
```
VersusFacit, line 17: These are always pulling from main. Is that what we actually want?

VersusFacit (commented):
Two comments, probably minor. Nice to see unit tests are passing. I'd love to see a functional test run pass too. Not sure if that's possible at this stage. It would make sense to also have that follow on after merging and hotfixing as needed I suppose...

dbt-athena/hatch.toml
mikealfare: Yes, that's how we currently do it. Once all adapters are in this repo, I'll update to install locally so that it's always the current branch. I want to do that in a separate PR though.

.github/workflows/publish.yml
mikealfare: Oh you might be right. I'm hesitant to put that in this PR though. For some reason I feel like it should be a separate PR since it's not directly related to migrating dbt-athena into this repo. Though I suppose it's also related to adding integration tests here. I'm also trying to minimize how many commits are in this PR since it won't be a squash merge. Are you comfortable with waiting?

As a side note, I believe what we want is `if: !failure()`.

mikealfare:
> Two comments, probably minor. Nice to see unit tests are passing. I'd love to see a functional test run pass too. Not sure if that's possible at this stage. It would make sense to also have that follow on after merging and hotfixing as needed I suppose...

It got pushed up off the screen. See my [post](https://github.com/dbt-labs/dbt-adapters/pull/399#issuecomment-2581505287) above.

.github/workflows/publish.yml
VersusFacit: Yeah, I can absolutely buy that. I just wanted to make sure we at least flag it and know that it's something we need to do.

VersusFacit (approved):
lets go!

PR #403, opened 2025-01-13
ADAP-1167: Add table format telemetry reporting to Athena adapter

VanTudor:
# Description

Send out table format telemetry data in the Athena Adapter

## Checklist

- [x] This PR considers the [contributing guide](https://github.com/dbt-labs/dbt-adapters#contributing)
- [x] This PR is small and focused on a single feature or bug fix
- [x] This PR includes unit testing, or unit testing is not necessary
- [x] This PR includes functional testing, or functional testing is not necessary


dbt-athena/tests/functional/adapter/test_retries_iceberg.py
```diff
@@ -60,6 +60,10 @@
 seeds__expected_target_post = "id,status\n" + "\n".join([f"{i},{i}" for i in range(PARALLELISM)])
 
 
+@pytest.mark.flaky
+@pytest.mark.skip(
+    reason="Unreliable test in practice. No apparent reliable way to trigger a retry, which causes the test to fail. Disabled until we get a resolution here https://github.com/dbt-labs/dbt-athena/pull/657/files#r1922043351"
```
VanTudor, line 65: <Drawing attention to this>

.github/workflows/pull-request-checks.yml
```diff
@@ -83,8 +83,29 @@ jobs:
...
             python-version: ${{ matrix.python-version }}
+            is-flaky-test-run: false
+        secrets: inherit
+    # Special case here for Athena, since we have two identical packages -- dbt-athena and dbt-athena-community
+    # We don't want certain tests within this adapter to run in parallel
+    integration-tests-flaky-athena:
+        uses: ./.github/workflows/_integration-tests.yml
+        strategy:
+            fail-fast: false
+            max-parallel: 1
```
VanTudor, line 94: I did these workflow changes so that `TestIcebergRetriesDisabled` is isolated in a separate job and so that it's forced to run in series. This is to rule out that parallel test runs write to the same place, causing the test to fail.

However, I ended up disabling that test (see comment below). I think it's debatable whether to still keep the workflow changes here (as scaffolding). I'm leaning towards yes, but happy to remove this stuff, in case you disagree.

dbt-athena/tests/functional/adapter/test_retries_iceberg.py
VanTudor: This test seems to fail randomly, irrespective of my changes. E.g. https://github.com/dbt-labs/dbt-athena/actions/runs/12756721311/job/35555467442#step:6:595

.github/workflows/_integration-tests.yml
```diff
@@ -94,7 +103,8 @@ jobs:
             with:
                 role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/${{ secrets.ASSUMABLE_ROLE_NAME }}
                 aws-region: ${{ vars.DBT_TEST_ATHENA_REGION_NAME }}
-        -   run: hatch run integration-tests
+        # run flaky tests command if is-flaky-test-run flag is true. Otherwise, run the other tests
+        -   run: ${{ inputs.is-flaky-test-run && 'hatch run integration-tests -m flaky -n1' ||  'hatch run integration-tests -m "not flaky"' }}
```
mikealfare, line 107: I'd rather break this out into two steps that condition on `inputs.is-flaky-test-run` than put the conditional in the command. That's how we had set it up for `dbt-redshift` and I think it's a bit cleaner to read. [Here's](https://github.com/dbt-labs/dbt-adapters/blob/0c40d81d914feabf7f04c7dd3f406db22a862120/.github/workflows/_integration-tests.yml#L185-L188) what it looks like in the migration feature branch.

.github/workflows/pull-request-checks.yml
mikealfare: I think we'll ultimately need to put this here because it needs to be considered when executing the strategy matrix, not within the call to the integration tests workflow. However, we might be able to tuck some of the pieces into pytest markers, hatch commands, etc. I'll write more in the overall comment.

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -1452,3 +1452,17 @@ def _run_query(self, sql: str, catch_partitions_limit: bool) -> AthenaCursor:
             LOGGER.debug(f"CAUGHT EXCEPTION: {e}")
             raise e
         return cursor
+
+    @classmethod
+    def _get_adapter_specific_run_info(cls, config: RelationConfig) -> Dict[str, Any]:
+        table_format: Optional[str] = None
+        if (
```
mikealfare, line 1459: This feels more like "look before you leap". Would you be against something like:

```python
def _get_adapter_specific_run_info(cls, config: RelationConfig) -> Dict[str, Any]:
    try:
        table_format = config._extra.get("table_type")
    except AttributeError:
        table_format = None
    return {
        "adapter_type": "athena",
        "table_format": table_format,
    }
```

dbt-athena/tests/unit/test_adapter_telemetry.py
```diff
@@ -0,0 +1,28 @@
...
+    assert res.adapter_name == "athena"
+    assert res.base_adapter_version == dbt.adapters.__about__.version
+    assert res.adapter_version == dbt.adapters.athena.__version__.version
+
+    assert res.model_adapter_details == {
+        "adapter_type": "athena",
+        "table_format": "iceberg",
+    }
+
+    assert type(res) is AdapterTrackingRelationInfo
```
mikealfare, line 28: I think `isinstance` checks are preferred over `type() is` checks, but I can't recall why. Perhaps due to subclassing.

mikealfare (commented):
Reading through this reminded me of some conversations we had months about flaky tests. The short version is that we might want to refer to them as `serial` tests and some tests may always be serial for one reason or another. Also, there's a `pytest` plugin for flaky tests which I believe works off of a marker named `flaky`, which would collide here. I'll go into a bit of detail here since we discussed this also be a framework to build on, and not just specific to `dbt-athena`.

We have "flaky" tests in several adapters now, including `dbt-athena`, `dbt-redshift`, and `dbt-bigquery`. These are all for different reasons. We're not clear on `dbt-athena`, but it seems like it could be resource collisions in S3. For `dbt-bigquery`, this was happening due to resource availability for compute on python models. On `dbt-redshift`, it seems like it could be a race condition between the head node and other nodes when collecting metadata on several small models in parallel. In any case, some of these will run in series for the foreseeable future. And being able to mark a test as serial in the code and having the CI checks validate that would be great; hence why I think your framework approach is the right idea instead of just solving it for `dbt-athena`.

Given the context, I think we should consider a marker named something along the lines of `serial`, `series`, etc. We should then add this marker to all adapters to make it available and to add it to the workflows regardless if whether there are tests that fit. This will make it easier to add tests in the future if needed. Finally, we need to account for the scenario where the new marker catches no tests. `pytest` returns a code of 5 instead of 0 in this scenario, so that successful tests can be distinguished from something like running against the wrong directory with no tests. We solved for this on `dbt-redshift` with this [hack](https://github.com/dbt-labs/dbt-redshift/blob/9e6516ac3715b93486cbab70669404b827f0a3a6/tests/conftest.py#L4-L12).

tl;dr; If we want to treat this as a framework (I agree), then we should agree on the appropriate name for the marker and then add the other pieces to implement it across adapters. And if we do that, I would argue that the framework should be its own PR, which then gets used here. That would allow us to easily revert either piece if needed and would make our intentions clearer when looking back in the future.

VanTudor:
> Reading through this reminded me of some conversations we had months about flaky tests. The short version is that we might want to refer to them as `serial` tests and some tests may always be serial for one reason or another. Also, there's a `pytest` plugin for flaky tests which I believe works off of a marker named `flaky`, which would collide here. I'll go into a bit of detail here since we discussed this also be a framework to build on, and not just specific to `dbt-athena`.
> 
> We have "flaky" tests in several adapters now, including `dbt-athena`, `dbt-redshift`, and `dbt-bigquery`. These are all for different reasons. We're not clear on `dbt-athena`, but it seems like it could be resource collisions in S3. For `dbt-bigquery`, this was happening due to resource availability for compute on python models. On `dbt-redshift`, it seems like it could be a race condition between the head node and other nodes when collecting metadata on several small models in parallel. In any case, some of these will run in series for the foreseeable future. And being able to mark a test as serial in the code and having the CI checks validate that would be great; hence why I think your framework approach is the right idea instead of just solving it for `dbt-athena`.
> 
> Given the context, I think we should consider a marker named something along the lines of `serial`, `series`, etc. We should then add this marker to all adapters to make it available and to add it to the workflows regardless if whether there are tests that fit. This will make it easier to add tests in the future if needed. Finally, we need to account for the scenario where the new marker catches no tests. `pytest` returns a code of 5 instead of 0 in this scenario, so that successful tests can be distinguished from something like running against the wrong directory with no tests. We solved for this on `dbt-redshift` with this [hack](https://github.com/dbt-labs/dbt-redshift/blob/9e6516ac3715b93486cbab70669404b827f0a3a6/tests/conftest.py#L4-L12).
> 
> tl;dr; If we want to treat this as a framework (I agree), then we should agree on the appropriate name for the marker and then add the other pieces to implement it across adapters. And if we do that, I would argue that the framework should be its own PR, which then gets used here. That would allow us to easily revert either piece if needed and would make our intentions clearer when looking back in the future.

Sounds good. `serial` seems more appropriate as a marker to me 👍. Thanks for pointing me to that part of the Redshift adapter. I'll open a different PR with the serial tests "framework".

I've rolled back the workflow changes here and just kept that problematic test disabled.

dbt-athena/tests/unit/test_adapter_telemetry.py
VanTudor: Yeah isinstance checks inheritance too, good point. Changed it.

dbt-athena/src/dbt/adapters/athena/impl.py
VanTudor: Done

.github/workflows/_integration-tests.yml
VanTudor: Makes sense. Will address in separate PR.

mikealfare (approved):

PR #462, opened 2025-01-13
ADAP-1184: Create scripts to streamline moving branches into the monorepo and backporting bug fixes to legacy repos

mikealfare:
We migrated adapters from their individual repos into the `dbt-adapters` monorepo. We were also able to transfer open issues to the monorepo. However, pull requests cannot be transferred; only their related feature branches can be migrated.

This script helps users migrate their feature branch into the `dbt-adapters` monorepo by rebasing their feature branch on `dbt-adapters@main`, and then creating a new branch on `dbt-adapters` that contains the commits of that rebase. This script should be run from `dbt-adapters`, or a related fork.


mikealfare:
@cla-bot check

scripts/migrate-adapter.sh
```diff
@@ -1,17 +1,17 @@
...
-target_branch=$3
+ADAPTER=$1
+SOURCE_BRANCH=$2
+TARGET_BRANCH=$3
 
 # create a remote for the legacy adapter repo and fetch the latest commits
-git remote remove old || true
-git remote add old https://github.com/dbt-labs/$repo.git
-git fetch old
+git remote remove adapter || true
```
colin-rogers-dbt, line 6: Should the remote name be `$ADAPTER`? What if I want to port PRs from multiple adapter repos?

mikealfare: I figured we would be doing this one at a time (especially if we're migrating the entire adapter). So we'd wind up adding/removing each time. I could update that though if you feel strongly about it.

colin-rogers-dbt (approved):

PR #606, opened 2025-01-14
ADAP-1126: Filter pull request checks to only impacted packages

mikealfare:
### Problem

Now that we have all of the adapters in the same repo, CI checks have grown in number. We should limit CI checks to only those that apply to affected packages. We can also use this as an opportunity to test changes to dependent packages, e.g. a change to `dbt-adapters` should trigger tests across all adapters.

### Solution

- Use the `dorny/path-filters` action that has been previously used to filter down to affected packages.
- Apply different filters based on the check being done, e.g. `dbt-tests-adapter` has no unit tests nor integration tests, but does need a changelog


.github/workflows/pull-request-checks.yml
```diff
@@ -98,10 +117,10 @@ jobs:
...
         if: always()
-        needs: [changelog-entry, code-quality, verify-builds, unit-tests, integration-tests]
+        needs: [changelog-entry-check, code-quality, verify-build, unit-tests, integration-tests]
         runs-on: ${{ vars.DEFAULT_RUNNER }}
         steps:
         -   uses: re-actors/alls-green@release/v1
             with:
                 jobs: ${{ toJSON(needs) }}
-                allowed-skips: 'changelog-entry'
+                allowed-skips: [unit-tests, integration-tests]
```
colin-rogers-dbt, line 126:
```suggestion
                allowed-skips: [unit-tests, integration-tests, changelog-entry-check]
```

PR #610, opened 2025-01-20
ADAP-1131: Migrate dbt-redshift into dbt-adapters

mikealfare:
Migrate `dbt-redshift` into this repo using the following shell script and call:

```shell
repo=$1
source_branch=$2
target_branch=$3

git remote remove old || true
git remote add old https://github.com/dbt-labs/$repo.git
git fetch old

# merge the updated branch from the legacy repo into the dbt-adapters repo
git checkout $target_branch
git merge old/$source_branch --allow-unrelated-histories

# remove the remote that was created by this process
git remote remove old || true
```

```shell
source ./scripts/migrate-adapter.sh dbt-redshift monorepo/stage monorepo/migrate-dbt-redshift
```

Update workflows to include `dbt-redshift`.

Make updates related with code quality checks due to slightly different config.

Merge this PR as a regular merge instead of a squash merge in order to retain git blame history.


github-christophe-oudar:
Is there a typo in the name of this PR? I feel like you meant
`ADAP-1131: Migrate dbt-redshift into dbt-adapters` instead of `ADAP-1131: Migrate dbt-bigquery into dbt-redshift` no?

VersusFacit (approved):
LGTM

mikealfare:
Yeah, that's a regression. That pin got bumped when doing the auth work, but in parallel with the monorepo migration. So the migration reverted it. Thanks for the catch. Please open a bug in the dbt-labs/dbt-adapters repo.

mikealfare:
> Yeah, that's a regression. That pin got bumped when doing the auth work, but in parallel with the monorepo migration. So the migration reverted it. Thanks for the catch. Please open a bug in the dbt-labs/dbt-adapters repo.

@padbk No need for the bug report, we already resolved this. Please install `1.9.2` from PyPI.

PR #755, opened 2025-01-29
Run flaky tests in their own job so that they can be rerun independently

mikealfare:
### Problem

The flaky tests for `dbt-redshift` and `dbt-bigquery` got lumped in with the rest of the tests in the same job during the monorepo migration. That means that all tests need to be rerun if the flaky tests fail. This can take a lot of time.

### Solution

Break out the flaky tests as separate jobs so that they can be rerun independently.

### DIFF
--- a/.github/workflows/_integration-tests.yml
+++ b/.github/workflows/_integration-tests.yml
@@ -126,8 +126,33 @@ jobs:
         -   uses: pypa/hatch@install
         -   run: hatch run integration-tests tests/functional -k "not TestPython"
             working-directory: ./${{ inputs.package }}
+
+    integration-tests-bigquery-flaky:
+        # we only run this for one python version to avoid running in parallel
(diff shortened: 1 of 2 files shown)


.github/workflows/_integration-tests.yml
```diff
@@ -235,9 +260,6 @@ jobs:
             REDSHIFT_TEST_IAM_USER_ACCESS_KEY_ID: ${{ vars.AWS_USER_ACCESS_KEY_ID }}
             REDSHIFT_TEST_IAM_USER_SECRET_ACCESS_KEY: ${{ secrets.AWS_USER_SECRET_ACCESS_KEY }}
             REDSHIFT_TEST_IAM_ROLE_PROFILE: ${{ vars.AWS_ROLE_PROFILE }}
-            DBT_TEST_USER_1: ${{ vars.DBT_TEST_USER_1 }}
```
mikealfare, line 238: These are specified at the top of the file, this is redundant.

VersusFacit (approved):

PR #756, opened 2025-01-29
Install local packages in editable format

mikealfare:
The local packages are not installing correctly. Even if they did, they install in non-editable mode, making development across packages clunky. `hatch` doesn't have a built-in way to handle this. The best workaround proposed is to install via `hatch` pre-install commands.


VersusFacit (approved):
awesome!!

PR #757, opened 2025-01-29
Add support target_session_attrs

AlexanderSabirov:
### Problem

When specifying multiple hosts in a PostgreSQL cluster, it is possible to connect to the server in read-only mode.

### Solution

When specifying multiple hosts in a PostgreSQL cluster, it is necessary to pass the parameter target_session_attrs. This PR adds support for that parameter.

### Checklist

### DIFF
--- a/dbt-postgres/src/dbt/adapters/postgres/connections.py
+++ b/dbt-postgres/src/dbt/adapters/postgres/connections.py
@@ -35,6 +35,7 @@ class PostgresCredentials(Credentials):
     sslkey: Optional[str] = None
     sslrootcert: Optional[str] = None
     application_name: Optional[str] = "dbt"
+    target_session_attrs: Optional[str] = None
     retries: int = 1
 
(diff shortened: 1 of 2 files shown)


AlexanderSabirov:
Hello everyone!

I wanted to share a note about connecting to the PostgreSQL cluster. Because of our architecture, we can’t use a VIP or a load balancer for this.

Right now, if we list all the cluster hosts like host = server1,server2,server3, the connection will go to the first available server, even if it’s in read-only mode.

To fix this, I’ve added the target_session_attrs=read-write parameter in the Merge Request. This ensures that the connection only goes to a server that can handle read-write operations. You can find more details in the PostgreSQL docs: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS.

@mikealfare 
@colin-rogers-dbt 
Could you please review the Merge Request when you have a moment?

Thanks!

PR #761, opened 2025-01-30
Mark "available" functions for recording

peterallenwebb:
This changeset annotates (all except one of) the @available functions to support record/replay support within the new "Available" group. It also adds custom serialization support for the functions which need it.

### Checklist


colin-rogers-dbt (approved):

dbt-adapters/src/dbt/adapters/sql/impl.py
```diff
@@ -255,6 +257,10 @@ def validate_sql(self, sql: str) -> AdapterResponse:
         return adapter_response
 
     # This is for use in the test suite
+    @available
```
mikealfare, line 260: What's the rationale for making this available? From the comment, I struggle to see why this method even exists on this class, and isn't instead implemented on the test fixture.

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -890,6 +890,7 @@ def string_add_sql(
             )
 
     # This is used by the test suite
+    @available
```
mikealfare, line 893: See above

mikealfare (commented):
I assume you've already looked into this since we're using this elsewhere, but how does this impact run timing? Do the recordings happen after the method executes or before? Does the next action need to wait for the recording to happen, i.e. could this be blocking? Or is it more async, e.g. fire an event at a queue and move on?

dbt-adapters/src/dbt/adapters/sql/impl.py
peterallenwebb: @colin-rogers-dbt added this one, and I'm not sure it's strictly necessary?

peterallenwebb:
> I assume you've already looked into this since we're using this elsewhere, but how does this impact run timing? Do the recordings happen after the method executes or before? Does the next action need to wait for the recording to happen, i.e. could this be blocking? Or is it more async, e.g. fire an event at a queue and move on?

When the code is loaded without DBT_RECORD_MODE set, all of these decorators short-circuit. Since the resulting runtime code is unaffected, there is no performance implication.

Some performance impact when DBT_RECORD_MODE is turned on will be unavoidable. We'll have to evaluate and manage it  when we decide we want to record runs in production, but this work is intended to unlock recording these function calls in development scenarios.

PR #787, opened 2025-02-05
ADAP-1208: Fix publish workflow for `dbt-tests-adapter`

mikealfare:
We need to allow for unit tests to be skipped for `dbt-tests-adapter`. While changing this I noticed that we require changelog generation for internal publishing because we reference the branch in that workflow call. This is not correct since we don't publish a changelog internally; the reference should be to the input branch instead. Given multiple changes happening in quick succession, I made some of the exception configurable via GH environment variables instead of hard coded in the workflow.


.github/workflows/publish.yml
```diff
@@ -49,58 +53,69 @@ defaults:
...
         with:
             package: ${{ inputs.package }}
             branch: ${{ inputs.branch }}
-            repository: ${{ github.repository }}
         secrets: inherit
 
-    generate-changelog:
+    publish-prep-checks:
+        name: "Publish prep checks"
+        if: always()
```
colin-rogers-dbt, line 76: Shouldn't we skip this if integration/unit tests fail?

mikealfare: This will fail if integration/unit tests fail and will pass if they pass or are skipped. The failure of this job prevents downstream jobs from running because of `if: always() && needs.publish-prep-checks.result == 'success'`. I think this is the behavior we want.

colin-rogers-dbt (approved):

PR #790, opened 2025-02-05
ADAP-1170: Add base implementation of catalog support

colin-rogers-dbt:
resolves #

### Checklist

### DIFF
--- a/dbt-adapters/src/dbt/adapters/base/impl.py
+++ b/dbt-adapters/src/dbt/adapters/base/impl.py
@@ -65,6 +65,11 @@ from dbt.adapters.base.relation import (
 )
 from dbt.adapters.cache import RelationsCache, _make_ref_key_dict
 from dbt.adapters.capability import Capability, CapabilityDict
+from dbt.adapters.catalogs import (
+    CatalogIntegration,
+    CatalogIntegrationClient,
+    CatalogIntegrationConfig,
+)
 from dbt.adapters.contracts.connection import Credentials
 from dbt.adapters.contracts.macros import MacroResolverProtocol
 from dbt.adapters.contracts.relation import RelationConfig
@@ -269,6 +274,7 @@ class BaseAdapter(metaclass=AdapterMeta):
     Relation: Type[BaseRelation] = BaseRelation
     Column: Type[BaseColumn] = BaseColumn
     ConnectionManager: Type[BaseConnectionManager]
+    CATALOG_INTEGRATIONS: Dict[str, Type[CatalogIntegration]] = {}
 
     # A set of clobber config fields accepted by this adapter
     # for use in materializations
@@ -295,6 +301,14 @@ class BaseAdapter(metaclass=AdapterMeta):
         self._macro_resolver: Optional[MacroResolverProtocol] = None
         self._macro_context_generator: Optional[MacroContextGeneratorCallable] = None
         self.behavior = DEFAULT_BASE_BEHAVIOR_FLAGS  # type: ignore
+        self._catalog_client = CatalogIntegrationClient(self.CATALOG_INTEGRATIONS)
+
+    def add_catalog_integration(self, catalog: CatalogIntegrationConfig) -> CatalogIntegration:
+        return self._catalog_client.add(catalog)
+
+    @available
+    def get_catalog_integration(self, name: str) -> CatalogIntegration:
+        return self._catalog_client.get(name)
 
     ###
     # Methods to set / access a macro resolver
--- a/dbt-adapters/src/dbt/adapters/catalogs/_client.py
+++ b/dbt-adapters/src/dbt/adapters/catalogs/_client.py
@@ -1,0 +1,54 @@
+from typing import Dict, Type
+
+from dbt.adapters.catalogs._exceptions import (
+    DbtCatalogIntegrationAlreadyExistsError,
+    DbtCatalogIntegrationNotFoundError,
+    DbtCatalogIntegrationNotSupportedError,
+)
+from dbt.adapters.catalogs._integration import (
+    CatalogIntegration,
+    CatalogIntegrationConfig,
+)
+
+
+class CatalogIntegrationClient:
+    """
+    A repository class that manages catalog integrations
+
+    This class manages all types of catalog integrations,
+    supporting operations like registering new integrations and retrieving existing ones.
+    There is only one instance of this class per adapter.
+
+    Attributes:
+        __supported_catalogs (Dict[str, Type[CatalogIntegration]]): a dictionary of supported
+            catalog types mapped to their corresponding factory classes
+        __catalog_integrations (Dict[str, CatalogIntegration]): a dictionary of catalog
+            integration names mapped to their instances
+    """
+
+    def __init__(self, supported_catalogs: Dict[str, Type[CatalogIntegration]]):
+        self.__supported_catalogs = supported_catalogs
+        self.__catalog_integrations: Dict[str, CatalogIntegration] = {}
+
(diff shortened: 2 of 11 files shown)


dbt-adapters/src/dbt/adapters/base/catalog.py
```diff
@@ -0,0 +1,58 @@
...
+    glue = "glue"
+    unity = "unity"
+
+
+@dataclass
+class CatalogIntegrationConfig:
+    catalog_name: str
+    integration_name: str
+    table_format: str
+    catalog_type: str
```
colin-rogers-dbt, line 23:
```suggestion
    integration_type: str
```

dbt-adapters/src/dbt/adapters/clients/catalogs.py
```diff
@@ -0,0 +1,40 @@
...
+    def integrations(self) -> dict[str, CatalogIntegrationProtocol]:
+        return self._integrations
+
+    def add_integration(self, integration: CatalogIntegrationProtocol, catalog_name: str):
+        if catalog_name in self._integrations:
+            raise DbtCatalogIntegrationAlreadyExistsError(catalog_name)
+        self._integrations[catalog_name] = integration
+
+
+_CATALOG_CLIENT = CatalogIntegrations()
```
colin-rogers-dbt, line 27: refactor to be a class instance attached to the adapter

dbt-adapters/src/dbt/adapters/catalogs/_integration.py
```diff
@@ -0,0 +1,71 @@
...
+            - this is determined by the catalog integration type, hence it is a class attribute
+        table_format (Optional[str]): the table format this catalog uses
+            - this is commonly determined by the catalog integration type, hence it is usually a class attribute
+            - it should only be required from the user for catalogs that support multiple formats
+        external_volume (Optional[str]): external storage volume identifier
+            - while this is a separate concept from catalogs, we feel it is more user-friendly to group it with the catalog configuration
+            - a result of this grouping is that there can only be one external volume per catalog integration, but many catalogs can share the same volume
+    """
+
+    allows_writes: bool = False
```
aranke, line 63: Can we change this to an enum for future-proofing?

dbt-adapters/src/dbt/adapters/catalogs/_integration.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def __init__(self, config: CatalogIntegrationConfig):
+        self.name: str = config.name
+        self.catalog_type: str = config.catalog_type
+        if config.table_format:
+            self.table_format: str = config.table_format
+        if config.external_volume:
+            self.external_volume: Optional[str] = config.external_volume
```
aranke, line 71:
```suggestion
self.table_format: Optional[str] = config.table_format or None
self.external_volume: Optional[str] = config.external_volume or None
```

aranke (approved):
Mostly LGTM, some small comments.
No need for re-review from me.

dbt-adapters/src/dbt/adapters/base/impl.py
```diff
@@ -295,6 +302,19 @@ def __init__(self, config, mp_context: SpawnContext) -> None:
         self._macro_resolver: Optional[MacroResolverProtocol] = None
         self._macro_context_generator: Optional[MacroContextGeneratorCallable] = None
         self.behavior = DEFAULT_BASE_BEHAVIOR_FLAGS  # type: ignore
+        self.catalog_client = CatalogIntegrationClient(self.CATALOG_INTEGRATIONS)
+
+    def add_catalog_integrations(
```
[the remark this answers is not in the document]
mikealfare: I'm fine removing this, I only had it because that's how this started; then I realized that we might want to add one as well.

dbt-adapters/src/dbt/adapters/catalogs/_integration.py
mikealfare: I'm having a difficult time coming up with an alternative mechanism. What would the values be? Do we want to roll reads/writes into one attribute? Is there another state I'm not thinking of?

dbt-adapters/src/dbt/adapters/catalogs/_integration.py
mikealfare: I want to allow `table_format` to default to the class attribute if not provided. I'm pretty sure this would force the user to provide a value, or the maintainer would need to update the config to always pass the default value. I think this works for `external_volume` though. I went back and forth on these because they're not always required but we're specifying them in the protocol.

dbt-adapters/src/dbt/adapters/catalogs/_client.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Attributes:
+        __supported_catalogs (Dict[str, Type[CatalogIntegration]]): a dictionary of supported
+            catalog types mapped to their corresponding factory classes
+        __catalog_integrations (Dict[str, CatalogIntegration]): a dictionary of catalog
+            integration names mapped to their instances
+    """
+
+    def __init__(self, supported_catalogs: Dict[str, Type[CatalogIntegration]]):
+        self.__supported_catalogs = supported_catalogs
```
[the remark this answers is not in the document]
mikealfare: @aranke What does the name mangling affect? From what it looks like, this is the intended behavior from Python; it obfuscates the attribute so that it's harder to find (since nothing is truly private). But is there a process that I'm unaware of that relies on this variable being referenced as configured?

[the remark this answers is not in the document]
aranke: I don't think it really affects anything, just not obvious.

dbt-adapters/src/dbt/adapters/catalogs/_integration.py
```diff
@@ -0,0 +1,71 @@
+from typing import Any, Dict, Optional
+from typing_extensions import Protocol
+
+
+class CatalogIntegrationConfig(Protocol):
```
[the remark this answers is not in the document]
aranke: From Claude:

> I understand where you're coming from, but I think there might be a misunderstanding about what Protocols are designed for.
> 
> Protocols indeed allow objects to pass instance checks without explicit inheritance, but they're specifically intended for defining interfaces with methods (behavioral interfaces). What we have here is purely a data structure with no methods.
> 
> While dataclasses do come with some auto-generated methods (like `__init__`, `__repr__`, etc.), those are just conveniences that make working with data structures easier and more consistent.
> 
> The main issue with using Protocol here is that it sends the wrong signal to other developers. A Protocol suggests "implement this interface with these methods" rather than "here's a configuration data structure." Using a Protocol for a data-only structure goes against Python's typing conventions and could confuse future maintainers.
> 
> If structural typing without inheritance is what you're after, here's how you can use isinstance with dataclasses:

```python
from dataclasses import dataclass
from typing import Optional, Dict, Any

@dataclass
class CatalogIntegrationConfig:
    name: str
    catalog_type: str
    table_format: Optional[str] = None
    external_volume: Optional[str] = None
    adapter_properties: Optional[Dict[str, Any]] = None

# This works with a proper dataclass instance
config1 = CatalogIntegrationConfig("my_catalog", "iceberg_rest")
assert isinstance(config1, CatalogIntegrationConfig)  # True
```

dbt-adapters/src/dbt/adapters/catalogs/_integration.py
aranke:
```
from enum import Enum

class CatalogIntegrationMode(Enum):
    READ = "r"
    WRITE = "w"
```

mikealfare (approved):

PR #797, opened 2025-02-06
Fix snowflake testing

mikealfare:
We inadvertently prefixed the integration tests command with `- `, which tells `hatch` to proceed to the next command regardless of the result of the prior command. This made it so that integration tests were passing in CI despite the tests themselves failing. This hid a few broken tests. In particular, we supplied an incorrect config for Iceberg dynamic tables in one test and we failed to notice that our OAuth tokens expired. Functional code was not impacted and is working as intended.

### DIFF
--- a/dbt-snowflake/tests/functional/iceberg/models.py
+++ b/dbt-snowflake/tests/functional/iceberg/models.py
@@ -70,8 +70,9 @@ select * from {{ ref('first_table') }}
 _MODEL_BASIC_DYNAMIC_TABLE_MODEL_WITH_PATH = """
 {{
   config(
-    transient = "transient",
     materialized = "dynamic_table",
+    snowflake_warehouse='DBT_TESTING',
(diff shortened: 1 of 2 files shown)


dbt-snowflake/tests/functional/iceberg/models.py
```diff
@@ -70,8 +70,9 @@
 _MODEL_BASIC_DYNAMIC_TABLE_MODEL_WITH_PATH = """
 {{
   config(
-    transient = "transient",
     materialized = "dynamic_table",
+    snowflake_warehouse='DBT_TESTING',
```
mikealfare, line 74: It looks like the base was copied from the table materialization model and updated with `dynamic_table` instead of copying from the Iceberg materialization and adding the new `base_location_root` config option.

colin-rogers-dbt (approved):

PR #800, opened 2025-02-07
Standardize `mypy` config

mikealfare:
Code quality happens via pre-commit at the repo root, so `mypy` config does not need to live in each repo. This config is moved to `pre-commit-config.yml` now. Implementing this config caught several typecheck error. Most were marked with `# type:ignore`, to be resolved later. This is fine since this is effectively current state.


dbt-bigquery/third-party-stubs/agate/data_types.pyi
colin-rogers-dbt, line 1: We don't need these stubs anymore?

.pre-commit-config.yaml
```diff
@@ -46,13 +46,15 @@ repos:
     rev: v1.11.2
     hooks:
     -   id: mypy
-        exclude: dbt-adapters/src/dbt/adapters/events/adapter_types_pb2.py|dbt-tests-adapter/src/dbt/__init__.py
```
colin-rogers-dbt, line 49: don't we want to exclude _pbt2.py?

dbt-bigquery/third-party-stubs/agate/data_types.pyi
mikealfare: `mypy` runs without them. I was surprised too, but it's not balking at it.

.pre-commit-config.yaml
mikealfare: Again, I thought this would be an issue, but it's somehow no longer an issue. I was concerned it wasn't running at all, but it wouldn't have generated all of the conflicts that I found if it weren't running. And it found issues in the `dbt-adapters` package.

colin-rogers-dbt (approved):

PR #801, opened 2025-02-07
Fix publish workflow conditions

mikealfare:
- fix typo in fromJSON
- hard code values in conditionals to avoid redirection

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -54,7 +54,7 @@ defaults:
 jobs:
     unit-tests:
         if: |
-            !contains(fromJson(vars.SKIP_UNIT_TESTS), inputs.package) &&
+            !contains(fromJSON('["dbt-tests-adapter"]'), inputs.package) &&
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -54,7 +54,7 @@ defaults:
 jobs:
     unit-tests:
         if: |
-            !contains(fromJson(vars.SKIP_UNIT_TESTS), inputs.package) &&
```
colin-rogers-dbt, line 57: You could make `SKIP_UNIT_TESTS` an env variable in the workflow like:
```
env:
   SKIP_UNIT_TESTS='["dbt-tests-adapter"]

mikealfare: It crossed my mind to do that but I was concerned that it would be over-engineering it again, which is sort of what led here. But since you're agreeing with that, I'll do that.

colin-rogers-dbt (approved):

PR #804, opened 2025-02-07
ADAP-1236: Add bump version workflow

mikealfare:
We have been manually bumping the version, but there should be a workflow for this to reduce the maintenance burden for on-call.


.github/workflows/publish-oss.yml
```diff
@@ -44,41 +48,40 @@ defaults:
...
             inputs.skip-integration-tests == false &&
             !contains(fromJSON('["dbt-adapters", "dbt-tests-adapter"]'), inputs.package)
+        needs: bump-version  # bump-version can be skipped and is not directly used
         uses: ./.github/workflows/_integration-tests.yml
         with:
             packages: ${{ toJSON(inputs.package) }}
             branch: ${{ inputs.branch }}
         secrets: inherit
 
-    publish-prep-checks:
```
mikealfare, line 66: This is replaced by:
```yaml
if: always() && !failure()
```

mikealfare:
It worked! https://test.pypi.org/project/dbt-athena/1.9.2/

.github/workflows/publish-oss.yml
```diff
@@ -25,6 +24,10 @@ on:
                 description: "Choose the branch to publish from"
                 type: string
                 default: "main"
+            version:
+                description: "Does the version need to be bumped?"
+                type: string
+                default: ""
```
emmyoop, line 30: This reads as a yes/no question but it is looking for how to bump the version, right?

.github/workflows/publish-oss.yml
```diff
@@ -44,41 +47,41 @@ defaults:
...
-            with:
-                jobs: ${{ toJSON(needs) }}
-                allowed-skips: ${{ toJSON(needs) }}
-
     generate-changelog:
-        needs: publish-prep-checks
-        if: |
-            always() &&
-            needs.publish-prep-checks.result == 'success'
+        if: always() && !failure()
```
emmyoop, line 83: This is right - but makes me dislike the logic handling of workflows a little more each time I see it.

.github/workflows/_bump-version.yml
```diff
@@ -0,0 +1,80 @@
...
+                type: string
+                default: "main"
+            version:
+                description: "Choose the version to bump to (e.g. 1.2.3rc1, patch, release)"
+                type: string
+                default: "beta"
+        secrets:
+            FISHTOWN_BOT_PAT:
+                description: "Token to commit/merge changes into branches"
+                required: true
```
emmyoop, line 47: You don't need this defined for the workflow_dispatch trigger.  Wither this trigger all secrets available in the context are just available.  They just need to be defined for workflow_call.

.github/workflows/_bump-version.yml
```diff
@@ -0,0 +1,80 @@
...
+
+permissions:
+    contents: write
+
+defaults:
+    run:
+        shell: bash
```
emmyoop, line 54: This is optional to define.  The ubuntu runner already defaults to use bash.  You have to define it for actions since they don't have a runner defined in them.

.github/workflows/publish-oss.yml
mikealfare: All CI/CD tools are wrong, but some are useful.

emmyoop (approved):

PR #817, opened 2025-02-13
Fix pull request integration tests

mikealfare:
We deserialized the affected adapters JSON too early, resulting in passing a sequence instead of a raw JSON string.

While making this change, I noticed that `dbt-athena-community` was not pointing at `dbt-athena`'s tests and `dbt-redshift` integration tests were not triggered by macros updates in `dbt-postgres`.


mikealfare:
I created an example of this working by creating a second PR that merges into this branch to trigger `pull_request_target`. Here's the run: https://github.com/dbt-labs/dbt-adapters/actions/runs/13314899015?pr=818

colin-rogers-dbt (approved):

PR #826, opened 2025-02-18
ADAP-1190: Test support bigframes

jialuoo:
resolves #

### Checklist


dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
```diff
@@ -163,3 +176,181 @@ def _update_batch_from_config(
...
+
+        notebook_execution_job.service_account = (
+            self._GoogleCredentials._service_account_email
+        )
+        notebook_execution_job.gcs_output_uri = "gs://{}/{}/logs".format(
+            self._credentials.gcs_bucket, self._model_file_name
+        )
+        notebook_execution_job.display_name = self._get_batch_id()
+        request = aiplatform_v1beta1.CreateNotebookExecutionJobRequest(
+            parent=f"projects/{self._credentials.execution_project}/locations/{self._credentials.bigframes_region}",
```
mikealfare, line 338: As mentioned above, I feel like there must be a way to pass in `project` and `region` as kwargs, given how ubiquitous they are. I'd rather rely on the referenced library's knowledge of this path than hard code it here.

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
```diff
@@ -163,3 +176,181 @@ def _update_batch_from_config(
...
+        gcs_log_uri = (
+            f"{notebook_execution_job.gcs_output_uri}/{job_id}/{self._model_name}.py"
+        )
+        gcs_log = self._read_json_from_gcs(gcs_log_uri)
+        # TODO(jialuo): improve the logger info.
+        _logger.info(
+            f"The colab notebook runtime outputs from GCS: {gcs_log['cells'][0]['outputs']}"
+        )
+
+        _ = self._ai_platform_client.get_notebook_execution_job(name=res.name)
```
mikealfare, line 356: Is there a default polling retry on this? I would have expected that we use `self._result_polling_policy` here. Looking back, I don't see where else that gets used.

Separately, we don't need to specify `_ =` here if we're not doing anything with the result.

mikealfare (commented):
I left a ton of comments, many of them thinking out loud more than anything. We can discuss at our next sync if you'd like.

I know you mentioned that you got this working. I'd like to have some tests for this to demonstrate this functionality and to lock it in to prevent regressions later. Do you have any tests you would recommend? I can update our test instance with any additional configuration that is needed to test this in CI.

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -110,6 +111,7 @@ class BigqueryConfig(AdapterConfig):
     max_staleness: Optional[str] = None
     enable_list_inference: Optional[bool] = None
     intermediate_format: Optional[str] = None
+    dataframe_python_syntax: Optional[str] = "spark"
```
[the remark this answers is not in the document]
amychen1776: What are the possible inputs other than spark?

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
```diff
@@ -163,3 +179,141 @@ def _update_batch_from_config(
...
+            # TODO(jialuo): improve the logger info.
+            # TODO(jialuo): raise errors here. There are some situations when
+            # the notebook failed but the pipeline still showed success.
+            _logger.info(
+                f"The colab notebook runtime outputs from GCS: {gcs_log['cells'][0]['outputs']}"
+            )
+        else:
+            _logger.debug(f"Failed to read log from GCS URI: {gcs_log_uri}")
+
+        _ = self._ai_platform_client.get_notebook_execution_job(name=res.name)
```
mikealfare, line 319: Does this specifically wait for the job to complete? Or is it more of a status check? Our current approach is to wait for the job to complete so that we know we can run downstream models.

Also, can we return the job here? In the two dataproc submission methods we return the Job and the Batch, respectively, during `submit`. I'm not sure how this gets used off the top of my head, but it would be consistent and could make testing and troubleshooting a little easier.

mikealfare (commented):
This is looking much better. I still have a few more points of feedback but it's coming along. At this point we might want to start adding some tests to make sure this is all working together as expected (and continues to work into the future).

mikealfare:
This was implemented in #921. Closing as a duplicate.

PR #827, opened 2025-02-18
handle new is_iceberg col

colin-rogers-dbt:
With the new [snowflake bundle](https://docs.snowflake.com/en/release-notes/bcr-bundles/2025_01_bundle) we need to handle cases for the time being where this column is/not present in `show objects`. 

Once this is the default we can remove this branching and remove the behavior flag. 

resolves #

### Checklist

### DIFF
--- a/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
@@ -283,7 +283,7 @@ class SnowflakeAdapter(SQLAdapter):
         if self.behavior.enable_iceberg_materializations.no_warn:
             database, schema, identifier, relation_type, is_dynamic, is_iceberg = result
         else:
-            if "is_iceberg" in result.column_names:
+            if "is_iceberg" in result:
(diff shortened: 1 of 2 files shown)


VersusFacit:
can we enable the new bundle on the adapter snowflake cluster so this code path is hit?

amychen1776 (commented):
Written with this assumption that if they turn off the behavior flag, we will get the metadata by default. Please correct me if I'm wrong. :::

We should change the behavior Flag message to call out that if you have activated [2025_01](https://docs.snowflake.com/en/release-notes/bcr-bundles/2025_01_bundle), there is no need for using this behavior flag. We will deprecate this flag in the next minor release of dbt-snowflake.

VersusFacit:
> We should change the behavior Flag message to call out that if you have activated [2025_01](https://docs.snowflake.com/en/release-notes/bcr-bundles/2025_01_bundle), there is no need for using this behavior flag. We will deprecate this flag in the next minor release of dbt-snowflake.
+1. We can easily include this in the PR since we're in monorepo land :)

PR #828, opened 2025-02-19
ADAP-1231: Automate internal releases

mikealfare:
We should release internally on a regular basis to ensure that we're always on `main` in Cloud. Automated tests will prevent us from inadvertently releasing a broken package in the event that CI checks were not sufficient.

### DIFF
--- a/.github/workflows/_publish-internal.yml
+++ b/.github/workflows/_publish-internal.yml
@@ -6,11 +6,7 @@ on:
             package:
                 description: "Choose the package to publish"
                 type: string
-                default: "dbt-adapters"
-            deploy-to:
-                description: "Choose whether to publish to test or prod"
-                type: string
-                default: "prod"
+                required: true
             branch:
                 description: "Choose the branch to publish"
                 type: string
@@ -21,21 +17,21 @@ on:
                 description: "Choose the package to publish"
                 type: choice
                 options:
-                -   "dbt-adapters"
                 -   "dbt-athena"
                 -   "dbt-bigquery"
                 -   "dbt-postgres"
                 -   "dbt-redshift"
                 -   "dbt-snowflake"
                 -   "dbt-spark"
-            deploy-to:
-                description: "Choose whether to publish to test or prod"
-                type: environment
-                default: "test"
             branch:
                 description: "Choose the branch to publish"
                 type: string
                 default: "main"
+
+# don't publish to the same target in parallel
+concurrency:
+    group: PyPI_Internal-${{ inputs.package }}
+    cancel-in-progress: true
 
 defaults:
     run:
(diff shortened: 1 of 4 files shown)


.github/workflows/_publish-internal.yml
```diff
@@ -6,11 +6,7 @@ on:
             package:
                 description: "Choose the package to publish"
                 type: string
-                default: "dbt-adapters"
-            deploy-to:
-                description: "Choose whether to publish to test or prod"
-                type: string
-                default: "prod"
+                required: true
```
mikealfare, line 9: We don't publish `dbt-adapters` internally, at least not yet, hence it should not be a default.

.github/workflows/_publish-internal.yml
```diff
@@ -6,11 +6,7 @@ on:
             package:
                 description: "Choose the package to publish"
                 type: string
-                default: "dbt-adapters"
-            deploy-to:
```
mikealfare, line 10: We don't publish to a test repository, so this is always "prod".

.github/workflows/_publish-internal.yml
```diff
@@ -21,22 +17,22 @@ on:
...
-                description: "Choose whether to publish to test or prod"
-                type: environment
-                default: "test"
             branch:
                 description: "Choose the branch to publish"
                 type: string
                 default: "main"
 
+# don't publish to the same target in parallel
```
mikealfare, line 31: This concurrency should be here, instead of copy/pasted in multiple downstream callers.

.github/workflows/publish-internal.yml
```diff
@@ -0,0 +1,68 @@
+name: "Publish to internal"
```
mikealfare, line 1: This workflow was created separately because we need to run tests and publish all in isolated streams. This prevents failing tests from one adapter blocking the release of another adapter.

.github/workflows/publish-oss.yml
```diff
@@ -25,14 +25,6 @@ on:
                 description: "Choose the branch to publish from"
                 type: string
                 default: "main"
-            pypi-internal:
```
mikealfare, line 28: Since we pulled out publishing internally, the only reason to run this workflow is to publish to OSS. Hence neither of these inputs are needed.

.github/workflows/publish-scheduled.yml
```diff
@@ -0,0 +1,23 @@
+name: "Scheduled publishing"
+run-name: "Publish to internal - Scheduled"
+
+on:
+    schedule:
+    -   cron: "0 0 * * SUN"
```
mikealfare, line 6: This will run 3hr prior to the daily scheduled tests. This should be enough time to avoid collisions.

.github/workflows/publish-oss.yml
mikealfare, line 1: Should we consider testing more broadly since this is now focused on OSS releases? Or should we rely on scheduled tests which cover all supported scenarios and use this more as a sanity check?

mikealfare (commented):
Here's a self-review with some rationale and guidance for the decisions being made.

colin-rogers-dbt (approved):

PR #832, opened 2025-02-21
ADAP-1233: Fix the databricks startup fixture

mikealfare:
We need to ensure the databricks cluster is ready before running tests. The current method requires the `project` fixture, which makes a call to create a test schema. This call would also need to wait, so we cannot rely on the `project` fixture.

Instead, we will vendor some of the connection functionality and use non-dbt methods to call the cluster directly.

While doing this work, I moved some fixtures around to make them a bit more organized as well as scope them appropriately. Specifically, I moved all of the profile fixtures to the functional tests level so they don't run as part of unit tests. I also put these fixtures in a separate module to remove noise from co-locating them with the databricks start-up fixture.

I added an `__init__.py` file in the functional tests directory to assist with test discovery. Omitting this creates the potential for name collisions, which we traditionally handle by prefixing additional quantifiers everywhere.

### DIFF
--- a/dbt-spark/tests/functional/fixtures/start_databricks_cluster.py
+++ b/dbt-spark/tests/functional/fixtures/start_databricks_cluster.py
@@ -11,7 +11,7 @@ except ImportError:
 
 from dbt.adapters.spark.connections import SparkConnectionManager
 
-from .profiles import databricks_http_cluster_target
+from tests.functional.fixtures.profiles import databricks_http_cluster_target
 
 
 # Running this should prevent tests from needing to be retried because the Databricks cluster isn't available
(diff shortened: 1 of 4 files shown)


dbt-spark/tests/functional/fixtures/start_databricks_cluster.py
```diff
@@ -0,0 +1,65 @@
...
+        try:
+            cursor.execute("SELECT 1", async_=False)
+            return
+        except Exception:
+            sleep(30)
+
+    raise Exception("Databricks cluster did not start in time")
+
+
+def _cursor():
```
colin-rogers-dbt, line 51: your code is clever
yet fills my heart with sadness
we do what we must

colin-rogers-dbt (approved):

dbt-spark/tests/functional/fixtures/start_databricks_cluster.py
mikealfare: I aim to please.

mikealfare:
The failing tests are known issues.

PR #744, opened 2025-01-27
Add null equality condition to equals macro

VersusFacit:
resolves #110 

### Problem

testing things here

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #151, opened 2024-04-10
Fix incremental delete+insert SQL

ataft:
resolves dbt-labs/dbt-adapters#150

### Problem
The delete query for the 'delete+insert' incremental_strategy with 2+ unique_key columns is VERY inefficient. In many cases, it will hang and never return for deleting small amounts of data (<100K rows).

### Solution
Improve the query by switching to a much more efficient delete strategy:

```
delete from table1
where (col1, col2) in (
    select distinct col1, col2 from table1_tmp
)
```

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-adapter/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development, and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc.) or this PR has already received feedback and approval from Product or DX

PR #834, opened 2025-02-21
Improve performance of `delete+insert` incremental strategy with null equality check changes

adrianburusdbt:
This PR merges the changes for null equality from here: https://github.com/dbt-labs/dbt-adapters/pull/744 into existing PR: https://github.com/dbt-labs/dbt-adapters/pull/151?email_source=slack

### Checklist

### DIFF
--- a/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
+++ b/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
@@ -61,36 +61,23 @@
     {%- set dest_cols_csv = get_quoted_csv(dest_columns | map(attribute="name")) -%}
 
     {% if unique_key %}
-        {% if unique_key is sequence and unique_key is not string %}
-            delete from {{ target }}
-            using {{ source }}
-            where (
-                {% for key in unique_key %}
-		    {% set source_unique_key = (source ~ "." ~ key) | trim %}
-		    {% set target_unique_key = (target ~ "." ~ key) | trim %}
-                    {{ equals(source_unique_key, target_unique_key) }}
-                    {{ "and " if not loop.last}}
-                {% endfor %}
-                {% if incremental_predicates %}
-                    {% for predicate in incremental_predicates %}
-                        and {{ predicate }}
-                    {% endfor %}
-                {% endif %}
-            );
-        {% else %}
-            delete from {{ target }}
-            where (
-                {{ unique_key }}) in (
-                select ({{ unique_key }})
-                from {{ source }}
-            )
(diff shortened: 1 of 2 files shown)


VanTudor:
Code lgtm, but let's address those failing tests

dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
```diff
@@ -61,36 +61,23 @@
...
-
+        {% if unique_key is string %}
+        {% set unique_key = [unique_key] %}
         {% endif %}
+
+        {%- set unique_key_str = unique_key|join(', ') -%}
+
+        delete from {{ target }}
+        where ({{ unique_key_str }}) in (
+            select distinct {{ unique_key_str }}
```
mikealfare, line 72: I think the `distinct` here is unnecessary and would make it take longer since it needs to reduce dupes. I don't have proof of that, but I generally don't use `distinct` when using a `where <x> in <y>` clause.

mikealfare (commented):
Overall this is much cleaner. Assuming tests are happy then this looks good.

dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/merge.sql
VersusFacit: Hey Mike, so this is actually turned out to be a really interesting problem to me. `distinct` should be faster in theory, but in practice, cardinality of the subquery's result set is key.

Higher cardinality means the DBMS will do far more scans of the target result set than needed.

Lower cardinality means the deduplication step is irrelevant and hence a performance lowering event.

We can't predict the cardinality for the unique key strings that users pick, but we do note in docs that users should employ `delete+insert` for more efficient merges in the context of high cardinality sets. Hence, let's default to `distinct` as always present for this workflow and avoid the possibility that indexing/algorithms under the hood fail to prune extraneous unmatched rows in the result set.

VersusFacit (approved):
Researched the implications and tests passing. This is judged to be functionally equivalent to the original macro, the performance benefits notwithstanding.

jhsb25:
any ideas when this will be available, delete+insert are painfully slow currently

VersusFacit:
@jhsb25 should hit Cloud in about a week upper bound at current pace unless we had some rollbacks for some reason

will-sargent-dbtlabs:
@VersusFacit -- Any idea if this was included in core 1.9.3?  The core release notes were pretty slim and don't mention an adapters bump, so I'm a bit unclear.. Thanks!

tagging: @LProcopi15

PR #838, opened 2025-02-21
ADAP-1235: Fix test_persist_docs missing column test

mikealfare:
We are looking for a specific string in the exception that Spark throws. It looks like that description has changed but the exception is the same. This just updates what we expect.

### DIFF
--- a/dbt-spark/tests/functional/adapter/persist_docs/test_persist_docs.py
+++ b/dbt-spark/tests/functional/adapter/persist_docs/test_persist_docs.py
@@ -168,4 +168,3 @@ class TestPersistDocsMissingColumn:
                 "Missing field name in table" in res[0].message,
             ]
         )
-        assert "name `name` cannot be resolve" in res[0].message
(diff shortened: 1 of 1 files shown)


dbt-spark/tests/functional/adapter/persist_docs/test_persist_docs.py
```diff
@@ -150,7 +150,17 @@ def properties(self):
...
+            ('42000', '[42000] [Simba][Hardy] (80) Syntax or semantic analysis error thrown in server while executing query.
+            Error message from server:
+                org.apache.hive.service.cli.HiveSQLException: Error running query: [UNRESOLVED_COLUMN.WITH_SUGGESTION]
+                org.apache.spark.sql.AnalysisException: [UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter with name `name` cannot be resolve (80) (SQLExecDirectW)'
+            )
+        """
         run_dbt(["seed"])
         res = run_dbt(["run"], expect_pass=False)
-        assert "Missing field name in table" in res[0].message
+        assert "[UNRESOLVED_COLUMN.WITH_SUGGESTION]" in res[0].message
```
colin-rogers-dbt, line 165: looks like our databricks environments still raise the old message so we might want to check if either is true:
```suggestion
        assert any("[UNRESOLVED_COLUMN.WITH_SUGGESTION]" in res[0].message, "Missing field name in table" in res[0].message)
```

mikealfare: Yeah, I noticed that. I think it's based on profile. But if you're fine with the "or" scenario, so am I.

VanTudor (approved):

PR #840, opened 2025-02-22
ADAP-1229: Update workflows with run names that include inputs

mikealfare:
It's difficult to tell what the inputs were for workflows, so let's put them in the run name.


.github/workflows/_changelog-entry-check.yml
```diff
@@ -1,4 +1,5 @@
-name: "Changelog entry check"
+name: "# Changelog entry check"
```
colin-rogers-dbt, line 1: What's the significance of the `#`?

mikealfare: It's alphabetically after letters

mikealfare: As it turns out, GitHub does not care about this and they still show up on top. I guess all we have is pinning.

PR #855, opened 2025-02-28
Mp/add insert overwrite incremental strategy

### LINKED ISSUE #736: [ADAP-432] [Feature] Snowflake Adapter supports INSERT OVERWRITE

- [X] I am requesting a straightforward extension of existing dbt-snowflake functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Snowflake has an optional parameter for inserts which does TRUNCATE behavior before INSERT. It's one transaction and avoids dropping or recreating tables.

https://docs.snowflake.com/en/sql-reference/sql/insert#optional-parameters
`INSERT [ OVERWRITE ] INTO <target_table>`

Support this feature in dbt

There is no single-transaction alternative feature to TRUNCATE / DELETE * and then INSERT. This is documented by Snowflake.

### Who will this benefit?

Anyone who needs to avoid DROP/RECREATE and at the same time wants to full load will benefit, with the added bonus of keeping the behavior single-transaction.

In our use case, we don't want tables to be dropped as the table has permissions and policies which need to be kept in place despite which process is doing the load. At the same time, the rows of the table are completely replaced during load.

Yes, I need to understand materializations better but in my naivete this seems like a simple and valuable feature

### Anything else?
Matt Winkler already built this:
(https://github.com/matt-winkler/dbt_workspace/blob/master/dbt/macros/materializations/snowflake_custom_incremental/insert_overwrite.sql)](https://github.com/matt-winkler/dbt_workspace/tree/master/dbt/macros/materializations/snowflake_custom_incremental)

VersusFacit:
resolves #736 

### Problem

We need to support `insert overwrite` for key workflows that can't yet take advantage of this DDL.

### Solution

After some back and forth, we've settled on `insert_overwrite` being a pseudo incremental strategy to preserve the `materialized=table` interface and instead trust the user to know that this is an atypical use of an incremental model. In other DBMSs, it's clearly an incremental model workflow, just not on Snowflake where it's more of a truncate + insert. 

I looked into -- this is for both standard and iceberg tables. The same DDL makes this easy :) 

### Checklist


colin-rogers-dbt (approved):

markproctor1:
@VersusFacit Thanks! I'm testing this now at MGB.

1. First Run, builds the table no problem
`
{{
    config(
        materialized='incremental',
        incremental_strategy='insert_overwrite',
    )
}}
select 1 as one`

2. Second Run, simple build again, fails:
Compilation Error in model ... parameter 'predicates' was not provided
`
{{
    config(
        materialized='incremental',
        incremental_strategy='insert_overwrite',
    )
}}
select 1 as one`

I propose this fix, passing an empty `predicates`.
`{% macro get_incremental_insert_overwrite_sql(arg_dict) -%}
  {{ adapter.dispatch('insert_overwrite_get_sql', 'dbt')(
      arg_dict["target_relation"],
      arg_dict["temp_relation"],
      arg_dict["unique_key"],
      arg_dict["dest_columns"],
      arg_dict.get("predicates", [])
  ) }}
{%- endmacro %}`

For Snowflake specifically, we don't want predicates on an INSERT OVERWRITE.

A simple WHERE condition in the SELECT would handle any need to partially load rows.

The mechanism for INSERT_OVERWRITE in Snowflake of course is single-transaction delete+insert.

PR #857, opened 2025-02-28
Add new logic to handle is_iceberg duplicates and avoid unpacking errors

VersusFacit:
### Problem

We had this cause an incident upon adoption of bundle `2025_01` and since then with the workaround of having the iceberg support flag on, it still manifests as the following warning:

```
DuplicateColumnWarning: Column name "is_iceberg" already exists in Table. Column will be renamed to "is_iceberg_2".
```

### Solution

1. Ensure there is always an `is_iceberg` column when the behavior flag is set
2. Remove it if the flag is not enabled
3. Catch the cases where the flag is off and users are on `2025_01` which auto includes it

I tested this manually for iceberg tests on `2024_08`, then flipped our cluster over to `2025_01`.

image

Current state as of making this PR!

### Checklist


dbt-snowflake/src/dbt/include/snowflake/macros/adapters.sql
```diff
@@ -150,7 +150,7 @@
     {# -- Gated for performance reason. If you don't want Iceberg, you shouldn't pay the
        -- latency penalty. #}
     {% if adapter.behavior.enable_iceberg_materializations.no_warn %}
-      select all_objects.*, is_iceberg
+      select all_objects.*, is_iceberg as is_iceberg_dbt
```
mikealfare, line 153: I think we need a comment here. I needed to re-read this whole thing a few times before I realized that the issue was that `all_objects` contains `is_iceberg` after a particular bundle is enabled. This goes back to us not conditioning on platform version (e.g. bundle), though that's certainly out of scope here.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -268,22 +268,39 @@ def list_relations_without_caching(
 
         # this can be collapsed once Snowflake adds is_iceberg to show objects
         columns = ["database_name", "schema_name", "name", "kind", "is_dynamic"]
-        if self.behavior.enable_iceberg_materializations.no_warn:
+
+        # If the is_iceberg column is already on schema_objects query, we can ignore
+        # normalizing using our homemade is_iceberg
+        if (
```
mikealfare, line 274: I don't think this conditional is quite right. We only want to do any renaming if iceberg materializations are enabled. And then within that condition, we either need to rename the all caps version, as is currently the process, or rename the dbt-calculated columns, as you do below. And then we only append `is_iceberg` if iceberg materializations are enabled. The if block below will append it when it exists, regardless of whether the user has enabled this flag.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -268,22 +268,39 @@ def list_relations_without_caching(
...
 
     def _parse_list_relations_result(self, result: "agate.Row") -> SnowflakeRelation:
-        # this can be collapsed once Snowflake adds is_iceberg to show objects
+        # We won't deprecate this for awhile as the bundle with is_iceberg will need time to be adopted
         if self.behavior.enable_iceberg_materializations.no_warn:
             database, schema, identifier, relation_type, is_dynamic, is_iceberg = result
         else:
-            database, schema, identifier, relation_type, is_dynamic = result
+            # protective removal of any additional fields
+            database, schema, identifier, relation_type, is_dynamic, *_ = result
```
mikealfare, line 303: This doesn't need to be updated if we nest the append of `is_iceberg` above.

mikealfare (commented):
I feel like we're leaking the iceberg metadata into the implementation too much; it should remain tucked behind the behavior flag.

PR #870, opened 2025-03-05
ADAP-1231: Open a lockfile PR after internal releases

mikealfare:
We should open a lockfile PR after internal releases so that we can review and approve any incoming version bumps. We should also post this PR as a Slack a notification to the PR review channel to make it easier to see.


.github/workflows/scheduled-publishing.yml
```diff
@@ -0,0 +1,50 @@
...
+                -   "dbt-spark"
+            fail-fast: false
+        uses: ./.github/workflows/publish-internal.yml
+        with:
+            package: ${{ matrix.package }}
+        secrets: inherit
+
+    open-lockfile-pr:
+        runs-on: ${{ vars.DEFAULT_RUNNER }}
+        steps:
```
emmyoop, line 27: You'll want a needs so this doesn't run until the publish job finishes

.github/workflows/scheduled-publishing.yml
```diff
@@ -0,0 +1,50 @@
...
+                GH_TOKEN: ${{ secrets.FISHTOWN_BOT_PAT }}
+                REPO: "dbt-labs/dbt-mantle"
+                WORKFLOW: "open-lockfile-pr.yml"
+                PR_TITLE: "Generate new lockfile for dbt-mantle"
+        -   uses: ravsamhq/notify-slack-action@v2
+            with:
+                token: ${{ secrets.GITHUB_TOKEN }}
+                status: ${{ job.status }}
+                notification_title: "dbt-mantle lockfile is ready for review"
+                message_format: "View PR: https://github.com/dbt-labs/dbt-adapters/pull/${{ steps.pr.outputs.number }}"
```
emmyoop, line 46: This is the wrong link.  
```suggestion
                message_format: "View PR: https://github.com/dbt-labs/dbt-mantle/pull/${{ steps.pr.outputs.number }}"
```

.github/workflows/scheduled-publishing.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    open-lockfile-pr:
+        runs-on: ${{ vars.DEFAULT_RUNNER }}
+        steps:
+        -   id: pr
+            shell: bash
+            run: |
+                RUN_ID=$(gh workflow run $WORKFLOW --repo $REPO)
+                echo $RUN_ID
+                gh run watch $RUN_ID
```
emmyoop, line 33: You can add a `--exit-status` flag here so that the step will fail with a non-zero exit code if the called workflow does.  Not common for the called workflow though, so maybe not needed.  Just makes debugging a bit easier if it does happen.

emmyoop (changes requested):
I would assume it's possible but not likely that there would be no internal release updates which means no lockfile PR would be needed?  I'm not aware of the rate of changes on adapters.  If it's possible, you should account for that and not send a slack message for a PR review.

.github/workflows/scheduled-publishing.yml
mikealfare: I initially added that more to check if the PR could be created. I thought it might fail if there are no changes in the lockfile. But I ultimately chose not to. I'm less concerned about the workflow failing due to a bug.

.github/workflows/scheduled-publishing.yml
mikealfare: Good catch; I updated this.

emmyoop (approved):
Look good!

PR #872, opened 2025-03-05
Remove type-checking dependencies from build requirements

mikealfare:
Type stubs should be listed in `.pre-commit-config.yaml`, not in build requirements.


mikealfare:
We use the type checking stubs in some places for functional code. This is not a simple swap of reqs location. I'll close this for now.

PR #873, opened 2025-03-05
fix(athena): cancelling a connection cancels the athena query

### LINKED ISSUE #406: [Bug] Cancelling dbt invocation does not cancel the query on Athena
👋🏻 Hello dbt-athena community,

### Current Behavior

When I try to cancel a dbt invocation, the invocation can be cancelled but the query is not cancelled in Athena. Therefore it can lead to concurrent requests when retrying.

### Expected Behavior

When the dbt invocation cancel process is done, the query should be cancelled in Athena

### Steps To Reproduce

1. Run a model
2. Cancel the invocation (for instance with CTRL + c)
3. Check in Athena console if the query is still running or not

### Relevant log output

_No response_

### Environment

```markdown
- OS: MacOS (the same problem in Docker image based on Debian)
- Python: 3.12.7
- dbt-core: 1.8.9
- dbt-athena: 1.8.4
```

### Additional Context

_No response_

Jrmyy:
resolves #406 

### Problem
When a dbt invocation is canceled, the query still runs in Athena, leading to error such as `Destination table already exists` and concurrent requests creating the same table (the one not cancelled and the new one relaunched by the retry).

### Solution
The issue was that the `cancel_open` in the ConnectionsManager was not implemented. I changed the way the Connection is aware of the current cursors and requests running, and creating a cancel method in the Connection cancelling all the requests of the cursors. 

### Checklist


colin-rogers-dbt:
@cla-bot recheck

Jrmyy:
👋🏻 Hello @colin-rogers-dbt, what would be the next steps for this PR? It is not critical but a little bit annoying to cancel a dbt invocation without cancelling the athena query, which leads to concurrent queries creating the same table...

PR #877, opened 2025-03-05
Add microbatch incremental strategy for dbt-athena

### LINKED ISSUE #409: [dbt-athena] Microbatch Strategy
As part of https://github.com/dbt-labs/dbt-core/issues/10624, dbt-athena needs to implement a microbatch incremental strategy.

The base tests from dbt-adapters 1.6 should also be inherited and included as part of this work.

QMalcolm:
resolves #409

### Problem

dbt-athena didn't support the microbatch incremental strategy

### Solution

Add microbatch as an incremental strategy to dbt-athena using `insert_overwrite`

### Checklist


dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -57,7 +70,9 @@
       {% endcall %}
     {%- endif -%}
     {% set build_sql = "select '" ~ query_result ~ "'" -%}
-  {% elif partitioned_by is not none and strategy == 'insert_overwrite' %}
+
+  -- Insert Overwrite and Microbatch Strategies --
+  {% elif strategy in ("insert_overwrite", "microbatch") %}
```
jessedobbelaere, line 75: Does this work with Iceberg? 
I thought Iceberg only was only officially supporting `append` and `merge` according to https://dbt-athena.github.io/docs/configuration/materializations/iceberg#incremental-tables

QMalcolm: It did not, but I've now updated the PR to also have microbatch support for iceberg tables via the merge strategy 🙂

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -57,7 +72,9 @@
       {% endcall %}
     {%- endif -%}
     {% set build_sql = "select '" ~ query_result ~ "'" -%}
-  {% elif partitioned_by is not none and strategy == 'insert_overwrite' %}
+
+  -- Insert Overwrite Strategy --
+  {% elif strategy in ("insert_overwrite") %}
```
mikealfare, line 77: Is there a reason we dropped the `partitioned_by is not none` check here?

QMalcolm: It's a duplicate check. When we get to that line partitioned_by has already been guaranteed (when the strategy is `insert_overwrite`. This is explained by the commit message of [d02a24d](https://github.com/dbt-labs/dbt-adapters/pull/877/commits/d02a24d750c7cab4b8a92671602d078116b2a862) 🙂 
> In dbt-athena's `incremental.sql`, `partioned_by` was being checked twice.
First on line 33 where if `partioned_by` is `None` and the `strategy` is
`insert_overwrite`, then the strategy gets changed to `append`. Later on
line 66 we were again checking whether `partioned_by` was `None` in reference
to using the `insert_overwrite` strategy. However, due to the line 33 conditional,
the first part of the conditional on line 66 (checking `partioned_by`) would always
evaluate true if the strategy was still `insert_overwrite`.

In the end result, `partitioned_by` is being guaranteed by [lines 34 through 46](https://github.com/dbt-labs/dbt-adapters/blob/2e0d711d4858cf0733121a1faef511953bed22d2/dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql#L34)

mikealfare (approved):

PR #879, opened 2025-03-05
Issue a warning for dbt-athena-community users to point them towards dbt-athena

mikealfare:
We will no longer support `dbt-athena-community` post-1.9.x. This package will continue to be supported under `dbt-athena`; it was migrated in the 1.8.x series.


mikealfare:
This doesn't actually trigger.

PR #880, opened 2025-03-06
Raise deprecation warning for dbt-athena-community

mikealfare:
We are pulling support for the backwards compatibility associated with `dbt-athena-community` after the `1.9` support cycle. Users will need to install `dbt-athena` directly for `1.10` and beyond. This will raise a warning if users are installing `dbt-athena` via `dbt-athena-community`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,6 @@ repos:
         -   --namespace-packages
         -   --ignore-missing-imports
         -   --warn-redundant-casts
-        -   --warn-unused-ignores
         -   --pretty
         -   --show-error-codes
         additional_dependencies:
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -52,7 +52,6 @@ repos:
         -   --namespace-packages
         -   --ignore-missing-imports
         -   --warn-redundant-casts
-        -   --warn-unused-ignores
```
mikealfare, line 55: I had to do this because `pre-commit run` is incorrectly raising "unused ignore" exceptions even though `pre-commit run --all-files` is passing. This prevents commits from passing `pre-commit` checks despite CI passing (because it uses `pre-commit run --all-files`). I can't figure out why the staged files, which should be a proper subset, fail despite the full set passing. Removing this resolves the issue with minimal risk.

VersusFacit (approved):

PR #904, opened 2025-03-12
[Regression] Revert incorrect dependency pin for redshift-connector to re-enable browser IDC auth method

mikealfare:
We inadvertently downgraded this pin when migrating to the monorepo.


VersusFacit (approved):
Per `setup.py` in https://github.com/dbt-labs/dbt-redshift/pull/950/files, 
```
        "redshift-connector>=2.1.3,<2.2",
```
looks right

PR #909, opened 2025-03-12
ADAP-1237: Mitigate the impact of flaky tests on CI and releases

mikealfare:
Failing flaky tests cause minor overhead for certain tasks, e.g CI checks and scheduled checks. They also become blocking for releases since the release branch gets cleaned up after the failure. We could keep the release branch around on failure and only clean up on success, but the hope is that simply rerunning the tests will resolve the issue in most scenarios.

- rerun tests marked as flaky if they fail using the `flaky` library
- rerun these tests once (this can be configured if we want)
- update `dbt-bigquery` to use the flaky marker just like `dbt-redshift` instead of relying on the test class starting with `TestPython`
- update the integration test workflows to finish non-flaky tests for each of these adapters first, before running flaky tests since running flaky and non-flaky tests in parallel defeats the purpose of separating them


QMalcolm (approved):
🚀

VersusFacit (approved):

PR #910, opened 2025-03-12
Add aliases DBT_INTERNAL_DEST and DBT_INTERNAL_SOURCE to the delete/insert strategy

mikealfare:
We alias the source and target tables in the merge and insert/overwrite strategies, but not in the insert/delete strategies. This appears to be an oversight.


VersusFacit (approved):

e-jones:
Heads up that for our Redshift implementation this alias addition caused our models to fail [DBT slack](https://getdbt.slack.com/archives/CJARVS0RY/p1743469833649769?thread_ts=1743469802.292739&cid=CJARVS0RY)

We pinned our `requirements.txt` to `dbt-adapters==1.14.3` and that seemed to resolve it. Not sure why Redshift would support aliases in some operations but not others as that would seem inconsistent.

ronco:
We're having the same issue as @e-jones .  We pinned the version of `dbt-adapters` to resolve it.

tadeuah:
We had the same issue, setting the `dbt-adapters==1.14.3` worked.

PR #918, opened 2025-03-14
ADAP-1250: Dynamic table metadata queries now ignore the QUOTED_IDENTIFIERS_IGNORE_CASE

### LINKED ISSUE #917: ADAP-1250: [Bug] Allow QUOTED_IDENTIFIERS_IGNORE_CASE=True for dynamic tables
### Is this a new bug?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [x] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

When enabling QUOTED_IDENTIFIERS_IGNORE_CASE=True, I cannot re-create a dynamic table. Instead I get the following error:

```
SnowflakeDynamicTableConfig.__init__() missing 6 required positional arguments: 'name', 'schema_name', 'database_name', 'query', 'target_lag', and 'snowflake_warehouse'
```
after dbt sends the following code:
```sql
select
    "name",
    "schema_name",
    "database_name",
    "text",
    "target_lag",
    "warehouse",
    "refresh_mode"
from table(result_scan(last_query_id()))
```

### Expected Behavior

The dynamic table should be created successfully.

### Steps To Reproduce

1. Set QUOTED_IDENTIFIERS_IGNORE_CASE=True
2. Create a dynamic table model
3. Run `dbt run --select my_dynamic_table` to create the dynamic table in the database
4. Run `dbt run --select my_dynamic_table` a second time to trigger the metadata query to run since the dynamic table already exists
5. 💥 

### Relevant log output

```shell

```

### Environment

```markdown
- OS:
- Python:
- dbt-adapters:
- <adapter>:
```

### Additional Context

_No response_

mikealfare:
resolves #917

### Problem

Snowflake's QUOTED_IDENTIFIERS_IGNORE_CASE setting doesn't just allow case-insensitive comparisons, it updates column resolution to uppercase for quoted columns. Aliasing a column as `"my_column"` will result in it being named `"MY_COLUMN"`. This includes Snowflake metadata and the results of `SHOW` statements.

### Solution

We can safely ignore this for metadata tables as the column names are predictable and should not experience case-sensitive collisions.

### Checklist


dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/describe.sql
```diff
@@ -10,6 +10,7 @@
 --      - catalog: the metadata associated with the iceberg catalog
 -#}
     {%- set _dynamic_table_sql -%}
+        ALTER SESSION SET QUOTED_IDENTIFIERS_IGNORE_CASE = FALSE;
```
colin-rogers-dbt, line 13: nitpick but shouldn't this be lowercase?

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/describe.sql
```diff
@@ -30,6 +31,8 @@
         {% set _ = results.update({'catalog': run_query(_get_describe_iceberg_catalog_sql(relation))}) %}
     {% endif %}
 
+    ALTER SESSION UNSET QUOTED_IDENTIFIERS_IGNORE_CASE;
```
colin-rogers-dbt, line 34: is there any chance this causes issues with other models if we re-use connections? Does this impact globally setting  `QUOTED_IDENTIFIERS_IGNORE_CASE`?

mikealfare: This only alters it for the session. And it appears a session is short lived for us. I needed to use a pre-hook in the test just for it to run as part of the same session. Simply running the statement before running dbt did not "stick". I think the risk is relatively low that it would impact anything else.

colin-rogers-dbt (approved):

PR #931, opened 2025-03-21
relax constraint on google-cloud-storage to allow newer versions

### LINKED ISSUE #930: [Bug] Unable to install recent version of google-cloud-storage
### Is this a new bug?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [x] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

I have some other code interacting with GCS and wanted to install a recent version of google-cloud-storage. I found that dbt-bigquery pins the version of google-cloud-storage to 2.4, which is three years old and 19 versions behind current.

### Expected Behavior

I can install google-cloud-storage 3.1.0 (current) and future versions.

### Steps To Reproduce

1. requirements.txt with:

```
google-cloud-storage>=3.1.0
dbt-bigquery>=1.9.1
```

2. `pip install -r requirements.txt`
3. Conflict

Image

### Relevant log output

```shell

```

### Environment

```markdown
- OS: Ubuntu (via GitHub Codespaces)
- Python: 3.12.1
- dbt-adapters: 1.14.3
- dbt-bigquery: 1.9.1 
(these are just the latest versions at time of writing, nothing special about them)
```

### Additional Context

_No response_

brabster:
resolves #930
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#N/A

### Problem

Loosens version constraint on dbt-bigquery.

### Solution

google-cloud-storage>=2.4.0 allows the old versions of the current constraint, but will default to and allow the most recent version.

Looking through the history, it appears there was [a conflict-related problem back in Dec 2022 that caused this version constraint to be tightened](https://github.com/dbt-labs/dbt-adapters/commit/537b56d10fcd897660ad21b5dcd35575c04bbaa4#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R78). Dependency resolution is working fine with this constraint for me in 2025, so I would like the option as a consumer of using the latest version of this dependency if it no longer causes problems.

I have run unit tests to look for any significant API changes and they all pass for me.

### Checklist


dbt-bigquery/pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "dbt-adapters>=1.7,<2.0",
     # 3.20 introduced pyarrow>=3.0 under the `pandas` extra
     "google-cloud-bigquery[pandas]>=3.0,<4.0",
-    "google-cloud-storage~=2.4",
+    "google-cloud-storage>=2.4",
```
VersusFacit, line 30: We'd be happy to merge this if you could hard cap this from above on the major. We don't know how the latest will play with older pinned dependencies so it's safer to pin from above on the major.

```suggestion
    "google-cloud-storage>=2.4,<3.0",
```

VersusFacit (dismissed):

brabster:
Allowing up to 3.0 is a good improvement, bringing us up to Dec 2024, so I'm happy for that to be merged and close this issue off.

I'm curious how projects tackle this though - can I ask when and how you will push through to the next major version? If the test suite is passes on 3.1.0 (latest), why not push on through? It seems like fixing the upper limit potentially stacks up incompatible changes, making any future discovery of an urgent reason to update in future more risky?

mikealfare:
> Allowing up to 3.0 is a good improvement, bringing us up to Dec 2024, so I'm happy for that to be merged and close this issue off.

It's odd that this wasn't already possible since `~=2.4` should be equivalent to `>=2.4,<3.0`. I know there's an issue with `pip` where it cannot resolve dependencies when these two styles are mixed on the same package. I don't know why that happens, but we've been slowly updating the pins to use the latter (`>=2.4,<3.0`) since it is more flexible.

> I'm curious how projects tackle this though - can I ask when and how you will push through to the next major version? If the test suite is passes on 3.1.0 (latest), why not push on through? It seems like fixing the upper limit potentially stacks up incompatible changes, making any future discovery of an urgent reason to update in future more risky?
>
> Removing the upper bound would seem to provide a useful early warning of any actual incompatibilities while the impact they pose is minimised?

We should be addressing this regularly for the reasons you stated. We generally do this via `dependabot` PRs ([example](https://github.com/dbt-labs/dbt-adapters/pull/936)). To be honest, we simply don't always get to it because there are more pressing issues and we choose stability instead. We generally allow the minor to move unrestricted because this theoretically does not introduce breaking changes. That's not guaranteed with a major bump. If we publish a package to PyPI that leaves the upper limit open, a major bump in one of the dependencies could then break that package. We would then need to publish a patch capping the dependency. Capping it ahead of time provides more control for us to opt-in to these updates instead of opt-out. Even if we do allow the range to cross major versions, we would likely still cap at the most recently tested minor version.

One distinction I'll call out is that we treat adapters as "applications" versus "libraries". I'll use those terms loosely since there's generally some consternation when bringing up this debate. The key point though is that an "application" is more of an end state package whereas a "library" is meant to be reused. Pinning dependencies in a library makes it much harder to use that library downstream as it limits functionality either directly or indirectly through transitive dependencies, as you're calling out. Additionally, the risk posed by a dependency experiencing a major bump can be addressed by the downstream package. However, with an "application" that would be left to the end-user, which is not a desired state.

All this to say, if you want to bump this to an upper bound of the most recent version (`<3.2`), I'd be comfortable merging that if tests pass.

brabster:
Hey thanks, I've bumped the upper limit as you suggested. Thanks so much for taking the time to explain all that so well! Really useful. I'm usually working on the application side of things, I guess the big difference is for me if there's going to be a problem it'll be in the build I do, so I can use a build failure to protect against issues. For a library developer, the problem will probably be in someone else's build first.

Let me know if I need to do anything else before you can merge!

mikealfare (approved):

MartinNowak:
google-cloud-storage is incompatible with urllib3>=2.6.0 and requires at least `google-cloud-storage>=3.7.0` for a fix, could this be relaxed further please, maybe `<4` to only cap at the next major version?.

https://github.com/googleapis/python-storage/pull/1658

PR #932, opened 2025-03-21
ADAP-1170: Migrate managed catalog to the new catalog framework

mikealfare:
We need to migrate the current Iceberg implementation over to the new catalog framework. The current implementation works off of `table_format` being defined in the model config. So if that's there and `catalog_name` is not defined, we need to raise a warning to the user to use the new config (`catalog_name = 'snowflake'`). In this scenario, we should fallback to the existing DDL. Otherwise, if `catalog_name` is defined, we should use the new DDL. If neither of these are defined, this is just a standard table.


mikealfare:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-adapters/932?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#932** <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-adapters/932?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-adapters/932?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#938** <a href="https://app.graphite.dev/github/pr/dbt-labs/dbt-adapters/938?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -65,6 +65,7 @@ class SnowflakeAdapter(SQLAdapter):
 
     AdapterSpecificConfigs = SnowflakeConfig
 
+    CATALOG_INTEGRATIONS = CATALOG_INTEGRATIONS
```
colin-rogers-dbt, line 68: why reassign the constant?

dbt-snowflake/src/dbt/adapters/snowflake/relation.py
```diff
@@ -151,75 +142,13 @@ def can_be_renamed(self) -> bool:
         """
         return self.type in self.renameable_relations and not self.is_iceberg_format
 
-    def get_ddl_prefix_for_create(self, config: RelationConfig, temporary: bool) -> str:
```
colin-rogers-dbt, line 154: if we remove this won't we break anyone who has custom macros that depend on this? We may want to leave it for backwards compatibility

dbt-snowflake/src/dbt/adapters/snowflake/parse_model.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def automatic_clustering(model: RelationConfig) -> Optional[bool]:
+    if not model.config:
+        return None
+
+    return model.config.get("automatic_clustering", False)
+
+
+def base_location(model: RelationConfig) -> Optional[str]:
```
colin-rogers-dbt, line 16: this function is probably worth unit testing on it's own as it's now a lot easier to test

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/create.sql
```diff
@@ -1,85 +1,14 @@
...
+        {{ snowflake__create_dynamic_table_iceberg_managed_sql(dynamic_table, relation, compiled_code) }}
     {%- else -%}
-        {{ _get_create_dynamic_standard_table_as_sql(dynamic_table, relation, sql) }}
+        {% do exceptions.raise_compiler_error('Unexpected model config for: ' ~ relation) %}
     {%- endif -%}
 
 {%- endmacro %}
-
-
-{% macro _get_create_dynamic_standard_table_as_sql(dynamic_table, relation, sql) -%}
```
colin-rogers-dbt, line 27: nit: I think moving these to separate files may actually make it harder to reason about this code as you'll have to swap between three files to reason about diffs / make updates. Down to try it this way for now but worth a thought

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/create.sql
```diff
@@ -1,85 +1,14 @@
...
-        {{ optional('refresh_mode', dynamic_table.refresh_mode) }}
-        {{ optional('initialize', dynamic_table.initialize) }}
-        as (
-            {{ sql }}
-        )
-
-{%- endmacro %}
-
-
-{% macro _get_create_dynamic_iceberg_table_as_sql(dynamic_table, relation, sql) -%}
```
colin-rogers-dbt, line 56: Should we keep this macro name as is? Can we extend the templating later to handle non-managed catalogs as needed? Just thinking that the diff here is really just `catalog = 'SNOWFLAKE'`

colin-rogers-dbt (commented):
Few thoughts, will take another pass later

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_iceberg_managed.py
```diff
@@ -0,0 +1,46 @@
+from dataclasses import dataclass
+from typing import Optional
+
+from dbt.adapters.catalogs import (
+    CatalogIntegration,
+    CatalogIntegrationConfig,
+)
+from dbt.adapters.contracts.relation import RelationConfig
+
+from dbt.adapters.snowflake import constants, parse_model
```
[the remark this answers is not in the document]
mikealfare: I only want one `constants` module for the adapter. And parsing fields out of the model config is not specific to catalogs.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
mikealfare: I'm assigning the value `dbt.adapters.snowflake.catalogs.CATALOG_INTEGRATIONS` to the class variable `SnowflakeAdapter.CATALOG_INTEGRATIONS`. They just have the same name. I can update the name if it's confusing.

dbt-snowflake/src/dbt/adapters/snowflake/relation.py
mikealfare: Oh, I suppose.

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/create.sql
mikealfare: I can push them back together if desired. I was trying to follow the "one macro per file" thing that we vaguely discussed in the past. But we can move that up one layer to something like command (create table...) and we wouldn't really lose that much discoverability.

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/create.sql
mikealfare: I'd like to follow a naming convention across relation types. I'm open to what that convention is, but I'd like to account for the different DDL options across different Iceberg catalog integrations with different templates, to make it clear what is and isn't supported, and avoid trying to set a setting that does not apply to a specific catalog integration.

colin-rogers-dbt: maybe the relation/format level macro like  `_get_create_dynamic_iceberg_table_as_sql` encapsulates the switching logic to call the correct catalog templating macro?

colin-rogers-dbt (approved):

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_native.py
```diff
@@ -0,0 +1,44 @@
...
+
+from dbt.adapters.snowflake import constants, parse_model
+
+
+@dataclass
+class NativeCatalogRelation:
+    catalog_type: str = constants.DEFAULT_NATIVE_CATALOG.catalog_type
+    catalog_name: Optional[str] = constants.DEFAULT_NATIVE_CATALOG.name
+    table_format: Optional[str] = constants.NATIVE_TABLE_FORMAT
+    external_volume: Optional[str] = None
```
VersusFacit, line 18: Shouldn't these be omitted since they are no part of the DDL for Native? Seems to imply these could be something other than none (which as far as I know they can't)

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
VersusFacit: Yes please. This context will be lost after this PR is merged.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -83,6 +84,24 @@ class SnowflakeAdapter(SQLAdapter):
...
 
+    def __init__(self, config, mp_context) -> None:
+        super().__init__(config, mp_context)
+        self.add_catalog_integration(constants.DEFAULT_NATIVE_CATALOG)
+        self.add_catalog_integration(constants.DEFAULT_ICEBERG_CATALOG)
+
+    def add_catalog_integration(
+        self, catalog_integration: CatalogIntegrationConfig
+    ) -> CatalogIntegration:
+        # Snowflake uppercases everything in their metadata tables
```
VersusFacit, line 95: this is not true actually. I have found exceptions where columns are lowercased, such as with the `is_iceberg` column.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -83,6 +84,24 @@ class SnowflakeAdapter(SQLAdapter):
...
+        # don't mutate the object that dbt-core passes in
+        catalog_integration = deepcopy(catalog_integration)
```
VersusFacit, line 97: Why is a deepcopy needed

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -432,15 +455,47 @@ def debug_query(self):
...
-            config
-            and hasattr(config, "_extra")
-            and (relation_format := config._extra.get("table_format"))
-        ):
-            table_format = relation_format
-
-        return {
-            "adapter_type": "snowflake",
-            "table_format": table_format,
+        # `config` is not a RelationConfig!
```
VersusFacit, line 458: what's this mean?

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -432,15 +455,47 @@ def debug_query(self):
...
+        run_info = {
+            "adapter_type": constants.ADAPTER_TYPE,
+            "table_format": None,
         }
+
+        if config and hasattr(config, "_extra"):
+
+            catalog = config._extra.get("catalog")
+
+            if _table_format := config._extra.get("table_format"):  # type:ignore
```
VersusFacit, line 468: this branching is a bit obscure to me. So the idea is take what they give you, then do a series of defaults? I feel like I'd rather have the default case just be to error out than to set the the table_format to Iceberg.

Or if that's not a valid state, that's a find opportunity for an exception

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -432,15 +455,47 @@ def debug_query(self):
...
+
+        Args:
+            model (RelationConfig): `config.model` (not `model`) from the jinja context
+
+        Returns:
+            Any: The constructed relation object generated through the catalog integration and parser
+        """
+        if catalog := parse_model.catalog_name(model):
+            catalog_integration = self.get_catalog_integration(catalog)
+            relation = catalog_integration.build_relation(model)
```
VersusFacit, line 499: could just return this

dbt-snowflake/src/dbt/adapters/snowflake/parse_model.py
```diff
@@ -0,0 +1,98 @@
...
+from dbt.adapters.contracts.relation import RelationConfig
+
+from dbt.adapters.snowflake import constants
+
+
+def automatic_clustering(model: RelationConfig) -> Optional[bool]:
+    if not model.config:
+        return None
+
+    return model.config.get("automatic_clustering", False)
```
VersusFacit, line 13:
```suggestion
    if model.config:
        return model.config.get("automatic_clustering", False)
    else:
        return None
```

dbt-snowflake/src/dbt/adapters/snowflake/parse_model.py
```diff
@@ -0,0 +1,98 @@
...
+
+def catalog_name(model: RelationConfig) -> Optional[str]:
+    if not model.config:
+        return None
+
+    if _catalog := model.config.get("catalog"):
+        # make catalog_name case-insensitive
+        return _catalog.upper()
+
+    _table_format = table_format(model)
```
VersusFacit, line 38: nit: walrus?

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create/standard.sql
```diff
@@ -0,0 +1,55 @@
...
+{%- set contract_config = config.get('contract') -%}
+{%- if contract_config.enforced -%}
+    {{- get_assert_columns_equivalent(compiled_code) -}}
+    {%- set compiled_code = get_select_subquery(compiled_code) -%}
+{%- endif -%}
+
+{%- set sql_header = config.get('sql_header', none) -%}
+{{ sql_header if sql_header is not none }}
+
+create or replace {{ transient }}table {{ relation }}
```
VersusFacit, line 27: nit: WS

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create/iceberg_managed.sql
```diff
@@ -0,0 +1,57 @@
...
+
+{%- set catalog_relation = adapter.build_catalog_relation(config.model) -%}
+
+{%- set copy_grants = config.get('copy_grants', default=false) -%}
+
+{%- set contract_config = config.get('contract') -%}
+{%- if contract_config.enforced -%}
+    {{- get_assert_columns_equivalent(compiled_code) -}}
+    {%- set compiled_code = get_select_subquery(compiled_code) -%}
+{%- endif -%}
```
VersusFacit, line 22: I'm a bit confused probably most about scoping more than anything else. Why is `compiled_code` set inside an `if` here then used below and what happens when it's empty?

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/replace/standard.sql
```diff
@@ -0,0 +1,27 @@
...
+        - SnowflakeRelation - required for relation.render()
+        - str - is already the rendered relation name
+    - sql: str - the code defining the model
+    Returns:
+        A valid DDL statement which will result in a new dynamic standard table.
+-#}
+
+create or replace dynamic table {{ relation }}
+    target_lag = '{{ dynamic_table.target_lag }}'
+    warehouse = {{ dynamic_table.snowflake_warehouse }}
```
VersusFacit, line 20: no quotes?

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/replace/iceberg_managed.sql
```diff
@@ -0,0 +1,32 @@
...
+    - sql: str - the code defining the model
+    Returns:
+        A valid DDL statement which will result in a new dynamic iceberg table.
+-#}
+
+{%- set catalog_relation = adapter.build_catalog_relation(config.model) -%}
+
+create or replace dynamic iceberg table {{ relation }}
+    target_lag = '{{ dynamic_table.target_lag }}'
+    warehouse = {{ dynamic_table.snowflake_warehouse }}
```
VersusFacit, line 22: no quotes?

VersusFacit (commented):
Went over this as best I could with the context I have!

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
mikealfare: I'm going to move the assignment here and pull it from the `catalogs` package. It's sort of misplaced anyway.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
mikealfare: Wild, I could have sworn when I was troubleshooting this all the metadata came through as uppercase. Now it's all showing up as lowercase.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
mikealfare: I don't want to mutate the integration itself, nor any of its attributes. If I understand normal copy correctly, the objects that the initial class points to are still the same objects on the copy, hence mutating those objects indirectly mutates the original object. For example, this works:
```python
from copy import copy
from dataclasses import dataclass


MY_CONFIG = {"name": "Fred"}


@dataclass
class MyObj:
    config: dict
    

my_first_obj = MyObj(config=MY_CONFIG)
my_second_obj = copy(my_first_obj)
my_second_obj.config["name"] = "Flintstone"
assert my_first_obj.config["name"] == "Flintstone"
```

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
mikealfare: We typed the input of this method as a `RelationConfig` in both the base and here. However, while troubleshooting this, I determined that the object that gets passed into this method is not actually a `RelationConfig` instance. I couldn't update the type annotation here without updating it in the base or `mypy` would raise an error. And I felt like that update is outside of the scope of this PR.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
mikealfare: In the current setup, if `table_format` is not set, we return `None`; we don't raise an exception. I wouldn't want to start raising an exception at this point, especially on a method that I believe is used for logging/auditing.

I want to keep this backwards compatible, so if the user explicitly sets a `table_format`, we should go with it. In the previous flow, the remainder is to simply return `None`. However, now we have catalogs with `table_format`s. If we don't have a catalog, then we're creating a normal table, hence the native format. If we have a catalog, then we should look up the table format on that catalog (not what I did here). I'll update this to look up the `table_format` on the catalog integration.

mikealfare: Ah, I just remembered why I cannot do the latter. This is a class method, which means it does not have a catalog integration client where it can look up catalogs by name. If we had created this method as an instance method, then we could do the lookup. I think this is the correct solution, but falls outside of the scope of this PR.

Given this limitation, we can only look at the name of the catalog since that's all `catalog` is. If `catalog` is the name of the native catalog, then it's a native table type. Otherwise, it's some other catalog. We currently only support Iceberg catalogs, making this a relatively safe bet. There is no motivation to create a second native catalog since you really would only do that to set the external volume, which does not apply to a native catalog.

We should update this method to be an instance method and then update the check here to lookup the `table_format` from the provided catalog. We should do this at the same time as updating the type annotation for the `config` argument.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
mikealfare: Agreed. I think this was left from debugging where I wanted to look at the relation before returning it.

dbt-snowflake/src/dbt/adapters/snowflake/parse_model.py
mikealfare: The initial check is really a `mypy` thing. Because `config` is optional, `mypy` raises an error when trying to access it. I started on more complicated methods where I didn't want a bunch of logic inside of an if block; I then copy/pasted that structure to the rest of the methods. So I could see an argument for consistency, but this also works.

dbt-snowflake/src/dbt/adapters/snowflake/relation.py
mikealfare: I'll add this and `get_iceberg_ddl_options` back in, though they may only work as expected when the user explicitly sets the `table_format`. I think I updated the code in a way where that's actually not the case, but I can't guarantee it.

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/replace/iceberg_managed.sql
mikealfare: That's how it was before, this is just moved.

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create/iceberg_managed.sql
mikealfare: `compiled_code` is never empty if we got this far (or it shouldn't be at least). It gets passed in as an argument, and then if this model is contracted, we need to generate a "superquery" from the compiled code, as the subquery, that selects the specific columns from the compiled code. This will fail if the user has updated the model to have a different data schema. `get_select_subquery` updates the compiled code in place to wrap it with that select query.

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create/iceberg_managed.sql
```diff
@@ -0,0 +1,57 @@
...
+{%- set sql_header = config.get('sql_header', none) -%}
+{{ sql_header if sql_header is not none }}
+
+create or replace iceberg table {{ relation }}
+    {%- if contract_config.enforced %}
+    {{ get_table_columns_and_constraints() }}
+    {%- endif %}
+    {{ optional('cluster by', catalog_relation.cluster_by, '(', '') }}
+    {{ optional('external_volume', catalog_relation.external_volume, "'") }}
+    catalog = 'SNOWFLAKE'
```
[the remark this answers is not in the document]
mikealfare: Will do. What's wild is that you can't omit this. It's not like this is the default.

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create/standard.sql
mikealfare: Are you referring to the lack of a space between `{{ transient }}` and `table`? I did this because the resulting sql will be spaced correctly. If this is not a transient table, then this should be:
`create or replace table <my_table>...`
If this is a transient table, then this should be:
`create or replace transient table <my_table>...`
The difference is `transient ` (with the trailing space. You could also do it as the leading space. Either way, this code looks a little wonky so that the generated code is spaced correctly.

If you're referring to something else, then I'm lost.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
VersusFacit: Yep your understanding is right. I was just curious about the extra resources taken for the deepcopy!

VersusFacit (approved):

PR #947, opened 2025-03-26
dbt-athena: Rework iceberg full refresh for high availability

### LINKED ISSUE #458: feature: Add high availability for iceberg in incremental models and full refresh
# Context

When running incremental tables with table_type=iceberg the high availability behaviour is missing.
When an incremental table is build for the first time, having ha behvior is not relevant, but if the incremental table needs to be re-builded it currently works this way:
- target table dropped 
- target table re-created and filled up 

Instead will be good to have:
- target table not touched
- new target table re-created (in an unique location)
- old target table rename
- new target table assume the final name
- old target table (backup) is dropped

alex-antonison:
resolves #458

### Problem

When `dbt-athena` does a full refresh on an Iceberg table, it deletes the table and then re-creates the table. This results in the Iceberg table being unavailable for however long it takes to full refresh the table. For Iceberg tables that take longer to produce, this is very problematic as it causes significant amount of downtime.

### Solution

Instead of dropping the Iceberg at the beginning of the full refresh, instead first check if the Iceberg table already exists and if so, create the new full refresh table as a tmp relation and then upon completion, swap out the full refreshed Iceberg table for the existing Iceberg table. This approach adds in high availability support when doing a full refresh on an Iceberg table.

For this to work, it requires using s3_data_naming set to `schema_table_unique` as in order to prevent the s3 locations being inconsistent, it will create the new tmp_relation within the same s3 path using a UUID to keep them separate. Once the new fully refreshed tmp_relation is completed, can then simply use renaming to change the `tmp_relation` to the `target_relation`.

### Checklist


parsable-alex-antonison:
Realized there is an issue with my `renaming` approach where the underlying s3 location is not being updated. Need to revisit my approach.

alex-antonison:
With the help of @nicor88 I was able to come up with a solution to the s3 paths being inconsistent.

The solution does require that s3_data_naming is `schema_table_unique` because it relies on the table Iceberg s3 data being stored within a unique UUID location within the same s3 location as the existing table. With this approach, I can fully materialize the existing incremental iceberg table, then do a table name switch, and then drop the existing table relation.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -62,6 +63,54 @@
     {%- endif -%}
     {% set build_sql = "select '" ~ query_result ~ "'" -%}
 
+  -- Running in full refresh, support High Availability for Iceberg table type --
+  -- Must use s3_data_naming schema_table_unique in order to support high availability --
+  -- on a full fresh for an incremental iceberg table --
+  {% elif should_full_refresh() and table_type == 'iceberg' and s3_data_naming == 'schema_table_unique' %}
```
nicor88, line 69: you might want to do a similar check to what was done [here](https://github.com/dbt-labs/dbt-adapters/blob/33b110c56fc401f0a476be4da51cbd8fad1e1e84/dbt-athena/src/dbt/include/athena/macros/materializations/models/table/table.sql#L38C9-L38C41)

`('unique' in s3_data_naming and external_location is none)`

pretty much we allow this feature by default for all unique location, and location that are not explicitly set via external_location.

alex-antonison: Makes sense to me.

colin-rogers-dbt:
Overall this looks really solid, if you can add a test to validate I think we're golden

alex-antonison:
Hello @mikealfare @colin-rogers-dbt. I finally got around to adding tests and think this PR is now in a good spot for review.

colin-rogers-dbt (approved):

juliansteger-sc:
@alex-antonison we ran into an issue with a dbt model that existed as a `VIEW` in Glue / Athena, but got migrated to an  iceberg incremental `TABLE` in dbt - in this case a `full-refresh` fails because the following SQL failed:

```
alter table `schema_name`.`table_name` rename to `schema_name`.`table_name__bkp`
```

its certainly an edge case and can be resolved manually, but maybe there's a straightforward fix for this (e.g., don't run the the `rename` query when the existing relation is a view.)

alex-antonison:
I think since this is merged we would need to spin up a new issue highlighting this edge case and create a new PR.

It would be a fairly straightforward PR since you would just check the type.

PR #954, opened 2025-03-27
Fix the python model empty/sample mode fix to not break when the ref macro has been overridden

### LINKED ISSUE #953: [BUG] Python models break when the `ref` macro has been overridden to return a string
## Problem
In #911 (fix for #906) we altered the python.sql template for python models such that `.render()` would be called on refs in order to skip the extra filtering that happens on them in empty and sample mode. This was necessary because the extra filtering was breaking python models. However, people are known to override the `ref` macro, and sometimes in doing so change the return type of `ref` such that instead of returning a `<Adapter>Relation` object, they return a string. The problem with this is that strings do not have `.render()` method. Thus a new error cropped up in that looked something like

```
  Compilation Error in model my_python_model(models/mart/my_python_model.py)
    'str object' has no attribute 'render'
    
    > in macro py_script_postfix (macros/python_model/python.sql)
    > called by model my_python_model (models/mart/my_python_model.py)"
```

## Desired fix
1. We don't break when `ref()` returns a string
2. We still skip rendering the filtered statement in sample/empty mode

QMalcolm:
resolves #953

### Problem

We broke python models for people who had overridden the `ref` macro to return a string.

### Solution

Only call `render()` on the resolved `ref()` if the thing returned from `ref()` isn't a `string`.

### Checklist

### DIFF
--- a/dbt-adapters/src/dbt/include/global_project/macros/python_model/python.sql
+++ b/dbt-adapters/src/dbt/include/global_project/macros/python_model/python.sql
@@ -11,7 +11,17 @@
     {%- set ref_dict = {} -%}
     {%- for _ref in model.refs -%}
         {% set _ref_args = [_ref.get('package'), _ref['name']] if _ref.get('package') else [_ref['name'],] %}
-        {%- set resolved = ref(*_ref_args, v=_ref.get('version')).render() -%}
+        {%- set resolved = ref(*_ref_args, v=_ref.get('version')) -%}
+
+        {#
+            We want to get the string of the returned relation in order to skip sample/empty mode rendering logic.
+            However, people override the default ref macro, and often return a string instead of a relation. Thus,
+            to make sure we dont blow things up, we have ot ensure the resolved relation is not already a string.
+        #}
+        {%- if resolved is not string -%}
+            {%- set resolved = resolved.render() -%}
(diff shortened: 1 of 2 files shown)


dbt-adapters/src/dbt/include/global_project/macros/python_model/python.sql
```diff
@@ -11,7 +11,17 @@
...
         {% set _ref_args = [_ref.get('package'), _ref['name']] if _ref.get('package') else [_ref['name'],] %}
-        {%- set resolved = ref(*_ref_args, v=_ref.get('version')).render() -%}
+        {%- set resolved = ref(*_ref_args, v=_ref.get('version')) -%}
+
+        {#
+            We want to get the string of the returned relation in order to skip sample/empty mode rendering logic.
+            However, people override the default ref macro, and often return a string instead of a relation. Thus,
+            to make sure we dont blow things up, we have ot ensure the resolved relation is not already a string.
+        #}
+        {%- if resolved is not string -%}
```
aranke, line 21: To be more defensive, maybe we should check if the `render` method is defined on the object instead?

aranke (approved):
LGTM!

mikealfare (approved):

PR #962, opened 2025-04-01
Fix generate changelog linting

mikealfare:
As it turns out, `pre-commit run --all-files` only runs on staged files, not all files in the repo as the docs [suggest](https://pre-commit.com/#pre-commit-run).

### DIFF
--- a/.github/workflows/_generate-changelog.yml
+++ b/.github/workflows/_generate-changelog.yml
@@ -144,7 +144,7 @@ jobs:
                     changie batch ${{ needs.version.outputs.base }}
                 fi
                 changie merge
-                git status
+                git add .
             working-directory: ./${{ inputs.package }}
             env:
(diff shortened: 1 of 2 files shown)


.github/workflows/_generate-changelog.yml
```diff
@@ -144,7 +144,7 @@ jobs:
                     changie batch ${{ needs.version.outputs.base }}
                 fi
                 changie merge
-                git status
```
mikealfare, line 147: I verified this is the behavior of `--all-files` by:
- create a changelog entry with `changie`
- add extra newlines at the end of the entry
- do NOT run `git add .`
- run `pre-commit run --all-files`
- notice this passes
- DO run `git add.`
- run `pre-commit run --all-files`
- the file triggers the newline check, fixing the newlines

colin-rogers-dbt (approved):

PR #967, opened 2025-04-02
Update `list_relations_without_caching` macro to paginate using the same query

mikealfare:
- put the list relations macro in its own file so it's easier to find
- update list relations the macro to paginate using the same query so they don't fall out of sync (they did already)
- break out the sql itself into a separate macro that is called by the main macro
- update how we handle the `quoted_identifiers_ignore_case` setting to align with other areas; basically turn it off for this session and then unset it at the end of the list relations call


colin-rogers-dbt (approved):

mikealfare:
### Merge activity

* **Apr 2, 2:38 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dbt-labs/dbt-adapters/967).

PR #976, opened 2025-04-04
Raise an error when notebook execution failed

jialuoo:
resolves #

### Checklist


jialuoo:
@mikealfare As the tests from this PR show: the setups for bigframes probably has some issue, which complains "location 'None' is set to an unknown value." Please help take a look

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
```diff
@@ -352,7 +352,12 @@ def _format_outputs(self, output_list: list) -> str:
                             formatted_output += f"    {inner_key}: {inner_value}\n"
                     else:
                         formatted_output += f"    {value}\n"
+
+                    # Check errors from the output of notebook execution.
+                    if isinstance(value, str) and value.strip().lower() == "error":
```
mikealfare, line 357: I'd rather not parse the output. Does the SDK not return the error? We should be catching errors, not parsing logs. If needed, can wee poll the job result like we do [here](https://github.com/dbt-labs/dbt-adapters/blob/5882b1df1f8f9ddcd0f4f5fcd09001b1948432e9/dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py#L109-L113)?

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
```diff
@@ -384,6 +389,8 @@ def _process_gcs_log(self, gcs_log_uri: str) -> None:
             # Improve the output format for better readability.
             formatted_output = self._format_outputs(outputs)
             _logger.info(f"Colab notebook runtime outputs from GCS: {formatted_output}")
+        except RuntimeError as e:
```
mikealfare, line 392: This should be moved up prior to parsing the logs, and should raise a `DbtRuntimeError` from `e`.

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
jialuoo: Yeah, the SDK doesn't return the error (it only has one state, <JobState.JOB_STATE_UNSPECIFIED: 0>, for all the cases either failure or success). So we have to parse the output to raise error.

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
jialuoo: Yeah, DbtRuntimeError is better here. Fixed.
Not sure if I understand this `moved up prior to parsing the logs`. We need to parse the logs first to get the error message inside before raising the error.

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
jialuoo: Contacted the colab team, we could use GetNotebookExecutionJob RPC to do it (implemented).

mikealfare (commented):
This looks right, but I have one request. Can we add a test that shows that the error gets raise for a malformed model? I want to make sure that we fail if the user provides a malformed model.

mikealfare (approved):

PR #978, opened 2025-04-05
Support custom package in bigframe mode

jialuoo:
resolves #

### Checklist


dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
```diff
@@ -424,3 +431,65 @@ def _submit_bigframes_job(
...
+            version = importlib.metadata.version(normalized_name)
+            return True, version
+        except Exception:
+            # Unable to determine the version.
+            return False, ""
+
+    # Check the installation of individual packages first.
+    packages_to_install = []
+    for package in packages:
+        if "==" in package:
```
mikealfare, line 453: This implicitly suggests that packages will either be unpinned (e.g. `pandas`) or hard pinned (e.g. `pandas==2.2.3`). Have you considered other scenarios? I think it's reasonable that folks will specify lower bound (e.g. `pandas>=2.0`), upper bound (e.g. `pandas<2.2`), and soft (e.g. `pandas~=2.0`) pins as well, in addition to some combination of these (e.g. `pandas>=2.0,<2.2`). In those scenarios, we would not find an installed version and attempt to install the package.

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
```diff
@@ -424,3 +431,65 @@ def _submit_bigframes_job(
         self._process_gcs_log(gcs_log_uri)
 
         return self._ai_platform_client.get_notebook_execution_job(name=res.name)
+
+
+def _install_packages(packages: list[str]) -> None:
```
mikealfare, line 436: We should add a docstring to this function that notes that this code gets run in the notebook, and not where dbt is installed.

mikealfare (commented):
Can we add a test that shows this is successful? We could add packages with different pins, some of which are already installed and some of which are new.

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
jialuoo: Good point. Fixed it.

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
jialuoo: Added.

mikealfare (commented):
Looks good, now we just need some type of test that this works, to make sure we don't regress in the future.

mikealfare (approved):

PR #989, opened 2025-04-10
Fix list_relation changes from #967

mikealfare:
We are using a non-namespaced variable in the list relations pagination loop, which gets reset each time, causing the paginated relations to be the same list every time. We also are raising an error on this which was previously not raised, breaking some users. We converted this error to a warning.


dbt-snowflake/src/dbt/include/snowflake/macros/metadata/list_relations_without_caching.sql
```diff
@@ -12,26 +12,32 @@
...
+            {%- break -%}
         {%- endif -%}
 
-        {%- set show_objects_sql = snowflake__show_objects_sql(schema, max_results_per_iter, watermark) -%}
+        {%- set show_objects_sql = snowflake__show_objects_sql(schema, max_results_per_iter, paginated_state.watermark) -%}
         {%- set paginated_result = run_query(show_objects_sql) -%}
-        {%- do paginated_results.append(paginated_result) -%}
-        {%- set watermark = paginated_result.columns.get('name').values()[-1] -%}
+        {%- do paginated_state.paginated_results.append(paginated_result) -%}
+        {%- set paginated_state.watermark = paginated_result.columns[1].values()[-1] -%}
```
colin-rogers-dbt, line 40: why get by index instead of key?

mikealfare: Honestly, I made this change when troubleshooting because this is how we used to fetch this column. I just never reverted back. I'll fix it.

colin-rogers-dbt (approved):

PR #996, opened 2025-04-11
Forward Unexpected Operations on Record/Replay Wrappers

### LINKED ISSUE #990: Record/Replay: Detect Unwrapped Functions, Don't Fail
The record/replay mechanism relies on the RecordReplayHandle class to have every method and property as is present on the native database connection it wraps. However, we have seen unexpected calls to custom functions on those connections. We would like to gracefully pass those calls through, even if they are not expected and cannot be recorded. In that case we should log problem, but otherwise proceed normally.

peterallenwebb:
resolves #990 

### Problem

Unexpected operations on record/replay wrappers could cause runtime problems.

### Solution

Dynamically forward such operations, but log an event which can be monitored.

### Checklist


colin-rogers-dbt (approved):

colin-rogers-dbt:
@peterallenwebb is this dependent on a change in dbt-common? Looks like integ tests are failing with a missing exception?

PR #1003, opened 2025-04-16
Quick fix of the session in table.sql

jialuoo:
resolves #

### Checklist


jialuoo:
@mikealfare This is ready for review.

dbt-bigquery/src/dbt/include/bigquery/macros/materializations/table.sql
```diff
@@ -145,9 +145,9 @@ import bigframes.pandas as bpd
 bpd.options.bigquery.project = "{{ target.project }}"
 bpd.options.bigquery.location = "{{ target.location }}"
 bpd.options.compute.extra_query_labels["bigframes-dbt-api"] = "python-model-table"
+session = None
```
jialuoo, line 148: Another option is to change the default session value from None to bpd.get_global_session().

mikealfare: I like the `bpd.get_global_session()` option. Are there any permissions implications to that? I wouldn't think so, but it's worth verifying.

jialuoo: The default session is now obtained using session = bpd.get_global_session(). 

While creating this session has no direct permission implications, subsequent operations using the session may require enabling relevant APIs or roles in certain cases (e.g. create remote functions via session). These permissions are only needed for those later actions, not for the initial session creation.

mikealfare (approved):

dbt-bigquery/src/dbt/include/bigquery/macros/materializations/table.sql
mikealfare: Great. That's fine then as the user would be opting in to the behavior if they chose to do something with the session in their model.

PR #1014, opened 2025-04-23
retry_all + retry `could not open relation with OID` query exceptions

### LINKED ISSUE #642: [ADAP-1030] [Regression] Back-to-back hooks with a transaction using CTAS + DML fails in 1.5+
### Is this a regression in a recent version of dbt-redshift?

- [X] I believe this is a regression in dbt-redshift functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

So this is a weird one, I am struggling to really debug the issue and work out _why_ it is happening, let alone how to _fix_ it even with a workaround. I have a macro (used in a post hook) that creates a temp table, and uses this table to delete from then insert into a fixed table (based off a model). We use this in our packages to update our manifest based on the run outcomes, but I have simplified it for this report. 

Prior to version 1.5, this worked, and importantly this worked on multiple threads AND when the hook is called multiple times. In reality the hook shouldn't _need_ to be called multiple times, but it can happen if there is user error or something else going on. Setting this to a single thread does not solve the issue.

What now happens is we get an `could not open relation with OID 3251611` error on the running of the second hook. I have tried making the temp table not temporary, I have tried adding a commit inside the transaction block, outside the transaction block. I have tried dropping the temporary table before the transaction starts, and inside the start of the transaction, and i have tried adding `autocommit: false` to my profile - none of it seems to work.

I have confirmed this works on version 1.4.0, so it must be related to the change in connection method.

### Expected/Previous Behavior

The post hooks would be complete successfully. I'd even be fine with a workaround that allows it to still work, I am out of ideas.

### Steps To Reproduce

1. New project, redshift profile, dbt 1.5 or greater
2. Create a model called `test_table.sql` content:
    ```
    {{
      config(
        materialized = 'table',
        )
    }}
    select 1 as test_col
    ```
3. create macro `transaction_macro.sql` content:
    ```
    {% macro transaction_macro() %}
        {% if execute %}
            begin transaction;
              --temp table
              create temporary table trans_temp_table as (
                  select 1 as test_col
              );
      
              delete from {{target.schema}}.test_table where test_col in (select test_col from trans_temp_table);
              insert into {{target.schema}}.test_table  (select * from trans_temp_table);
              commit;
            end transaction;
    
            drop table trans_temp_table;
        {% endif %}
    {% endmacro %}
    ```
4. Add the macro twice as an on-end hook in the project
    ```
    on-run-end:
      - "{{ transaction_macro() }}"
      - "{{ transaction_macro() }}"
    ```
5. `dbt run --target redshift`. Compare if you run on dbt 1.4.0 that it is successful.

### Relevant log output

```shell
11:34:37  Running with dbt=1.6.6
11:34:38  Registered adapter: redshift=1.6.2
11:34:38  Unable to do partial parsing because profile has changed
11:34:38  Found 1 model, 2 operations, 0 sources, 0 exposures, 0 metrics, 401 macros, 0 groups, 0 semantic models
11:34:38  
11:34:40  Concurrency: 1 threads (target='redshift')
11:34:40  
11:34:40  1 of 1 START sql table model dbt_ryan.test_table ............................... [RUN]
11:34:46  1 of 1 OK created sql table model dbt_ryan.test_table .......................... [SUCCESS in 6.71s]
11:34:47  
11:34:47  Running 2 on-run-end hooks
11:34:47  1 of 2 START hook: dbt_demo.on-run-end.0 ....................................... [RUN]
11:34:59  1 of 2 OK hook: dbt_demo.on-run-end.0 .......................................... [SUCCESS in 12.41s]
11:34:59  2 of 2 START hook: dbt_demo.on-run-end.1 ....................................... [RUN]
11:34:59  Database error while running on-run-end
11:34:59  
11:34:59  Finished running 1 table model, 1 hook in 0 hours 0 minutes and 21.29 seconds (21.29s).
11:34:59  
11:34:59  Completed with 1 error and 0 warnings:
11:34:59  
11:34:59    on-run-end failed, error:
 could not open relation with OID 3251641
11:34:59  
11:34:59  Done. PASS=1 WARN=0 ERROR=1 SKIP=0 TOTAL=2
```

### Environment

```markdown
- OS: MacOS
- Python: 3.9.13
- dbt-core (working version): 1.4.0
- dbt-redshift (working version): 1.4.0
- dbt-core (regression version): 1.6.6
- dbt-redshift (regression version): 1.6.2

(I believe it has been failing since 1.5.0 but have not verified specifically, we have a customer reporting the issue on 1.5.9 at least)
```

### Additional Context

It's maybe related to this https://github.com/dbt-labs/dbt-redshift/issues/521 and possibly also this https://github.com/dbt-labs/dbt-redshift/issues/501 ?

colin-rogers-dbt:
This change introduces a retry loop in the dbt-redshift execute method. This allows us to retry all the statements in a query instead of just what we do today which is retry on each statement. 

While I'd rather not rely on the exception message this particular exception is returned as a `redshift_connector.error.ProgrammingError`, most of the time this exception class denotes queries that are not retryable (i.e. have some deterministic issue with the syntax/structure of the query). Also, this message hasn't changed in years (it's raised by the underlying postgres layer within redshift) so should be safe to rely on.  

resolves #642 

### Checklist


dbt-adapters/src/dbt/adapters/sql/connections.py
```diff
@@ -72,7 +75,7 @@ def add_query(
         auto_begin: bool = True,
         bindings: Optional[Any] = None,
         abridge_sql_log: bool = False,
-        retryable_exceptions: Tuple[Type[Exception], ...] = tuple(),
+        retryable_exceptions: Union[Callable, Tuple[Type[Exception], ...]] = tuple(),
```
mikealfare, line 78: Instead of updating the signature, would you consider creating a default exception handler in the base that gets overridden by the one you have below? Then we would check that after the existing check:
```python
def add_query(...):
    ...
    try:
        cursor.execute(sql, bindings)
    except retryable_exceptions as e:
        if attempt >= retry_limit:
            raise e
    except Exception as e:
        if not self.is_retryable_exception(e) or attempt >= retry_limit:
            raise e

def is_retryable_exception(self, e: Exception) -> bool:
    return False
```
and `dbt-redshift`:
```python

def is_retryable_exception(self, e: Exception) -> bool:
    if not isinstance(e, redshift_connector.Error):
        return False
    return "could not open relation with OID" in str(e)
```

dbt-redshift/tests/functional/adapter/test_exception_handling.py
```diff
@@ -0,0 +1,61 @@
...
+    def models(self):
+        return {
+            "concurrent_table.sql": """
+                {{ config(materialized='table') }}
+
+                select 1 as id, 'test' as name
+            """
+        }
+
+    def test_concurrent_relation_access_error(self, project, adapter):
```
mikealfare, line 19: Some comments here suggest this might be a flaky test. Could we add `@pytest.mark.flaky` to this to run it in the flaky suite, which also reruns automatically? I think it will only rerun once, though that can be configured; but it would further reduce the number of false positives we get.

dbt-adapters/src/dbt/adapters/sql/connections.py
colin-rogers-dbt: On further testing I've realized we're going to need to re-think this whole strategy as we don't need to retry the query we need to retry the whole transaction which for redshift is a series of queries.

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -517,37 +519,67 @@ def execute(
...
+                    return add_query_func(
+                        query,
+                        auto_begin,
+                        bindings=bindings,
+                        abridge_sql_log=abridge_sql_log,
+                        retryable_exceptions=retry_exceptions,
+                        retry_limit=self.profile.credentials.retries,
+                    )
+
+            except redshift_connector.Error as e:
```
mikealfare, line 562: Do we want to catch this exception here, or in `execute`? Doesn't this just queue the queries before (without) executing them?

colin-rogers-dbt: no that would be too obvious, instead `add_query` declares and calls `_execute_query_with_retry` within itself

dbt-redshift/.changes/unreleased/Features-20250424-091102.yaml
VersusFacit, line 1: Is one of these duplicate?

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -134,6 +139,7 @@ class RedshiftCredentials(Credentials):
     role: Optional[str] = None
     sslmode: UserSSLMode = field(default_factory=UserSSLMode.default)
     retries: int = 1
+    retry_all: bool = False
```
VersusFacit, line 142: I'm wondering if retry_all is obscuring what exactly. Should we say `retry_all_errors` so uses know we aren't retrying failures?

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -506,46 +515,101 @@ def execute(
...
+            retries -= 1
+            # we need to actually close and open to get a new connection
+            # otherwise no queries will succeed on this connection
+            self.close(self.get_thread_connection())
+            self.open(self.get_thread_connection())
+            time.sleep(backoff)
+            # return with exponential backoff
+            return retries, min(backoff * 2, 60)
+
+        def _execute_internal(
```
VersusFacit, line 541: I think I might prefer this as _try_execute_loop but maybe that's just too rusty.

dbt-redshift/tests/functional/adapter/test_exception_handling.py
```diff
@@ -0,0 +1,66 @@
...
+    def profiles_config_update(self, dbt_profile_target):
+        outputs = {"default": dbt_profile_target}
+        outputs["default"]["retries"] = 5
+        outputs["default"]["threads"] = 4
+        return outputs
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "test_table.sql": """
```
VersusFacit, line 20: Let make these constants on top of the file.

dbt-redshift/src/dbt/adapters/redshift/connections.py
colin-rogers-dbt: yea I went with `retry_all` as that's what we do for spark and snowflake, I think consistency is worth it here

dbt-redshift/tests/functional/adapter/test_exception_handling.py
```diff
@@ -0,0 +1,66 @@
...
+                select 1 as test_col
+            """
+        }
+
+    @pytest.fixture(scope="class")
+    def macros(self):
+        return {
+            "macro.sql": """
+            {% macro transaction_macro() %}
+                {% if execute %}
```
VersusFacit, line 35: why was this needed? Very curious

dbt-redshift/tests/functional/adapter/test_exception_handling.py
```diff
@@ -0,0 +1,66 @@
...
+            "on-run-end": [
+                "{{ transaction_macro() }}",
+                "{{ transaction_macro() }}",
+            ]
+        }
+
+    def test_retry_on_relation_oid_not_found(self, project):
+        result, stdout = run_dbt_and_capture(["run", "--log-level=debug"])
+        print(stdout)
+        assert "could not open relation with OID" in stdout
```
mikealfare, line 65: Well done

VersusFacit (commented):
Logic is super clear. Only a couple of nits and questions about test. I can't quite follow all the semantics of what you're trying to test so I can't say if it's a legitimate test / your code change fixes it.

dbt-redshift/tests/functional/adapter/test_exception_handling.py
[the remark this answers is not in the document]
colin-rogers-dbt: it does we're inducing concurrent updates which causes redshift to raise the exception

dbt-redshift/src/dbt/adapters/redshift/connections.py
VersusFacit: Fair I'll buy consistency

dbt-redshift/tests/functional/adapter/test_exception_handling.py
colin-rogers-dbt: added some context in the test comment

dbt-redshift/.changes/unreleased/Features-20250424-091102.yaml
VersusFacit: That's fine. Just making sure :D I agree there can be two if needed!

dbt-redshift/tests/functional/adapter/test_exception_handling.py
```diff
@@ -0,0 +1,70 @@
...
+_TEST_MODEL = """
+                {{
+                  config(
+                    materialized = 'table',
+                    )
+                }}
+                select 1 as test_col
+            """
+
+_MACROS = {
```
mikealfare, line 13: Now that we're re-running tests, can we make this format follow the models format? i.e. create the macro itself as the constant and assign it:

```python
_TEST_MACRO = """
{% macro transaction_macro() %}
    ...
{% endmacro %}
```
```python
    @pytest.fixture(scope="class")
    def macros(self):
        return {"macro.sql": _TEST_MACRO}
```

VersusFacit (approved):
Approving with more context and appreciate the nits being undertaken!

PR #1020, opened 2025-04-25
Complete fix.

### LINKED ISSUE #963: [Bug] Iceberg tables throwing ambiguous column error.
### Is this a new bug?

- [x] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [x] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

Right now, dbt does an extra query to verify if a table is or is not an iceberg table for snowflake. This query is generating an error saying column IS_ICEBERG is ambiguous, what is not expected

### Expected Behavior

DBT can run the query without issues and validate if a table is iceberg or not.

### Steps To Reproduce

1. Run dbt with flag on
2. Error is supposed to be throw

### Relevant log output

```shell

```

### Environment

```markdown
- OS:
- Python:
- dbt-adapters:
- <adapter>:
```

### Additional Context

I think this issue is a really easy to fix, just remove this part of the code:
```
    {# -- Gated for performance reason. If you don't want Iceberg, you shouldn't pay the
       -- latency penalty. #}
    {% if adapter.behavior.enable_iceberg_materializations.no_warn %}
      select all_objects.*, is_iceberg
      from table(result_scan(last_query_id(-1))) all_objects
      left join INFORMATION_SCHEMA.tables as all_tables
        on all_tables.table_name = all_objects."name"
        and all_tables.table_schema = all_objects."schema_name"
        and all_tables.table_catalog = all_objects."database_name"
    {% endif -%}
```

from the macro `snowflake__list_relations_without_caching`

VersusFacit:
resolves #963 

### Problem

`is_iceberg` has been added to `SHOW OBJECTS`. This has led to some weird behavior such as double `is_iceberg` columns throwing warnings in logs. But more importantly an incoming bundle Monday will start breaking snowflake projects that have the `enable_iceberg_materializations` flag turned on today.

The changes in this PR essentially repurpose this flag `enable_iceberg_materializations` to instead control when to construct `is_iceberg` or not. I didn't want to change the user project interface for now since we're going to remove the flag entirely soon enough.

### Solution

Users who wait until Monday (or later by exception) get a latency penalty but that should be excessively few since Snowflake requires users upgrade with no way to change that. This is a complex in interface yet straightforward in code approach to deal with the possible mix of metadata queries in the dbt-snowflake userspace. In time all these hacks and workarounds and all traces of the behavior flag will be removed for good.

### Checklist


dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -273,37 +278,69 @@ def get_catalog_for_single_relation(
...
+
+        try:
+            _, objs = self.execute(f"show objects in {schema} limit 1", fetch=True)
+        except DbtDatabaseError as exc:
+            if "002043 (02000)" in str(exc):
+                return  # a la 'rollback transaction'
+
+        # Abuse of the behavior flag interface to overwrite it / change its semantics
+        # to a flag like: is there is_iceberg or must we reconstruct it?
+        # Don't do this. My hand was forced because of need of a schema.
```
VersusFacit, line 308: This is solution no. 4. I did not reach for this at first and actively tried to avoid it. The placement of this function makes it quite hard to efficiently determine behavior dependent on metadata schema during the runtime

colin-rogers-dbt (approved):

colin-rogers-dbt (changes requested):
Unsure we actually want to do this change in light of last week's investigations

VersusFacit:
We did not need this to survive bundle-pocalypse Aor 2025. 

We'll be removing the behavior flag entirely soon

PR #1025, opened 2025-04-25
Remove enable_iceberg_materializations behavior flag now that 2025_01 is required

mikealfare:
Now that the Snowflake bundle `2025_01` is required, we no longer need to gate `list_relations` on `enable_iceberg_materializations`. We can rely on `is_iceberg` always being present in `show objects`.


dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create.sql
```diff
@@ -122,10 +122,6 @@ alter table {{ relation }} resume recluster;
     - Iceberg does not support temporary tables (use a standard Snowflake table)
 -#}
 
-{%- if not adapter.behavior.enable_iceberg_materializations.no_warn -%}
-    {%- do exceptions.raise_compiler_error('Was unable to create model as Iceberg Table Format. Please set the `enable_iceberg_materializations` behavior flag to True in your dbt_project.yml. For more information, go to https://docs.getdbt.com/reference/resource-configs/snowflake-configs#iceberg-table-format') -%}
```
VersusFacit, line 126: as a followup item we should remove this docs item

VersusFacit (approved):

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create.sql
mikealfare: Fair. I'll roll this into the general Iceberg catalog docs updates since all of that will change with 1.10.

PR #1049, opened 2025-05-01
Keep limit in get_test_sql

aranke:
resolves failing dbt-core runs: https://github.com/dbt-labs/dbt-core/actions/runs/14774226697/job/41479499471

### Problem

dbt-core tests were failing

### Solution

Add limit back to `get_test_sql` that was removed in #376.

### Checklist

### DIFF
--- a/dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql
+++ b/dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql
@@ -56,7 +56,7 @@
 
   {% call statement('main', fetch_result=True) -%}
 
-    {{ get_test_sql(main_sql, fail_calc, warn_if, error_if)}}
+    {{ get_test_sql(main_sql, fail_calc, warn_if, error_if, limit)}}
 
   {%- endcall %}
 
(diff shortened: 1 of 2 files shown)


dbt-adapters/src/dbt/include/global_project/macros/materializations/tests/test.sql
```diff
@@ -56,7 +56,7 @@
 
   {% call statement('main', fetch_result=True) -%}
 
-    {{ get_test_sql(main_sql, fail_calc, warn_if, error_if)}}
+    {{ get_test_sql(main_sql, fail_calc, warn_if, error_if, limit)}}
```
dbeatty10, line 59: https://github.com/dbt-labs/dbt-adapters/pull/376 explicitly stopped passing in the limit parameter because we do **not** want to apply the limit twice when failures are stored (just once!).

So I think we'd want something like this instead:

```suggestion
    {{ get_test_sql(main_sql, fail_calc, warn_if, error_if, limit=none)}}
```

dbeatty10:
Closing in favor of https://github.com/dbt-labs/dbt-adapters/pull/1052

PR #1053, opened 2025-05-01
Align on `pytest` version of `7.x`

mikealfare:
We inadvertently updated which version of `pytest` we use in our tests in #1016. This aligns all adapters to use `pytest~=7.0`.


VersusFacit (approved):
Smashing

PR #1054, opened 2025-05-01
Always run `dbt-spark` integration tests on `py3.10` 

mikealfare:
`dagger-io` does not support `py3.9`, which is our current default python version. Once `py3.9` hits end of life later this year, we can update our default version and revert this patch.

We were previously running these tests on `py3.12` unknowingly. See #1016 to see how we addressed this (setting `HATCH_PYTHON`).

### DIFF
--- a/.github/workflows/_integration-tests.yml
+++ b/.github/workflows/_integration-tests.yml
@@ -378,7 +378,9 @@ jobs:
         -   uses: actions/setup-python@v5
             with:
                 python-version: ${{ inputs.python-version }}
-        -   run: echo "HATCH_PYTHON=${{ inputs.python-version }}" >> $GITHUB_ENV
+        # TODO: update 3.10 to ${{ inputs.python-version }} once we update default python to >=3.10
+        # the version of dagger-io that we use does not support py3.9
+        -   run: echo "HATCH_PYTHON=3.10" >> $GITHUB_ENV
         -   uses: pypa/hatch@install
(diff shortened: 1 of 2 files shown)


mikealfare:
Workflow changes do not run tests. [Here](https://github.com/dbt-labs/dbt-adapters/actions/runs/14783964415) is a run on this branch for `dbt-spark`. It was run with an input of Python 3.9 (which is what failed), but should run on py3.10 and succeed because of this PR.

.github/workflows/_integration-tests.yml
```diff
@@ -378,7 +378,8 @@ jobs:
         -   uses: actions/setup-python@v5
             with:
                 python-version: ${{ inputs.python-version }}
-        -   run: echo "HATCH_PYTHON=${{ inputs.python-version }}" >> $GITHUB_ENV
+        # TODO: update 3.10 to ${{ inputs.python-version }} once we update default python to >=3.10
+        # the version of dagger-io that we use does not support py3.9
+        -   run: echo "HATCH_PYTHON=3.10" >> $GITHUB_ENV
         -   uses: pypa/hatch@install
-        -   run: hatch run pip install -r dagger/requirements.txt
```
mikealfare, line 383: This was moved into the integration test call to make sure it's installed in the correct environment. That could also be done here, but it felt more appropriate to be in the script instead of here. Also, the default environment installs it via `hatch run setup`, so now it's all in the same file.

colin-rogers-dbt (approved):

PR #1068, opened 2025-05-06
Fix issue with check column snapshots in new_record mode.

peterallenwebb:
Resolves https://github.com/dbt-labs/dbt-core/issues/11269

### Problem

When the new_record mode is used in combination with check columns in snapshots, it was possible for a column to be re-inserted into the source table after a deletion, without being updated in the target table on the next run of `dbt snapshot`. Specifically, this would happen if the checked columns of the re-added row matched the deleted row exactly.

### Solution

Row change logic was modified so that a re-added row identical to a deleted row is not ignored.

### Checklist


dbt-adapters/src/dbt/include/global_project/macros/materializations/snapshots/helpers.sql
```diff
@@ -109,7 +109,9 @@
         left outer join snapshotted_data
             on {{ unique_key_join_on(strategy.unique_key, "snapshotted_data", "source_data") }}
             where {{ unique_key_is_null(strategy.unique_key, "snapshotted_data") }}
-            or ({{ unique_key_is_not_null(strategy.unique_key, "snapshotted_data") }} and ({{ strategy.row_changed }})
+            or ({{ unique_key_is_not_null(strategy.unique_key, "snapshotted_data") }} and (
+               {{ strategy.row_changed }} {%- if strategy.hard_deletes == 'new_record' -%} or snapshotted_data.{{ columns.dbt_is_deleted }} = 'True' {% endif %}
```
colin-rogers-dbt, line 113: `snapshotted_data.{{ columns.dbt_is_deleted }}` is a varchar column or boolean? Should this be True without quotes?

peterallenwebb: @colin-rogers-dbt It is a varchar, so it is correct as written.

colin-rogers-dbt (approved):

PR #1071, opened 2025-05-07
[Bug] Apply cluster key to tables after table creation during a ctas statement

mikealfare:
### Problem

Some users are seeing issues when creating table materializations with a cluster by key. Per Snowflake docs, this should be done afterwards via an alter statement.

### Solution

Remove the cluster by clause from the CTAS statement and run it as a separate alter statement afterwards.

### Checklist

### DIFF
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create.sql
@@ -87,7 +87,6 @@ create or replace {{ transient }}table {{ relation }}
     {%- if contract_config.enforced %}
     {{ get_table_columns_and_constraints() }}
     {%- endif %}
-    {{ optional('cluster by', catalog_relation.cluster_by, '(', '') }}
     {% if copy_grants -%} copy grants {%- endif %}
     as (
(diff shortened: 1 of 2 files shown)


colin-rogers-dbt (approved):

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create.sql
```diff
@@ -87,7 +87,6 @@ create or replace {{ transient }}table {{ relation }}
     {%- if contract_config.enforced %}
     {{ get_table_columns_and_constraints() }}
     {%- endif %}
-    {{ optional('cluster by', catalog_relation.cluster_by, '(', '') }}
```
mikealfare, line 90: We were also applying the cluster key after this create statement via an alter statement, so we only need to remove this clause.

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create.sql
```diff
@@ -144,7 +143,6 @@ create or replace iceberg table {{ relation }}
...
```
mikealfare, line 147: We were also applying the cluster key after this create statement via an alter statement, so we only need to remove this clause.

PR #1075, opened 2025-05-08
Add unit test for relation with oid exception handling and skip functional test

colin-rogers-dbt:
Replace functional test for relation with oid exception handling with a unit test.

This is a tricky exception to reliably induce, in this case it seems like a recent update to the redshift-connector changed the query execution just enough to render the test scenario invalid. Leaving it in as I'd like to find a reliable replacement and unskip this test at some point

resolves #

### Checklist

### DIFF
--- a/dbt-redshift/tests/functional/adapter/test_exception_handling.py
+++ b/dbt-redshift/tests/functional/adapter/test_exception_handling.py
@@ -32,6 +32,7 @@ _MACROS = {
 
                       delete from {{target.schema}}.test_table where test_col in (select test_col from trans_temp_table);
                       insert into {{target.schema}}.test_table  (select * from trans_temp_table);
+                      insert into {{target.schema}}.test_table  (select * from trans_temp_table);
                       commit;
                     end transaction;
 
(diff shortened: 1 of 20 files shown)


peterallenwebb (approved):

dbt-redshift/tests/functional/adapter/test_exception_handling.py
```diff
@@ -32,6 +32,7 @@
 
                       delete from {{target.schema}}.test_table where test_col in (select test_col from trans_temp_table);
                       insert into {{target.schema}}.test_table  (select * from trans_temp_table);
+                      insert into {{target.schema}}.test_table  (select * from trans_temp_table);
```
mikealfare, line 35: I think this only gets used in the test that we're now skipping. Was this intentionally updated or is it an artifact of troubleshooting?

mikealfare (approved):
Just the one nit where it looks like a copy/paste error, but nothing blocking.

dbt-redshift/tests/functional/adapter/test_exception_handling.py
colin-rogers-dbt: artifact of trouble shooting.... will delete

PR #1091, opened 2025-05-13
Add query comments to system queries

### LINKED ISSUE #1090: [ADAP-669] [CT-2779] [Bug] Incremental model merge/drop queries missing the dbt app comment to identify as dbt query

### Current Behavior

The merge query and the drop table query that are sent to BigQuery for incremental models that use the insert/overwrite+dynamic strategy are missing the comment on the first row, which includes a JSON string to identify the query as being issued from dbt along with the dbt version, which looks like this: "/* {"app": "dbt", "dbt_version": "1.5.1", "profile_name": ...............*/"

### Expected Behavior

The merge query and the drop table query that are sent to BigQuery for incremental models that use the insert/overwrite+dynamic strategy include a comment on the first row, which includes a JSON string to identify the query as being issued from dbt along with the dbt version, which looks like this: "/* {"app": "dbt", "dbt_version": "1.5.1", "profile_name": ...............*/"

### Steps To Reproduce

Use the below config to setup a incremental model using the insert/overwrite dynamic strategy, the perform 2 dbt runs, the first one will do a full refresh and the second will do an incremental run, which includes 3 queries: create tmp table, merge using tmp table, drop tmp table. The last 2 steps of merge and drop will not have the dbt app comment on the first line of the query, or anywhere in the query body.

### Relevant log output

_No response_

### Environment

```markdown
- OS: Win 11
- Python: 3.11.4
- dbt: 1.5.1
```

bigquery

### Additional Context

_No response_

VersusFacit:
resolves #1090 

## Problem
Wowee, wowee, this one took me a fair bit of effort to find a case to reproduce it. But we stuck the landing 🏂 

* The BigQuery adapter still had a few helper paths (`get_column_schema_from_query`,  
  `get_columns_in_select_sql`, and the `copy_partitions` flow they trigger) that called
  `raw_execute()` directly.  
* `raw_execute()` bypasses dbt’s query-comment injection, so those statements
  (e.g. the final `COPY … TO …` job produced when `copy_partitions=True`) reached
  BigQuery without the structured  
  `/* {"app": "dbt", "dbt_version": …, "node_id": …} */` header.  
* Existing tests only looked at statements beginning with `CREATE | MERGE | DROP`,
  so the leak went undetected.

## Solution
* Introduced **`raw_execute_with_comment()`**, a thin wrapper that prepends
  `add_query_comment()` to every SQL string before delegating to `raw_execute()`.
* Re-routed all helper calls that previously used `raw_execute()` to use the new
  wrapper instead:
* Added a regression test that used the edge case of `copy_partitions=True` to test this behavior
* I used TDD to verify this test is actually functioning

Result: **every query the BigQuery adapter sends—data-load scripts *and*
metadata helpers—now carries the structured dbt query comment.**

### Checklist

### DIFF
--- a/dbt-bigquery/src/dbt/adapters/bigquery/connections.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/connections.py
@@ -274,6 +274,28 @@ class BigQueryConnectionManager(BaseConnectionManager):
                 limit=limit,
             )
 
+    def raw_execute_with_comment(
+        self,
+        sql: str,
+        use_legacy_sql: bool = False,
+        limit: Optional[int] = None,
+        dry_run: bool = False,
+    ):
+        """
+        A lightweight wrapper over raw_execute that prepends the dbt query comment.
+
+        This exists as a "third way" between raw_execute (fully manual, no preprocessing)
+        and execute (postprocessing and formatting). This is useful when you need query
+        auditing but no Adapter Response.
+        """
+        sql = self._add_query_comment(sql)
+        return self.raw_execute(
+            sql,
+            use_legacy_sql=use_legacy_sql,
+            limit=limit,
+            dry_run=dry_run,
+        )
+
     def execute(
         self, sql, auto_begin=False, fetch=None, limit: Optional[int] = None
     ) -> Tuple[BigQueryAdapterResponse, "agate.Table"]:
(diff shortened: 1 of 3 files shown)


github-christophe-oudar:
That looks good! It's cleaner than my rougher approach with `execute`! Thank you for spending time on that long lasting issue! 🙌

colin-rogers-dbt (approved):

PR #1092, opened 2025-05-13
update error check to account for both types of logging

mikealfare:
The logging is slightly different depending on whether it's JSON format or text format. The original test may pass locally because that's text(?) format whereas CI uses JSON (or the other way around).


emmyoop (approved):
This makes sense.

colin-rogers-dbt (approved):

PR #1101, opened 2025-05-15
Fixed Casing Issue In Snapshot Target Validation

peterallenwebb:
Resolves: https://github.com/dbt-labs/dbt-core/issues/11385

### Problem

Snapshots would fail, claiming that columns in the target table did not exist, when in fact they did.

### Solution

Remove case sensitivity in the column name checks.

### Checklist


dbt-adapters/src/dbt/adapters/base/impl.py
```diff
@@ -872,6 +872,7 @@ def valid_snapshot_target(
         # aren't always present.
         for column in ("dbt_scd_id", "dbt_valid_from", "dbt_valid_to"):
             desired = column_names[column] if column_names else column
+            desired = desired.lower() if desired is not None else desired
```
mikealfare, line 875: Do we specifically want to change the case of the column name that we're appending? Or do we just want to do a case-insensitive search to know whether to append it? For example:
```python
for column in ("dbt_scd_id", "dbt_valid_from", "dbt_valid_to"):
    desired = column_names[column] if column_names else column
    if desired and desired.lower() not in names:
        missing.append(desired)
```

peterallenwebb: Agreed. Done!

mikealfare (approved):

dbt-adapters/src/dbt/adapters/base/impl.py
```diff
@@ -872,7 +872,7 @@ def valid_snapshot_target(
         # aren't always present.
         for column in ("dbt_scd_id", "dbt_valid_from", "dbt_valid_to"):
             desired = column_names[column] if column_names else column
-            if desired not in names:
+            if desired and desired.lower() not in names:
```
colin-rogers-dbt, line 875: we might also need to strip out escaped quotes

colin-rogers-dbt (approved):

dbt-tests-adapter/src/dbt/tests/adapter/simple_snapshot/fixtures.py
```diff
@@ -100,7 +100,7 @@
     ip_address,
     updated_at,
     test_valid_from,
-    test_valid_to,
+    TEST_VAILD_TO,
```
ericj-db, line 103: @peterallenwebb same here

PR #1105, opened 2025-05-20
Add iceberg catalog support for dbt-bigquery

mikealfare:
This adds base functionality for using Iceberg tables in `dbt-bigquery`. This is done by supplying `catalog='iceberg_managed'` and a `storage_uri` in the model. It will use the default connection. In the future, we may elect to build `storage_uri` from a catalog-level root and the model name and to configure the connection to be used. This feature requires additional permissions. In particular it requires the role `BigQuery Connection Admin` to create the Iceberg table ([docs](https://cloud.google.com/bigquery/docs/iceberg-tables#required-roles)).

Limitations:
- Users must use the default connection
- Users must supply the full storage URI on each model
- Users cannot recreate the table without updating the storage URI since that location will be in use already
- Users must first drop an Iceberg table (standard or incremental) if they want to replace it with a materialized view


mikealfare:
The new tests fail because they require a new GH env var for the storage uri. This new env var is added to the workflow in this PR, but the PR checks use the workflows on `main`. Here's a run of the integration tests against this branch: https://github.com/dbt-labs/dbt-adapters/actions/runs/15162434809/job/42631537403

mikealfare:
The tests on this PR will fail because the workflows need a new env var that the workflow on `main` does not yet have. Here's a run on this branch:
https://github.com/dbt-labs/dbt-adapters/actions/runs/15176020200

PR #1108, opened 2025-05-20
Fix dev and ci to install all requirements during pre-installation to ensure transients dependencies are also installed from main

mikealfare:
We bumped `dbt-common` to 2.0 and then updated the pin in dependent packages to the next major version. However, we have not published that pin to PyPI yet for `dbt-core`. This breaks when trying to install a package where `dbt-core` is a transient dependency (`dbt-tests-adapter`) and when also installing `dbt-common` (all packages). We should ensure that we're installing all dbt- packages from `main` for dev and ci; this issue made it evident that we were not. While we later install these dependencies as part of the normal dependencies from main, different transient dependencies may have been installed when using PyPI in the pre-installation.

### DIFF
--- a/dbt-athena-community/hatch.toml
+++ b/dbt-athena-community/hatch.toml
@@ -4,6 +4,8 @@ sources = ["src"]
 
 [envs.default]
 pre-install-commands = [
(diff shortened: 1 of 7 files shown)


mikealfare:
We decided against this approach

PR #1112, opened 2025-05-21
quote columns added on incremental runs BigQuery

### LINKED ISSUE #157: [CT-3130] [Bug] Unquoted columns in macro are causes BigQuery SQL exception

### Current Behavior

The macro: `default__alter_relation_add_remove_columns` generated by incremental config: `on_schema_change = 'append_new_columns'` is causing our build to fail when it encounters columns that are named as reserved keywords [BigQuery SQL syntax](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#reserved_keywords).
The macro generates an alter statement which does not quote the new columns.

When we try to add/remove a column at some points in this macro, we hit a BQ SQL error due to a column named after a reserved keyword name.

### Expected Behavior

We expect all column name references, within the macro to be wrapped in quotes via the `adapter.quote()` utility.
This will allow us to ALTER tables without failing the entire DBT build.

### Steps To Reproduce

1. Create an incremental `dbt-bigquery` model with the following configuration:
```j2
{{
    config(
        materialized = 'incremental',
        on_schema_change = 'append_new_columns',
        incremental_strategy = 'insert_overwrite',
    )
}}

```

2. Define some columns that are **not reserved keywords**
3. Build the model
4. Add a column named as a [reserved keyword](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#reserved_keywords). For example: `limit`.
5. Rebuild the model (incrementally / not a full-refresh).
6.  Bigquery should raise a syntax exception pointing at the column named as a reserved keyword.

### Relevant log output

```shell
Bigquery SQL stderr:
`Syntax error: Unexpected keyword LIMIT`
```

### Environment

```markdown
- OS:  OSX / DBT Cloud container?
- Python: 3.10.12
- dbt: 1.6.1
```

bigquery

### Additional Context

_No response_

Thrasi:
resolves #157

### Problem

The problem description is quite good in issue #157
On incremental runs with the config: `on_schema_change = 'append_new_columns'`, adding new columns is done using the 
default__alter_relation_add_remove_columns for BigQuery.
If the column name is a reserved keyword the sql fails because the column isn't quoted.

### Solution

Add a BigQuery specific macro where the columns are quoted.
### Checklist


colin-k-rogers:
This was definitely the right track but ended up addressing this in the base dbt-adapter

PR #1119, opened 2025-05-22
fixed issue with double quotes hardcoded for column

### LINKED ISSUE #1118: [Bug] Incorrect quoting when adding column to snapshot
### Is this a new bug?

- [x] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

If running `dbt snapshot` results in needing to add a new column, it erroneously uses a hard coded double quote for quoting the new column name, rather than the adapter-specific quoting method. I'm using mariadb, which needs backticks instead.

### Expected Behavior

It should use the adapter-specific quoting method, ie, `adapter.quote`

### Steps To Reproduce

1. make a snapshot of a table
2. add a new column to the table
3. run snapshot again
4. if your database doesn't use double quote for column names, you'll get an error

### Relevant log output

```shell

```

### Environment

```markdown
- OS: ubuntu 20.04
- Python: 3.9.22
- dbt-adapters: 1.14.8
- <adapter>: custom mariadb
```

### Additional Context

I'm working on putting in a PR. The relevant code is the default__create_columns macro in dbt-adapters/src/dbt/include/global_project/macros/materializations/snapshots/helpers.sql

stevetracvc:
resolves #1118

### Problem

Currently, if you create a snapshot, add a new column to the source table, then update the snapshot, it uses a hard coded double quote to quote the new column name.

### Solution

It should instead rely on the adapter-specific quoting method.

### Checklist

Not gonna lie, I don't use hatch and I really don't feel like figuring all that crap out just to do a one line PR. If someone else wants to take it over for me, go for it.

### DIFF
--- a/dbt-adapters/src/dbt/include/global_project/macros/materializations/snapshots/helpers.sql
+++ b/dbt-adapters/src/dbt/include/global_project/macros/materializations/snapshots/helpers.sql
@@ -8,7 +8,7 @@
 {% macro default__create_columns(relation, columns) %}
   {% for column in columns %}
     {% call statement() %}
-      alter table {{ relation.render() }} add column "{{ column.name }}" {{ column.data_type }};
+      alter table {{ relation.render() }} add column {{ adapter.quote(column.name) }} {{ column.data_type }};
     {% endcall %}
   {% endfor %}
 {% endmacro %}


colin-k-rogers:
@stevetracvc mind adding a changelog? Just need to call `changie new` from the dbt-adapters package root

PR #1128, opened 2025-05-28
Use svv_redshift_columns to retrieve sort and distkey column names

### LINKED ISSUE #1096: [Bug] Redshift Materialized View Config Change Detection Always Fails for MV with GROUP BY
### Is this a new bug?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [x] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

In the dbt-redshift adapter, materialized view model configs are first verified against the current config in Redshift. This is done through a set of queries here - [https://github.com/dbt-labs/dbt-redshift/blob/main/dbt/include/redshift/macros/relations/materialized_view/describe.sql](url)

These queries use `svv_table_info` and an internal materialized view table `'mv_tbl__{{ relation.identifier }}__%'` in order to verify that sort and dist keys haven't changed. However, when a materialized view uses a GROUP BY statement (and perhaps other conditions?) the internal column names are changed on these system tables. I have verified this with AWS Redshift support though I cannot find any documentation that describes this process.

For example, here is the DDL for the ``'mv_tbl__{{ relation.identifier }}__%'`` - 

`CREATE TABLE IF NOT EXISTS dbt_test_gold.mv_tbl__vw_test__0
(
    	grvar_1 VARCHAR(64)   ENCODE RAW
    	,grvar_2 VARCHAR(64)   ENCODE lzo
    	,aggvar_1 TIMESTAMP WITHOUT TIME ZONE   ENCODE az64
    	,aggvar_2 TIMESTAMP WITHOUT TIME ZONE   ENCODE az64
    	,__mv_needs_recompute_aggvar_1 BOOLEAN   ENCODE RAW
    	,__mv_needs_recompute_aggvar_2 BOOLEAN   ENCODE RAW
    	,num_rec BIGINT   ENCODE az64
)
DISTSTYLE KEY
 DISTKEY (grvar_1)
 SORTKEY (
	grvar_1
	)`

The sortkey and the distkey have been changed internally to **grvar_1**. 

The result of this is the config comparison will always fail for any materialized view with a GROUP BY because these fields will never match the DBT model config. The materialized view will then be dropped/recreated every time instead of incrementally refreshing.

### Expected Behavior

The config comparison should verify that no changes have occurred between the DBT config and the existing materialized view and continue processing.

### Steps To Reproduce

1. Create a materialized view that includes a GROUP BY statement in Redshift via a DBT model
2. Execute an initial `dbt run` to create the materialized view
3. Execute another `dbt run` to trigger the config comparison
4. Observe the DROP and CREATE table

### Relevant log output

```shell

```

### Environment

```markdown
- OS:
- Python: 3.11.3
- dbt-adapters: 1.9.4
- <adapter>: dbt-redshift
```

### Additional Context

_No response_

peterferris-st:
resolves #1096
[docs] N/A

### Problem

  Materialized view change detection is not handled correctly for any materialized view that has a GROUP BY statement in the view definition. This is because Redshift implements an internal column-naming convention in the `mv_tbl__{{ relation.identifier }}__ `table as well as `svv_table_info`. Both of these are used to obtain the current materialized view config. These column names will never match the config in the DBT model and thus will always drop and re-create the materialized view.

### Solution

  This PR modifies the macro which retrieves the config for a materialized view to use the `svv_redshift_columns` system table which provides us the correct column names. The distribution key parsing code is then modified to obtain the column name from this query rather than the query using `svv_table_info`. Logic is added to only look for a key column if the distribution style is KEY. No changes are necessary for the sortkey parsing since this was already using the `_column_descriptor_sql` query.
  
  Also added two tests which handle the above cases.

### Checklist


peterferris-st:
Hi DBT Team - is this something you'd be interested in resolving? Any feedback on this approach?

PR #1129, opened 2025-05-28
bigquery storage_uri updates

colin-rogers-dbt:
Split off file_format + base testing into: https://github.com/dbt-labs/dbt-adapters/pull/1134
BigQuery Catalog Config will now support the following:
```
"name": "basic_iceberg_catalog_integration",
"catalog_type": "generic",
"file_format": "parquet",
"table_format": "iceberg",
"external_volume": "gs://my_bucket",
```
we then generate `storage_uri` based like `"<external_volume>/_dbt/<schema>/<model>"`

Model Config will now support changing this behavor
Full override via specifying `storage_uri`:
```
{{ config (
     "catalog": basic_iceberg_catalog_integration
      "storage_uri": "gs://my_different_bucket" # overrides the external volume
      )
 }}
```
OR
mirroring the functionality in dbt-snowflake, override the `_dbt` signifier with `base_location_root`:
```
{{ config (
     "catalog": basic_iceberg_catalog_integration
      "base_location_root": "different" # gs://my_bucket/different/<schema>/<model>
      )
 }}
```

dependent on: https://github.com/dbt-labs/dbt-core/pull/11695

resolves #

### Checklist


dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_biglake_metastore.py
```diff
@@ -0,0 +1,61 @@
...
+            storage_uri=storage_uri,
+        )
+
+    def _calculate_storage_uri(self, model: RelationConfig) -> Optional[str]:
+        if not model.config:
+            return None
+        storage_uri_base = ""
+        if self.external_volume:
+            storage_uri_base += self.external_volume
+        location_root = (
```
mikealfare, line 55: I think this is equivalent to:
```python
location_root = model.config.get("base_location_root", "_dbt")
```
I suppose users could pass in `base_location_root = none`?

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_biglake_metastore.py
```diff
@@ -0,0 +1,61 @@
...
+            table_format=self.table_format,
+            file_format=self.file_format,
+            external_volume=external_volume,
+            storage_uri=storage_uri,
+        )
+
+    def _calculate_storage_uri(self, model: RelationConfig) -> Optional[str]:
+        if not model.config:
+            return None
+        storage_uri_base = ""
```
mikealfare, line 52: I think this is equivalent to:
```python
storage_uri_base = self.external_volume or ""
```

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_biglake_metastore.py
```diff
@@ -0,0 +1,61 @@
...
+        if self.external_volume:
+            storage_uri_base += self.external_volume
+        location_root = (
+            base_location_root
+            if (base_location_root := model.config.get("base_location_root"))
+            else "_dbt"
+        )
+        storage_uri_base += f"/{location_root}"
+        return f"{storage_uri_base}/{model.schema}/{model.name}"
```
mikealfare, line 61: I would just move `/{location_root}` into this return statement.

mikealfare (commented):
I have a few readability suggestions in the linked comments below.

I would consider updating the changelog entries to be a bit more formal, e.g. they should be capitalized at least.

Last, I would consider breaking this PR up into two PRs. I would either split by adapter, or more likely, by update (as the PR title suggests). `file_format` and the `storage_uri` updates, taken individually, feel like atomic changes, but the combination does not. And if we need to roll one back, we may not want to impact the other, especially if the issue is with `storage_uri`, which has no relation to `dbt-snowflake`.

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_biglake_metastore.py
VersusFacit: They're not equivalent. `get` will respect any falsey value whereas the `bool()` call implicit in the `if` will force it to `_dbt`.

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -816,7 +816,7 @@ def get_table_options(
                 if catalog_relation.table_format == constants.ICEBERG_TABLE_FORMAT:
                     opts["table_format"] = f"'{catalog_relation.table_format}'"
                     opts["file_format"] = f"'{catalog_relation.file_format}'"
-                    opts["storage_uri"] = f"'{catalog_relation.external_volume}'"
+                    opts["storage_uri"] = f"'{catalog_relation.storage_uri}'"
```
VersusFacit, line 819: Will this break existing customers or are the semantics unchanged?

dbt-bigquery/.changes/unreleased/Features-20250529-092932.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Features
+body: generate storage_uri from external_volume and populate file_fomat from top level
```
VersusFacit, line 2: typo

dbt-bigquery/src/dbt/adapters/bigquery/constants.py
```diff
@@ -13,25 +13,24 @@
 
 
 GENERIC_CATALOG_TYPE = "generic"
```
VersusFacit, line 15: Is this still in use anywhere? Will it be in use later?

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/__init__.py
```diff
@@ -1,4 +1,4 @@
-from dbt.adapters.bigquery.catalogs._generic import (
+from dbt.adapters.bigquery.catalogs._biglake_metastore import (
```
VersusFacit, line 1: similar to question I asked in dbt-bigquery/src/dbt/adapters/bigquery/constants.py. Curious about conventions / choices

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_biglake_metastore.py
VersusFacit: > With the current implementation, these will all result in location_root='_dbt', which may be unexpected if the user has supplied this config. If we just grab it using get, then I'm guessing this will blow up as an invalid storage_uri since we'll wind up with:

Yes exactly 🤝 And that's the essence of my last sentence. What do we want for these edge cases?

mikealfare: Ah, I misread, then I'm in agreement. @amychen1776 could you provide some direction here?

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/__init__.py
mikealfare: Are you referring to single underscore versus dunder?

VersusFacit: why the change in name of the module?

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_biglake_metastore.py
amychen1776: I want it to be exactly like how we do it on snowflake so like this

```
If no inputs are provided, dbt will output for base_location {{ external_volume }}/_dbt/{{ schema }}/{{ model_name }}
If base_location_root = foo, dbt will output {{ external_volume }}/foo/{{ schema }}/{{ model_name }}
If base_location_subpath = bar, dbt will output {{ external_volume }}/_dbt/{{ schema }}/{{ model_name }}/bar
If base_location = foo and base_location_subpath = bar, dbt will output {{ external_volume }}/foo/{{ schema }}/{{ model_name }}/bar

```

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/__init__.py
colin-k-rogers: we're renaming the catalog type from generic -> biglake_metastore so I figured it makes sense to rename the module as well

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_biglake_metastore.py
colin-k-rogers: @amychen1776 added some unit tests based off your above comment

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_biglake_metastore.py
colin-k-rogers: updated

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
colin-k-rogers: There aren't any existing customers as this hasn't gone out in a final version so shouldn't be an issue

amychen1776: +1 here --> no one is using this today on dbt-bigquery so it's totally fine

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_relation.py
```diff
@@ -0,0 +1,16 @@
...
+from typing import Optional
+
+from dbt.adapters.catalogs import CatalogRelation
+
+from dbt.adapters.bigquery import constants
+
+
+@dataclass
+class BigQueryCatalogRelation(CatalogRelation):
+    catalog_type: str = constants.BIGLAKE_CATALOG_TYPE
```
mikealfare: I think this default should be `"INFO_SCHEMA"` (or better, a new constant with this value). Otherwise, I think the rest of the config will be overwritten by the catalog defaults on `DEFAULT_ICEBERG_CATALOG`.

dbt-bigquery/src/dbt/adapters/bigquery/catalogs/_info_schema.py
```diff
@@ -0,0 +1,25 @@
...
+        Args:
+            model: `config.model` (not `model`) from the jinja context
+        """
+
+        return BigQueryCatalogRelation(
+            catalog_type=self.catalog_type,
+            catalog_name=self.catalog_name,
+            table_format=self.table_format,
+            file_format=self.file_format,
+            external_volume=self.external_volume,
```
mikealfare: I think `external_volume` would also be `None`, just like `storage_uri`, but I could be missing a scenario.

dbt-bigquery/tests/functional/adapter/catalog_integrations/test_catalog_integration.py
```diff
@@ -0,0 +1,67 @@
...
+                {
+                    "name": "basic_iceberg_catalog",
+                    "active_write_integration": "basic_iceberg_catalog_integration",
+                    "write_integrations": [
+                        {
+                            "name": "basic_iceberg_catalog_integration",
+                            "catalog_type": "biglake_metastore",
+                            "file_format": "parquet",
+                            "table_format": "iceberg",
+                            "external_volume": f"gs://{_BQ_BUCKET}",
```
mikealfare: Minor nit, but I would move this up to the top alongside `_BQ_BUCKET` and `_STATIC_URI` for consistency (because you're technically building it by prefixing with `gs://`).

mikealfare (commented):
One concern and one nit.

PR #1139, opened 2025-06-04
[ADAP-803] Bugfix: Set is_iceberg to true when information contains table_type=ICEBERG

TJGRauh:
resolves [ADAP-803]

N/A

### Problem

Existing Iceberg tables are not recognized when using a Hive metastore:

The existing table '' is in another format than 'delta' or 'iceberg' or 'hudi'

### Solution

Check the output of the database scan for "table_type=ICEBERG" in addition to "Provider: iceberg" since we get "Provider: hive" for Iceberg tables registered in a Hive metastore

### Checklist

### DIFF
--- a/dbt-spark/src/dbt/adapters/spark/impl.py
+++ b/dbt-spark/src/dbt/adapters/spark/impl.py
@@ -214,7 +214,9 @@ class SparkAdapter(SQLAdapter):
             )
             is_delta: bool = "Provider: delta" in information
             is_hudi: bool = "Provider: hudi" in information
-            is_iceberg: bool = "Provider: iceberg" in information
+            is_iceberg: bool = (
+                "Provider: iceberg" in information or "table_type=ICEBERG" in information
+            )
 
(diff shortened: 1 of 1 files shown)


dbt-spark/src/dbt/adapters/spark/impl.py
```diff
@@ -214,7 +214,9 @@ def _build_spark_relation_list(
             )
             is_delta: bool = "Provider: delta" in information
             is_hudi: bool = "Provider: hudi" in information
-            is_iceberg: bool = "Provider: iceberg" in information
```
TJGRauh, line 217: I assume that delta and hudi tables registered in a Hive metastore have the same issue which could likely be fixed in the same way, but I have neither experience in working with delta or hudi nor the setup to verify this...

nstans: can confirm delta tables are experiencing the same issue. though similarly not across the details of this package so uncertain if the fix for delta tables will be the same.

PR #1158, opened 2025-06-16
add support for optional iceberg builtin catalog relation parameters

colin-rogers-dbt:
resolves #

### Checklist

### DIFF
--- a/dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_built_in.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_built_in.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from dbt_common.dataclass_schema import StrEnum
 from typing import Optional
 
 from dbt.adapters.catalogs import (
@@ -8,10 +9,29 @@ from dbt.adapters.catalogs import (
 from dbt.adapters.contracts.relation import RelationConfig
 
 from dbt.adapters.snowflake import constants, parse_model
+from dbt.adapters.snowflake.constants import SnowflakeIcebergTableRelationConfig
+
+
+class SnowflakeStorageSerializationOptions(StrEnum):
+    """
+    Represents the storage serialization options for Snowflake catalogs.
+    """
+
+    COMPATIBLE = "COMPATIBLE"
+    OPTIMIZED = "OPTIMIZED"
 
 
 @dataclass
 class BuiltInCatalogRelation:
+    """
+    Parameters representing configuration of a built-in Snowflake Iceberg table
+    Ref: https://docs.snowflake.com/en/sql-reference/sql/create-iceberg-table-snowflake#syntax
+
+    Note that a lot of these parameters are optional, and the defaults for are inherited from the
+    schema/database/account. To not add to the complexity of the user configuration, we do not set defaults for these
+    parameters.
+    """
+
     base_location: Optional[str]
     catalog_type: str = constants.DEFAULT_BUILT_IN_CATALOG.catalog_type
     catalog_name: Optional[str] = constants.DEFAULT_BUILT_IN_CATALOG.name
@@ -21,6 +41,10 @@ class BuiltInCatalogRelation:
     cluster_by: Optional[str] = None
     automatic_clustering: Optional[bool] = False
     is_transient: Optional[bool] = False
+    data_retention_time_in_days: Optional[int] = None
+    max_data_extension_time_in_days: Optional[int] = None
+    storage_serialization_policy: Optional[str] = None
+    change_tracking: Optional[str] = None
 
 
 class BuiltInCatalogIntegration(CatalogIntegration):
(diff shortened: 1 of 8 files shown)


colin-k-rogers:
@cla-bot check

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_built_in.py
```diff
@@ -29,20 +53,106 @@ class BuiltInCatalogIntegration(CatalogIntegration):
...
+    max_data_extension_time_in_days: Optional[int] = None
+    storage_serialization_policy = None
+    change_tracking = None
 
     def __init__(self, config: CatalogIntegrationConfig) -> None:
         # we overwrite this because the base provides too much config
         self.name: str = config.name
         self.external_volume: Optional[str] = config.external_volume
+        if adapter_properties := config.adapter_properties:
+            if storage_serialization_policy := adapter_properties.get(
```
mikealfare, line 70: The default `None` is unnecessary for each of these since that's already the default for `.get()`.

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_built_in.py
```diff
@@ -29,20 +53,106 @@ class BuiltInCatalogIntegration(CatalogIntegration):
...
+                SnowflakeIcebergTableRelationConfig.storage_serialization_policy, None
+            ):
+                self.storage_serialization_policy = SnowflakeStorageSerializationOptions(
+                    storage_serialization_policy
+                ).value
```
mikealfare: Needing to add `.value()` everywhere is part of why I stopped using `StrEnum`. I don't think we need that once we move to the built-in though, which I think is py3.10? Also, I think you need to call `value` (`value()`), but I could be wrong.

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_built_in.py
```diff
@@ -29,20 +53,106 @@ class BuiltInCatalogIntegration(CatalogIntegration):
...
+            max_data_extension_time_in_days=max_data_extension_time_in_days,
+            change_tracking=self._resolve_change_tracking(model),
+            data_retention_time_in_days=data_retention_time_in_days,
+        )
+
+    def _resolve_storage_serialization_policy(self, model: RelationConfig) -> Optional[str]:
+        """
+        Resolves the storage serialization policy for the catalog integration.
+        """
+        policy = (
```
mikealfare: This might be over-engineering, but this pattern shows up a lot. Have you considered a method like this:

```python
def get_catalog_attr(catalog: CatalogIntegration, model: RelationConfig, name: str) -> Any:

    catalog_default = get_attr(catalog, name)

    if model.config:
        parameter_name = get_attr(SnowflakeIcebergTableRelationConfig, name)
        return model.config.get(parameter_name, catalog_default)
    return catalog_default
```

The usage could look like:

```python
    policy = get_catalog_attr(self, model, "storage_serialization_policy")
```

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_built_in.py
```diff
@@ -29,20 +53,106 @@ class BuiltInCatalogIntegration(CatalogIntegration):
...
+        if (
+            model.config
+            and (
+                change_tracking := model.config.get(
+                    SnowflakeIcebergTableRelationConfig.change_tracking, self.change_tracking
+                )
+            )
+            is not None
+        ):
+            if isinstance(change_tracking, bool):
```
mikealfare: I think there's a utility method for handling `bool` and `str` versions of a boolean. If not, you could do this with a `dict`, which in my opinion is more readable:

```python

COERCE_BOOL_TO_STR = {
    True: "TRUE",
    False: "FALSE",
    "true": "TRUE",
    "false": "FALSE",
}

if isinstance(change_tracking, str):
    change_tracking = change_tracking.lower()

try:
    return COERCE_BOOL_TO_STR[change_tracking]
except KeyError:
    raise ValueError("...")
```

dbt-snowflake/src/dbt/adapters/snowflake/constants.py
```diff
@@ -27,3 +23,11 @@
     file_format=None,
     adapter_properties={},
 )
+
+SnowflakeIcebergTableRelationConfig = SimpleNamespace(
```
mikealfare: The name on this was confusing to me. When reading it in the code, I thought this was going to be some default values for these parameters, not the name of the parameters themselves. Perhaps something like `IcebergParameters` would be less confusing since `Config` tends to be values throughout dbt?

dbt-snowflake/tests/functional/adapter/catalog_integrations/test_catalog_integrations.py
```diff
@@ -36,10 +58,22 @@ def catalogs(self):
...
-            "models": {
-                "basic_iceberg_table.sql": MODEL__BASIC_ICEBERG_TABLE,
-            }
+            "basic_iceberg_table.sql": MODEL__BASIC_ICEBERG_TABLE,
+            "iceberg_table_with_configs.sql": MODEL__ICEBERG_TABLE_W_CONFIGS,
         }
 
     def test_basic_iceberg_catalog_integration(self, project):
         run_dbt(["run"])
+        iceberg_sql = get_cleaned_model_ddl_from_file("basic_iceberg_table.sql")
```
mikealfare, line 67: If you're only checking the model file, then I would run `dbt compile` instead of `dbt run`. It should be a bit quicker.

dbt-snowflake/tests/unit/test_catalog_relation_built_in.py
```diff
@@ -54,3 +54,35 @@ def test_iceberg_base_location_built_in(fake_integration, config, expected):
...
+            {"change_tracking": "False"},
+            "FALSE",
+        ),
+        (
+            {"change_tracking": "True"},
+            "TRUE",
+        ),
+    ],
+)
+def test_change_tracking_model_config(fake_integration, config, expected):
```
mikealfare, line 69: Minor nit, but this is a ton of whitespace, to the point where it's a bit clunky to read. I would remove the setting name in the parameterization:

```python

@pytest.mark.parametrize(
    "user_input,expected_ddl",
    [
        (None, None),
        (False, "FALSE"),
        (True, "TRUE"),
        ("False", "FALSE"),
        ("True", "TRUE"),
    ],
)
def test_change_tracking_model_config(fake_integration, user_input, expected_ddl):
    model = deepcopy(model_base)
    if user_input is not None:
        model.config.update({"change_tracking": user_input})
    relation = fake_integration.build_relation(model)
    assert relation.change_tracking == expected_ddl
```

I would also create a test for testing the error conditions here, since we're raising `ValueErrors` in the code:

```python

@pytest.mark.parametrize(
    "user_input",
    [
        0,
        "",
        "None",
        "yes",
        "apple",
    ],
)
def test_change_tracking_invalid_model_config(fake_integration, user_input):
    model = deepcopy(model_base)
    model.config.update({"change_tracking": user_input})
    with pytest.raises(ValueError) as e:
        fake_integration.build_relation(model)
        assert "Invalid value for change_tracking:" in e.msg
        assert user_input in e.msg
```

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_built_in.py
colin-k-rogers: On reflection I don't think I actually _need_ the enum here

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_built_in.py
colin-k-rogers: I think I can actually remove this

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_built_in.py
colin-k-rogers: I looked and don't think there is a utility method that I can see (there is something similar in dbt-core)

dbt-snowflake/tests/functional/adapter/catalog_integrations/test_catalog_integrations.py
colin-k-rogers: Just tried and it looks like 'compile' doesn't result in actually writing out the target files 🤷

dbt-snowflake/tests/unit/test_catalog_relation_built_in.py
colin-k-rogers: It is a bit clunky but it's nice that it shares the same pattern as the unit test above `test_iceberg_base_location_built_in`

colin-k-rogers: ended up updating

PR #1181, opened 2025-06-30
create table and column descriptions for snowflake iceberg tables

sandeepmullangi2:
resolves #
https://github.com/dbt-labs/dbt-adapters/issues/1015

### Problem

When we want to add description to table and columns for snowflake iceberg table, its failing because we have to use Alter command but existing code was using command called COMMENT which works for dynamic/internal tables. This code will solve the problem when iceberg table is created with comments (including comments for columns) using dbt-snowflake adapter
### Solution

1. Ensured we have 3 conditions each for dynamic table, iceberg table and internal tables and storing each of the table type in relation_type
2. We use above relation_type in next conditional statement and ensure ALTER is present for iceberg and for all other table types COMMENT is enough.

### Checklist


dbt-snowflake/src/dbt/include/snowflake/macros/adapters.sql
```diff
@@ -95,10 +95,17 @@
 {% macro snowflake__alter_relation_comment(relation, relation_comment) -%}
     {%- if relation.is_dynamic_table -%}
         {%- set relation_type = 'dynamic table' -%}
+    {%- elif relation.is_iceberg_format -%}
```
colin-k-rogers: don't think we need this elif since your below check doesn't use the relation_type variable

Issue #657, opened 2025-01-15
[Bug] Extremely slow query to determine relations that uses all disk space (160GB)


michalc:
### Is this a new bug?

- [x] I believe this is a new bug
- [x] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When running `dbt --log-level debug run` (after following https://docs.getdbt.com/guides/manual-install?step=1, but for PostgreSQL), I see it gets stuck on this query:

```sql
with relation as (
    select
        pg_rewrite.ev_class as class,
        pg_rewrite.oid as id
    from pg_rewrite
),
class as (
    select
        oid as id,
        relname as name,
        relnamespace as schema,
        relkind as kind
    from pg_class
),
dependency as (
    select distinct
        pg_depend.objid as id,
        pg_depend.refobjid as ref
    from pg_depend
),
schema as (
    select
        pg_namespace.oid as id,
        pg_namespace.nspname as name
    from pg_namespace
    where nspname != 'information_schema' and nspname not like 'pg\_%'
),
referenced as (
    select
        relation.id AS id,
        referenced_class.name ,
        referenced_class.schema ,
        referenced_class.kind
    from relation
    join class as referenced_class on relation.class=referenced_class.id
    where referenced_class.kind in ('r', 'v', 'm')
),
relationships as (
    select
        referenced.name as referenced_name,
        referenced.schema as referenced_schema_id,
        dependent_class.name as dependent_name,
        dependent_class.schema as dependent_schema_id,
        referenced.kind as kind
    from referenced
    join dependency on referenced.id=dependency.id
    join class as dependent_class on dependency.ref=dependent_class.id
    where
        (referenced.name != dependent_class.name or
         referenced.schema != dependent_class.schema)
)

select
    referenced_schema.name as referenced_schema,
    relationships.referenced_name as referenced_name,
    dependent_schema.name as dependent_schema,
    relationships.dependent_name as dependent_name
from relationships
join schema as dependent_schema on relationships.dependent_schema_id=dependent_schema.id
join schema as referenced_schema on relationships.referenced_schema_id=referenced_schema.id
group by referenced_schema, referenced_name, dependent_schema, dependent_name
order by referenced_schema, referenced_name, dependent_schema, dependent_name;
```
After ~50 minutes, it then errors:
```
08:30:45  Finished running  in 0 hours 53 minutes and 15.01 seconds (3195.01s).
08:30:45  Encountered an error:
Database Error
[...]

PR #1253, opened 2025-08-10
Optimize extremely slow query that finds Postgres relations

michalc:
resolves https://github.com/dbt-labs/dbt-adapters/issues/657

### Problem

The macro postgres_get_relations in relations.sql was extremely slow and used an extremely high amount of temporary disk space on a system with high numbers of schemas, tables, and dependencies between database objects (rows in pg_depend). Slow to the point of not completing in 50 minutes and using more than 160GB disk space (at which point PostgreSQL ran out of disk space and aborted the query).

### Solution

The solution here optimises the query and so it runs in just under 1 second on my system. It does this by being heavily inspired by the definition of information_schema.view_table_usage, and specifically:

- Stripping out CTEs that can be optimisation blockers, often by causing CTEs to be materialised to disk (especially in older PostgreSQL, but I suspect in recent too in some cases).
- Removing unnecessary filtering on relkind: going via pg_rewrite (or rather, the equivalent row on pg_depend) is equivalent to that
- Avoiding sequential scans on any table by structuring joins/where clause to leverage indexes, especially on pg_depend
- Removing unnecessary filtering out system catalog tables from dependents (they are excluded by the remaining filters on referenced tables).
- Not having `select distinct ... from pg_dependent` in the innards of the query, and instead having a top level `select distinct` - on my system this saved over 45 seconds.
- Excluding self-relations that depend on themselves by using oid rather than using the names of tables and schemas.

I suspect this is also more robust because oids I think _can_ be repeated between system tables, and so when querying pg_depend filtering on classid and refclassid is required (and I think also means indexes are better leveraged).

Comparing calls to `explain` it reduces the largest "rows" value from 5,284,141,410,595,979 (over five quadrillion) to 219 and the actual run time from never completing within 50 minutes (because it used all of the 160GB available) to completing in ~500ms.

It also has some style/naming changes:

- Using a `distinct` on the top level rather than a group by for clarity (performance seemed the same in my case).
- Flips the definition of "referenced" and "dependent" in the query to match both the definitions in pg_depend, and the code at https://github.com/dbt-labs/dbt-adapters/blob/3325afe0b2cd9cede991736f872d7b6739a47038/dbt-postgres/src/dbt/adapters/postgres/impl.py#L113
- Re-orders the join to I think a slightly clearer order that "flows" from views -> the linking table (pg_depend) to the tables referenced in the views.
- Lowers the abstraction/indirection levels in naming/aliases, using names closer to the PostgreSQL catalog tables - this made it easier to write and understand, and so I suspect easier to make changes in future (I found I had to keep in mind the PostgreSQL definitions more than the output of the query when making changes).

### Checklist


dbt-postgres/src/dbt/include/postgres/macros/relations.sql
```diff
@@ -7,68 +7,39 @@
...
+        and joining_depend.refobjid != dependent_depend.refobjid
+
+    -- ... we can find the tables they query from in pg_class, but excluding system tables. Note we
+    -- don't need need to exclude _dependent_ system tables, because they only query from other
+    -- system tables, and so are automatically excluded by excluding _referenced_ system tables
+    join pg_class as referenced_class on referenced_class.oid = joining_depend.refobjid
+    join pg_namespace as referenced_namespace on referenced_namespace.oid = referenced_class.relnamespace
+        and referenced_namespace.nspname != 'information_schema'
+        and referenced_namespace.nspname not like 'pg\_%'
+
```
colin-k-rogers, line 40: I think we want to limit the kind of objects this pulls in:
```suggestion
where dependent_class.relkind in ('r', 'v', 'm')
```

colin-k-rogers:
> Comparing calls to explain it reduces the largest "rows" value from 5,284,141,410,595,979 (over five quadrillion) to 219 and the actual run time from never completing within 50 minutes (because it used all of the 160GB available) to completing in ~500ms.

This is absolutely wild, we haven't touched this in a while so I'm not surprised there's room for improvement here.

What version(s) of postgres have you tested this against?

michalc:
@colin-rogers-dbt 

> This is absolutely wild,

Yep!

> What version(s) of postgres have you tested this against?

Just 14 (which I know is a few versions away from the latest)

dbt-postgres/src/dbt/include/postgres/macros/relations.sql
michalc: I _think_ this is already limited by going via (the pg_depend references to) pg_rewrite? From https://www.postgresql.org/docs/current/catalog-pg-rewrite.html

> The catalog pg_rewrite stores rewrite rules for tables and views.

So if there are rows in pg_rewrite, it _has_ to be for a table or a view?

The runtime is admittedly extremely similar for me with/without the filter on relkind, so maybe it's doesn't matter too much, but all things being equal, it's more performant to not have unnecessary filters, especially since in this case the filter on relkind isn't covered by an index?

PR #1257, opened 2025-08-12
override default varchar datatype when altering iceberg table columns

colin-rogers-dbt:
Problem: Adding new varchar column to an existing iceberg table fails because we default to the snowflake standard size which is larger than the max iceberg column width. 

Solution: Add a macro to override default behavior when altering columns for iceberg tables. 

This is a somewhat hacky stop gap solution, likely we'll need to re-assess how we handle varchar size going forward to better account for new edge cases.

### Checklist


QMalcolm (approved):

will-sargent-dbtlabs:
@colin-rogers-dbt What about arrays of varchars?  Does this also impact that, or do we need to mitigate for that as well?  I have a customer with iceberg tables that have these.

PR #1332, opened 2025-09-11
Base adapter and postgres support function lifecycle

### LINKED ISSUE #1268: [UDFs] Base adapter and postgres adapter support the lifecyle of UDFs
### Housekeeping

- [x] I am a maintainer of dbt-adapters

### Short description

UDFs are being added to dbt-core! As such, the adapters need to support creating and replacing UDFs in the data warehouse. Each data warehouse is likely to require slightly different syntaxes for doing so. This issue is specifically about adding the support to the `base adapter` and the `postgres adapter`. 

### Acceptance criteria

* Base adapter provides an interface for creating/replacing UDFs
* Postgres adapter supports creating/replacing UDFs

### Suggested Tests

* Can a UDF be created in postgres
* Can a UDF be replaced in postgres

### Impact to Other Teams

N/A

### Will backports be required?

No, it's a new feature

### Context

* https://github.com/dbt-labs/dbt-core/discussions/11851

QMalcolm:
resolves #1268
related https://github.com/dbt-labs/dbt-core/pull/12008

### Problem

We need to support creating functions!

### Solution

Create a function materialization

### Checklist


dbt-postgres/hatch.toml
```diff
@@ -12,7 +12,7 @@ pre-install-commands = [
 ]
 dependencies = [
     "dbt-common @ git+https://github.com/dbt-labs/dbt-common.git",
-    "dbt-core @ git+https://github.com/dbt-labs/dbt-core.git#subdirectory=core",
+    "dbt-core @ git+https://github.com/dbt-labs/dbt-core.git@qmalcolm--11965-core-handles-lifecycle-of-function-nodes#subdirectory=core",
```
QMalcolm: Typically I'd say we should change this back before merging. However, we're merging to a feature branch, not `main`. As such it is safe to leave as is.

dbt-postgres/hatch.toml
```diff
@@ -67,7 +67,7 @@ pre-install-commands = [
...
```
QMalcolm: Typically I'd say we should change this back before merging. However, we're merging to a feature branch, not main. As such it is safe to leave as is.

dbt-tests-adapter/hatch.toml
```diff
@@ -11,7 +11,7 @@ pre-install-commands = [
 ]
 dependencies = [
     "dbt_common @ git+https://github.com/dbt-labs/dbt-common.git",
-    "dbt-core @ git+https://github.com/dbt-labs/dbt-core.git#subdirectory=core",
+    "dbt-core @ git+https://github.com/dbt-labs/dbt-core.git@qmalcolm--11965-core-handles-lifecycle-of-function-nodes#subdirectory=core",
```
QMalcolm: Typically I'd say we should change this back before merging. However, we're merging to a feature branch, not main. As such it is safe to leave as is.

dbt-adapters/src/dbt/include/global_project/macros/materializations/functions/function.sql
```diff
@@ -0,0 +1,19 @@
+{% materialization function, default %}
```
QMalcolm, line 1: I expect most of this logic to get pushed down a level once we introduced a second function type (UDTFs or UDAFs)

dbt-adapters/src/dbt/include/global_project/macros/materializations/functions/function.sql
```diff
@@ -0,0 +1,19 @@
...
+    {% if not adapter.supports(adapter.capability_for_string('UDFs')) %}
```
colin-k-rogers: Worth noting that we don't do this check for any other kind of materialization. Typically we just proceed optimistically and let the adapter raise an exception if it's not supported. 

also, nit: should `UDFs` just be `functions` since that's the materialization?

dbt-adapters/src/dbt/adapters/base/impl.py
```diff
@@ -1915,10 +1915,16 @@ def render_model_constraint(cls, constraint: ModelLevelConstraint) -> Optional[s
     def capabilities(cls) -> CapabilityDict:
         return cls._capabilities
 
+    @available
```
colin-k-rogers: if we don't need to do the supports check then we don't need this. Avoiding adding more to the implicit adapter interface is always nice

dbt-adapters/src/dbt/include/global_project/macros/materializations/functions/function.sql
QMalcolm: So I only saw two ways forward:
1. The base adapter implements a default macro per function type (UDF, UDAF, UDTF) per language (SQL, Python, Javascript, etc) that raises a not implemented exception
2. The base adapter implements a base implementation per function type per language, and then the individual adapters can opt in by declaring the capability as supported.

The perk of (2) is that it means the 1st/3rd party adapter has less to implement. It can override simply one of the sub helper macros. For instance, if the arg specification is slightly different, then all they need to override is the`get_formatted_udf_args` macro.

Do we want to go the route of (1) or is there another route I'm missing? 🙃

colin-k-rogers: To me option 3 is that the base adapter implements a base implementation per function type per language but instead of asking the adapter to opt in ask it to opt out. This is what we generally have done: proceed on the assumption that the default postgres like macros work or will be overridden by the adapter as needed. 

Not saying this is the _best_ approach, just the one we have done in the past

dbt-adapters/src/dbt/include/global_project/macros/materializations/functions/function.sql
```diff
@@ -0,0 +1,16 @@
+{% materialization function, default %}
+    {% set existing_relation = load_cached_relation(this) %}
+    {% set target_relation = this.incorporate(type=this.Function) %}
+
+    {{ run_hooks(pre_hooks) }}
+
+    {% set function_type_macro = get_function_macro('scalar', 'sql') %}
```
QMalcolm, line 7: We should stop hard coding these variables once we start supporting other function types + languages

colin-k-rogers: these can also be supplied by a method on the RelationConfig

PR #1368, opened 2025-10-03
Set client prefetch threads to 1 to prevent deadlock at exit

wiggzz:
Workaround for https://github.com/snowflakedb/snowflake-connector-python/issues/2213

### Problem

Snowflake queries can hang at exit

### Solution

This hanging only happens when there is a deadlock during the prefetch phase of the query - if there is only 1 thread, it cannot deadlock.

### Checklist

### DIFF
--- a/dbt-snowflake/src/dbt/adapters/snowflake/connections.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/connections.py
@@ -394,6 +394,7 @@ class SnowflakeConnectionManager(SQLConnectionManager):
                     application="dbt",
                     insecure_mode=creds.insecure_mode,
                     session_parameters=session_parameters,
+                    client_prefetch_threads=1, # disable client prefetch because it can cause deadlock at exit, see https://github.com/snowflakedb/snowflake-connector-python/issues/2213
                     **creds.auth_args(),
                 )
 


wiggzz:
Actually, looks like https://github.com/snowflakedb/snowflake-connector-python/commit/b301717582ce4543996206520e285c6ad86b640f is in `main` which will mean the next release should fix this. Unclear what the timeline on a new release is though.

wiggzz:
> Actually, looks like [snowflakedb/snowflake-connector-python@b301717](https://github.com/snowflakedb/snowflake-connector-python/commit/b301717582ce4543996206520e285c6ad86b640f) is in `main` which will mean the next release should fix this. Unclear what the timeline on a new release is though.

Ok the new snowflake connector version fix is in but it requires a specific timeout parameter to be set. I'll see how difficult it is to upgrade to 4.0.0 and go from there.

dbt-snowflake/src/dbt/adapters/snowflake/connections.py
```diff
@@ -410,6 +410,7 @@ def connect():
                     insecure_mode=creds.insecure_mode,
                     platform_detection_timeout_seconds=creds.platform_detection_timeout_seconds,
                     session_parameters=session_parameters,
+                    ocsp_root_certs_dict_lock_timeout=10,  # cert lock can cause deadlock without timeout, see https://github.com/snowflakedb/snowflake-connector-python/issues/2213
```
colin-k-rogers, line 413: is there a reasoning behind `10`? Can we make this a user configurable value?

wiggzz: My surface level understanding is that this is the amount of time that the connector will wait to obtain a lock to interact with some certs it uses during the request. I doubt that this will ever take longer than 10 seconds unless there is a deadlock, in which case we'll fail after 10 seconds. So I doubt there is a reason a user would care or want to configure it, I am personally a little confused why the connector exposes this at all. I think this should have been an implementation detail. But I may be misreading it, I'll try to take a deeper look tomorrow.

colin-k-rogers: Thanks, we can go with that for now and do a fast follow to make this configurable

kang8:
Hi! I noticed this PR was merged on Nov 3rd. Is there a timeline for when this will be included in a tagged release? Thanks!

PR #1385, opened 2025-10-12
Support BigQuery STRUCT schema change

### LINKED ISSUE #599: [CT-1710] [CT-1703] [Feature] `on_schema_change` should handle non-top-level schema changes

### Describe the feature

Currently, `on_schema_change` tracks schema changes involving top-level columns. This excludes nested changes, like the addition of a column in a `STRUCT` (BigQuery). It would be great to extend the functionality of `on_schema_change` to track changes in the schema for top-level columns as well as nested ones, too.

We've had to update schemas manually to accommodate nested changes. Or, we write the data to new tables and then migrate those new tables to a production context.

### Who will this benefit?

This type of feature will benefit those with meaty incremental models who do not want to or cannot afford to do a full refresh. This benefits those with models that change somewhat frequently - more holistic, automated handling of schema changes would be fantastic!

Sure! If you all see this as a feasible feature extension, I am happy to dig into the code and contribute.

### Anything else?

At the very least, I think the documentation around `on_schema_change` should be updated to make it very clear that it deals with top-level schema changes.

Kayrnt:
resolves #599

### Problem

Previously, dbt-bigquery did not support schema changes for STRUCT (nested) fields in incremental models. When new fields were added to STRUCT columns or existing fields were removed, the table schema was not updated before incremental materialization operations (like MERGE), leading to potential data inconsistencies or failures. 

### Solution

This PR implements STRUCT field synchronization for BigQuery incremental models by:

- Fixing the macro dispatch mechanism in the global `on_schema_change.sql` to properly use `adapter.dispatch` for adapter-specific overrides
- Adding the `@available` decorator to the `sync_struct_columns` method in `impl.py` to expose it to Jinja templates
- Implementing logic to detect and add missing STRUCT fields using BigQuery's schema update API before MERGE operations
- Adding graceful error handling for BigQuery's limitation that prevents removing fields from STRUCT columns (documented as a known limitation)
- Updating the BigQuery-specific `on_schema_change.sql` macro to call `sync_struct_columns` during schema change processing

The implementation supports `append_new_columns` mode fully, while `sync_all_columns` mode is skipped with appropriate warnings due to BigQuery API constraints (because of the unsupported nested column removal).
Functional tests have been added to cover the functionality.

### Checklist

### DIFF
--- a/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
@@ -721,14 +721,9 @@ class BigQueryAdapter(BaseAdapter):
             table = client.get_table(table_ref)
             schema_as_dicts = [field.to_api_repr() for field in table.schema]
 
-        apply_schema_patch = False
-
         if add_columns:
             additions = build_nested_additions(add_columns)
             schema_as_dicts = merge_nested_fields(schema_as_dicts, additions)
-            apply_schema_patch = True
-
-        if apply_schema_patch:
             new_schema = [SchemaField.from_api_repr(field) for field in schema_as_dicts]
             new_table = google.cloud.bigquery.Table(table_ref, schema=new_schema)
             client.update_table(new_table, ["schema"])


dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
```diff
@@ -107,7 +107,7 @@
 
     {% else %}
 
-      {% set schema_changes_dict = check_for_schema_changes(source_relation, target_relation) %}
+      {% set schema_changes_dict = adapter.dispatch('check_for_schema_changes', 'dbt')(source_relation, target_relation) %}
```
colin-k-rogers: let's update `check_for_schema_changes` to follow our standard pattern where it contains the dispatch logic + a default macro 

```
E         In dispatch: No macro named 'check_for_schema_changes' found within namespace: 'dbt'
E             Searched for: 'test.redshift__check_for_schema_changes', 'test.postgres__check_for_schema_changes', 'test.default__check_for_schema_changes', 'dbt.redshift__check_for_schema_changes', 'dbt.postgres__check_for_schema_changes', 'dbt.default__check_for_schema_changes'
```

Kayrnt: Oups! I didn't check other adapters 😱 
Will do 👍

colin-k-rogers:
@Kayrnt looks like we're now seeing some errors in incremental:
```
17:21:05  
17:21:05  Completed with 1 error, 0 partial successes, and 0 warnings:
17:21:05  
17:21:05  Failure in model incremental_deeply_nested_struct_append (models/incremental_deeply_nested_struct_append.sql)
17:21:05    Database Error in model incremental_deeply_nested_struct_append (models/incremental_deeply_nested_struct_append.sql)
  Value of type STRUCT<l1_field STRING, l1_new_field STRING, level2 STRUCT<l2_field STRING, l2_new_field STRING, level3 STRUCT<l3_field STRING, l3_new_field STRING>>> cannot be assigned to payload, which has type STRUCT<l1_field STRING, level2 STRUCT<l2_field STRING, level3 STRUCT<l3_field STRING, l3_new_field STRING>, l2_new_field STRING>, l1_new_field STRING> at [24:53]
  compiled code at target/run/test/models/incremental_deeply_nested_struct_append.sql
```

Kayrnt:
Right, I'll see why it was apparently passing on my end and fix it 👀

colin-k-rogers:
> Right, I'll see why it was apparently passing on my end and fix it 👀

If you can double check your local that would be great, if it's an issue with our CI config I can investigate

Kayrnt:
Just a quick update that it took me way more time to figure out the issues and have a fix. I think I'm on the right track and will be able to deliver the proper version later this week. Stay tuned! 😄

Kayrnt:
New tests all pass on my end and I ran code quality stuff too 🤞

dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
```diff
@@ -60,17 +65,22 @@
...
 
 {% macro sync_column_schemas(on_schema_change, target_relation, schema_changes_dict) %}
+  {{ return(adapter.dispatch('sync_column_schemas', 'dbt')(on_schema_change, target_relation, schema_changes_dict)) }}
+{% endmacro %}
+
+
+{% macro default__sync_column_schemas(on_schema_change, target_relation, schema_changes_dict) %}
 
   {%- set add_to_target_arr = schema_changes_dict['source_not_in_target'] -%}
+  {%- set remove_from_target_arr = schema_changes_dict['target_not_in_source'] -%}
```
Kayrnt, line 75: FYI though not directly visible here: the 2 variables are called in

```
{% set schema_change_message %}
    In {{ target_relation }}:
        Schema change approach: {{ on_schema_change }}
        Columns added: {{ add_to_target_arr }}
        Columns removed: {{ remove_from_target_arr }}
        Data types changed: {{ new_target_types }}
  {% endset %}
```
that is after the if block.
So it was not consistent because they would not be initialized.

Kayrnt:
@colin-rogers-dbt for the integration tests on snowflake:
```
Too many schemas in database DBT_TEST! dbt can only get
information about databases with fewer than 10000 schemas.
```

I'll let you handle that.
It was missing a changie at dbt-adapters (the root package) so I added one the last commit (+ rebase).
Everything else is passing 🙌

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -651,19 +654,275 @@ def update_table_description(
...
+                "Attempted to remove: {}. Consider using 'append_new_columns' mode "
+                "or recreating the table with full_refresh.".format(nested_removals)
+            )
 
-        new_table = google.cloud.bigquery.Table(table_ref, schema=new_schema)
-        client.update_table(new_table, ["schema"])
+        if drop_candidates:
+            relation_name = relation.render()
+            for column in drop_candidates:
+                drop_sql = f"ALTER TABLE {relation_name} DROP COLUMN {self.quote(column.name)}"
```
colin-k-rogers: 1. Should we do this via a macro? (not really sure)
2. Can we group these drops into a single call? i.e. can we do this: 
```ALTER TABLE `project.dataset.table_name`
DROP COLUMN column_name_1,
DROP COLUMN column_name_2,
DROP COLUMN column_name_3;```

File: dbt-bigquery/src/dbt/adapters/bigquery/impl.py [hunk 4]
```diff
@@ -651,19 +654,275 @@ def update_table_description(
...
+                logger.debug(
+                    'Dropping column `{}` from table "{}".'.format(column.name, relation_name)
+                )
+                client.query(drop_sql).result()
```
colin-k-rogers: instead of calling client.query we should probably use the `execute` method so we use a consistent retry / exception handling logic

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
Kayrnt: 1. Well that's a design decision, where else would you do it?
2. Right, that's also possible to do it according to https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#alter_table_drop_column_statement

Kayrnt:
I'm not getting any notification when there's a validation for CI so I was just waiting and finally got back on the PR to see the precommit failure 😢
I just pushed the simple fix... let me know when you will be able to add back the label @colin-rogers-dbt 😅 
Thanks!

Kayrnt:
Hum I tried to add a merge commit to avoid UT failures but it also removes the label 😅

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -651,19 +654,277 @@ def update_table_description(
...
+            else:
+                merged_fields.append(copy.deepcopy(field))
+
+        for path, addition in addition_lookup.items():
+            if "." not in path:
+                merged_fields.append(copy.deepcopy(addition))
+
+        return merged_fields
+
+    def _collect_field_dicts(
```
colin-k-rogers: Can we move these functions (`_collect_field_dicts`, `_find_missing_fields` `_merge_nested_fields`)  into a separate module (something like `struct_utils.py`)?

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -651,19 +660,197 @@ def update_table_description(
...
+                schema_changes_dict["source_not_in_target"] = [
+                    column
+                    for column in schema_changes_dict["source_not_in_target"]
+                    if column.name.split(".", 1)[0] not in handled_columns
+                ]
+
+            schema_changes_dict["target_columns"] = [
+                BigQueryColumn.create_from_field(field) for field in new_schema
+            ]
+        except google.api_core.exceptions.BadRequest as exc:
```
colin-k-rogers, line 839: why not raise this exception? If we fail to update the struct and a downstream model depends on it then that model will fail right?

Kayrnt: The initial intent was to stick to a non blocking behavior (like it was before this change).
However if you're ok with it, I'm ok to "fail fast" so I added a commit for that purpose.

Kayrnt:
@colin-rogers-dbt is the PR ok? I see unrelated failures.

colin-k-rogers:
> @colin-rogers-dbt is the PR ok? I see unrelated failures.

@Kayrnt yea we have been having some issues with our functional tests for python models. Working on that separately.

Check out this PR in the meantime: https://github.com/bqbooster/dbt-adapters/pull/1

Kayrnt:
Looks good, I merged your changes in my branch.

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -651,19 +658,196 @@ def update_table_description(
...
+            schema_as_dicts = [field.to_api_repr() for field in table.schema]
+
+        apply_schema_patch = False
+
+        if add_columns:
+            additions = build_nested_additions(add_columns)
+            schema_as_dicts = merge_nested_fields(schema_as_dicts, additions)
+            apply_schema_patch = True
+
+        if apply_schema_patch:
```
MichelleArk: small nit: is `apply_schema_patch` necessary? perhaps the following lines that compute the new schema + update the table could just be part of the `if add_columns:` handling above.

Kayrnt: Right, it's overkill, I simplified it 👍

colin-rogers-dbt (approved):

colin-k-rogers:
@Kayrnt can you update from main? Would like to get this merged so we can release next week

Kayrnt:
Sure, I hit rebase 👍

PR #1057, opened 2025-05-02
Fix #1056: Keep table and column metadata as strings

aranke:
resolves #1056

### Problem

Agate was inferring numeric table comments as numbers, which led to conflicts when merging columns.

### Solution

- Always treat table comments as text.
- Expand to more table and column metadata to always treat as text.
- 🎩 that the solution looks good in [catalog.json](https://github.com/user-attachments/files/20011883/catalog.json)

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #1404, opened 2025-10-23
🩹 Remove table_owner from text only cols

### LINKED ISSUE #1135: [Bug] (BigQuery) docs generate: "table_owner" does not match the name of any column in this table
### Is this a new bug?

- [x] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [x] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

When running `dbt docs generate`, the following warning appears repeatedly before completion:
```
agate/type_tester.py:92: RuntimeWarning: "table_owner" does not match the name of any column in this table.
  warnings.warn('"%s" does not match the name of any column in this table.' % name, RuntimeWarning)
``` 

### Expected Behavior

No RuntimeWarning messages

### Steps To Reproduce

1. dbt-core 1.9.4, dbt-bigquery 1.9.1
2. profiles with bigquery as connection type
3. Run dbt docs 
4. The warning message appears repeatedly after Building catalog message

### Relevant log output

```shell
#31 4.523 21:09:52  Running with dbt=1.9.4
#31 8.064 21:09:55  Registered adapter: bigquery=1.9.1
#31 8.311 21:09:55  Unable to do partial parsing because saved manifest not found. Starting full parse.
#31 46.64 21:10:34  Found x models, y snapshots, z data tests, a seeds, b operations, i sources, m exposures, n macros
#31 46.95 21:10:34  
#31 46.95 21:10:34  Concurrency: 8 threads (target='dev')
#31 46.95 21:10:34  
#31 246.3 21:13:53  Building catalog
#31 262.1 /lib/python3.11/site-packages/agate/type_tester.py:92: RuntimeWarning: "table_owner" does not match the name of any column in this table.
#31 262.1   warnings.warn('"%s" does not match the name of any column in this table.' % name, RuntimeWarning)
...repeated many times
#31 341.7 21:15:29  Catalog written to /dbt/targets/catalog.json
```

### Environment

```markdown
- OS: macOS or Linux-5.15.0
- Python: 3.11.12
- dbt-core: 1.9.4
- dbt-bigquery: 1.9.1
```

### Additional Context

This bug / warning seems to have been introduced after PR #1057  is merged, specifically due to the addition of `table_owner` to the `text_only_columns` list.  It looks like the BigQuery version of the catalog macro (https://github.com/dbt-labs/dbt-adapters/blob/f38ece924ce79b2bd049bfaa73eeda817f84e989/dbt-bigquery/src/dbt/include/bigquery/macros/catalog/catalog.sql) does not have the `table_owner` field. Therefore, agate issues the warning in the `_catalog_filter_table()` method.

The warning goes away if `table_owner` is removed from the list.

morgan-dgk:
### Problem
PR #1057 added several new column names to the `text_only_columns` arg to the `table_from_rows` call in the `_catalog_filter_table` class method. This method is called as part of the `dbt docs generate` task.

However, some of the new column names included are not included in the list of columns returned by certain adapter implementations for the `__get_catalog` and `__get_catalog_relations`.
For example, the relevant BigQuery and Athena macros do not return a column named `table_owner` (see #1135).

This results in the agate type checker emitting a warning when attempting typing checking runs.

### Solution
Any columns added to the `text_only_columns` argument for the call to `table_from_rows` in the base implementation should presumably be columns that are returned by all adapters implementations.

This PR removes the `table_owner` name from the list of text_only_columns included in the base implementation of `_catalog_filter_table` and moves this to a Snowflake-specific adapter implementation.

I investigated whether agate provides a means of suppressing this warning but there does not appear to be a mechanism to do so. Additionally, I am unsure whether this would be desirable. 

resolves #1135

### Checklist

### DIFF
--- a/dbt-adapters/src/dbt/adapters/base/impl.py
+++ b/dbt-adapters/src/dbt/adapters/base/impl.py
@@ -1329,7 +1329,6 @@ class BaseAdapter(metaclass=AdapterMeta):
                 "table_name",
                 "table_type",
                 "table_comment",
-                "table_owner",
                 "column_name",
                 "column_type",
                 "column_comment",
--- a/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
@@ -118,10 +118,30 @@ class SnowflakeAdapter(SQLAdapter):
     def _catalog_filter_table(
         cls, table: "agate.Table", used_schemas: FrozenSet[Tuple[str, str]]
     ) -> "agate.Table":
+        from dbt_common.clients.agate_helper import table_from_rows
+        from dbt.adapters.base.impl import _catalog_filter_schemas
+
         # On snowflake, users can set QUOTED_IDENTIFIERS_IGNORE_CASE, so force
         # the column names to their lowercased forms.
+
         lowered = table.rename(column_names=[c.lower() for c in table.column_names])
-        return super()._catalog_filter_table(lowered, used_schemas)
+
+        table = table_from_rows(
+            lowered.rows,
+            lowered.column_names,
+            text_only_columns=[
+                "table_database",
+                "table_schema",
+                "table_name",
+                "table_type",
+                "table_owner",
+                "table_comment",
+                "column_name",
+                "column_type",
+                "column_comment",
+            ],
+        )
+        return table.where(_catalog_filter_schemas(used_schemas))
 
     def _make_match_kwargs(self, database, schema, identifier):
         # if any path part is already quoted then consider same casing but without quotes


dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -120,8 +120,27 @@ def _catalog_filter_table(
     ) -> "agate.Table":
         # On snowflake, users can set QUOTED_IDENTIFIERS_IGNORE_CASE, so force
         # the column names to their lowercased forms.
+        from dbt_common.clients.agate_helper import table_from_rows
+        from dbt.adapters.base.impl import _catalog_filter_schemas
+
```
colin-k-rogers: nit: move imports above comment
```suggestion
        from dbt_common.clients.agate_helper import table_from_rows
        from dbt.adapters.base.impl import _catalog_filter_schemas
        
       # On snowflake, users can set QUOTED_IDENTIFIERS_IGNORE_CASE, so force
       # the column names to their lowercased forms.
```

dbt-adapters/src/dbt/adapters/base/impl.py
colin-k-rogers, line 1326: one other option here would be to add a method that adapters implement to override this instead of needing to override the whole method

i.e. 
```
def _get_text_only_cols() -> List[str]
    return [...]
```

Then we snowflake / whomever can implement this method to override the default

morgan-dgk: Yep, any preference @colin-rogers-dbt? I was initially looking to keep this change as small as possible.

PR #1414, opened 2025-10-29
Support the volatility config of function nodes

### LINKED ISSUE #1345: [UDFs] A function's `volatility` is propagated to the create/replace function logic

### Describe the feature

In https://github.com/dbt-labs/dbt-core/issues/12030 we're adding the concept of `volatility` to functions. In adapters we should ensure that:
 
* If a volatility is specified for a function, then the create function statement run against the data warehouse uses the appropriate volatility keyword
* If volatility is None for a function, then no volatility keyword is used in the create function statement run against the data warehouse.

_No response_

### Who will this benefit?

_No response_

Yes

### Anything else?

_No response_

QMalcolm:
resolves #1345

### Problem

dbt-core added the ability to set function node volatility in https://github.com/dbt-labs/dbt-core/pull/12100. Thus in adapters we needed to begin interpolating the correct volatility value into the function materialization macros.

### Solution

* Create new macros for interpolating volatility for a given function node
* Test that the volatility gets set (and still successfully creates the function)

### Checklist

### DIFF
--- a/dbt-redshift/src/dbt/include/redshift/macros/materializations/functions/scalar.sql
+++ b/dbt-redshift/src/dbt/include/redshift/macros/materializations/functions/scalar.sql
@@ -14,12 +14,16 @@
 {% endmacro %}
 
 {% macro redshift__scalar_function_volatility_sql() %}
-    {% if model.config.get('volatility') == 'deterministic' %}
+    {% set volatility = model.config.get('volatility') %}
+    {% if volatility == 'deterministic' %}
         IMMUTABLE
-    {% elif model.config.get('volatility') == 'stable' %}
+    {% elif volatility == 'stable' %}
         STABLE
+    {% elif volatility == 'non-deterministic' or volatility == none %}
+        VOLATILE
     {% else %}
-        {# At this point, either they've set `non-deterministic` or they've set nothing. In either case, we default to VOLATILE #}
+        {% do unsupported_volatility_warning(volatility) %}
+        {# We're ignoring the unknown volatility. But redshift requires a volatility to be set, so we default to VOLATILE #}
         VOLATILE
     {% endif %}
 {% endmacro %}
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/materializations/functions/scalar.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/materializations/functions/scalar.sql
@@ -13,12 +13,13 @@
 {% endmacro %}
 
 {% macro snowflake__scalar_function_volatility_sql() %}
-    {% if model.config.get('volatility') == 'deterministic' %}
+    {% set volatility = model.config.get('volatility') %}
+    {% if volatility == 'deterministic' %}
         IMMUTABLE
-    {% elif model.config.get('volatility') == 'stable' %}
-        {% do exceptions.warn("`Stable` function volatility is not supported by Snowflake, and will be ignored") %}
-    {% elif model.config.get('volatility') == 'non-deterministic' %}
+    {% elif volatility == 'non-deterministic'%}
         VOLATILE
+    {% elif volatility == 'stable' or volatility != none %}
+        {% do unsupported_volatility_warning(volatility) %}
     {% endif %}
     {# If no volatility is set, don't add anything and let the data warehouse default it #}
 {% endmacro %}
--- a/dbt-adapters/src/dbt/include/global_project/macros/materializations/functions/scalar.sql
+++ b/dbt-adapters/src/dbt/include/global_project/macros/materializations/functions/scalar.sql
@@ -45,12 +45,25 @@
 {% endmacro %}
 
 {% macro default__scalar_function_volatility_sql() %}
-    {% if model.config.get('volatility') == 'deterministic' %}
+    {% set volatility = model.config.get('volatility') %}
+    {% if volatility == 'deterministic' %}
         IMMUTABLE
-    {% elif model.config.get('volatility') == 'stable' %}
+    {% elif volatility == 'stable' %}
         STABLE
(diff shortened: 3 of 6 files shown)


dbt-snowflake/src/dbt/include/snowflake/macros/materializations/functions/scalar.sql
```diff
@@ -10,3 +11,14 @@
        {{ model.compiled_code }}
     $$
 {% endmacro %}
+
+{% macro snowflake__scalar_function_volatility_sql() %}
+    {% if model.config.get('volatility') == 'deterministic' %}
+        IMMUTABLE
+    {% elif model.config.get('volatility') == 'stable' %}
+        {% do exceptions.raise_compiler_error("`Stable` function volatility is not supported for Snowflake") %}
```
QMalcolm: Alternatively we could
* Ignore `stable`
* Raise a warning

The benefit of doing so would be that switching a project from another data warehouse to snowflake wouldn't break functions declared "stable". The drawback of not raising an error is that a user might set `stable` for a function in snowflake, think doing so is doing so does something, but it doesn't 🤔 
cc: @graciegoheen

QMalcolm: Relatedly, in dbt-bigquery we're just _ignoring_ volatility. What we decided to do with `stable` in `dbt-snowflake` is also what we should do with _any_ volatility setting in `dbt-bigquery`

dbt-redshift/src/dbt/include/redshift/macros/materializations/functions/scalar.sql
```diff
@@ -14,3 +12,14 @@
...
+
+{% macro redshift__scalar_function_volatility_sql() %}
+    {% if model.config.get('volatility') == 'deterministic' %}
+        IMMUTABLE
+    {% elif model.config.get('volatility') == 'stable' %}
+        STABLE
+    {% else %}
+        {# At this point, either they've set `non-deterministic` or they've set nothing. In either case, we default to VOLATILE #}
+        VOLATILE
+    {% endif %}
```
colin-k-rogers: nit:
```suggestion
    {% set volatility = model.config.get('volatility') %}
    {% if volatility == 'deterministic' %}
        IMMUTABLE
    {% elif volatility == 'stable' %}
        STABLE
    {% else %}
        {# At this point, either they've set `non-deterministic` or they've set nothing. In either case, we default to VOLATILE #}
        VOLATILE
    {% endif %}
```

colin-k-rogers: also do we need to do any validation/warning if they have set something other than these options?

colin-k-rogers: (also in the future if we add another option do we need to remember to come back and raise a warning?)

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/functions/scalar.sql
```diff
@@ -10,3 +11,14 @@
...
 {% endmacro %}
+
+{% macro snowflake__scalar_function_volatility_sql() %}
+    {% if model.config.get('volatility') == 'deterministic' %}
+        IMMUTABLE
+    {% elif model.config.get('volatility') == 'stable' %}
+        {% do exceptions.warn("`Stable` function volatility is not supported by Snowflake, and will be ignored") %}
+    {% elif model.config.get('volatility') == 'non-deterministic' %}
+        VOLATILE
+    {% endif %}
```
colin-k-rogers: is this more future proof?
```suggestion
    {% set volatility = model.config.get('volatility') %}
    {% if volatility == 'deterministic' %}
        IMMUTABLE
    {% elif model.config.get('volatility') == 'non-deterministic' %}
        VOLATILE
     {% elif volatility is not none %}
        {% do exceptions.warn("user passed unsupported volatility config value: " ~ volatility ~"; it will be ignored") %}
    {% endif %}
```

dbt-tests-adapter/pyproject.toml
```diff
@@ -26,6 +26,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
+    "dbt-common>=1.34.0,<2.0",
```
colin-k-rogers, line 29: can we let dbt-adapters set the range for dbt-common here?

QMalcolm: We can. My thought here was that we're specifying this range specifically for testing purposes (the `EventCatcher` is only used in testing). If putting it in `dbt-adapters` is more straight forward though, I'm good with doing so.

dbt-redshift/src/dbt/include/redshift/macros/materializations/functions/scalar.sql
QMalcolm: Core only accepts an `Enum` of possible volatility types `deterministic`, `stable`, and `non-deterministic`. So in order for a 4th option to come into existence, we'd first have to extend the enum in core. Additionally, have a hard time imagining any other volatility type being invented, but perhaps I'm being unimaginative 🤷🏻 Future proofing doesn't _hurt_ anything.

QMalcolm: Okay I think I've now added logic to sufficiently handle this 🙂

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/functions/scalar.sql
QMalcolm: Okay I think I've now added logic to sufficiently handle this 🙂

PR #1418, opened 2025-11-01
Issue 429 athena adapter partitioning

### LINKED ISSUE #429: [Bug] `truncate()` partition transformation does not work when it includes more than 100 partitions

### Current Behavior

When you use a truncate() partition transformation for a column that will result in more than 100 partitions, the batch partitioning functionality starts up and allows you to exceed 100 partitions. 

```
{{
    config(
        materialized = 'table',
        table_type = 'iceberg',
        force_batch = true,
        partitioned_by = ['truncate(string_partition,2)', 'month(date_partition)']
    )
}}
```

However, when the query reaches out to Athena to pull in the distinct partitions, it uses truncate() in the query which is not a supported method of extracting values from a string in Athena.

```
select distinct truncate(string_partition,2), date_trunc('month', date_partition)
from "awsdatacatalog"."data_lake"."table__ha__tmp_not_partitioned"
order by truncate(string_partition,2), date_trunc('month', date_partition)
```

Instead, it could use something like substring() to pull back the unique partial values

```
select distinct substring(string_partition,1,2), date_trunc('month', date_partition)
from "awsdatacatalog"."data_lake"."table__ha__tmp_not_partitioned"
order by substring(string_partition,1,2), date_trunc('month', date_partition)
```

### Expected Behavior

When I do a `truncate()` Iceberg partition transformation on a column, it is capable of handling something with greater than 100 partitions.

### Steps To Reproduce

Create a model with a column that when a partition transformation of truncate() is used, it will result in more than 100 partitions.

### Environment

```markdown
- OS: MacOS
- Python: 3.11
- dbt: 1.7.7
- dbt-athena-community: 1.7.1
```

### Additional Context

This is out of a Slack conversation: https://getdbt.slack.com/archives/C013MLFR7BQ/p1709755667814619

This method was referenced as where the work would need to be changed: https://github.com/dbt-athena/dbt-athena/blob/289be4f4f44f3d5a6cf575d8fe218209c4a41171/dbt/adapters/athena/impl.py#L1279

Apache Iceberg Truncate Partition documentation: https://iceberg.apache.org/spec/#truncate-transform-details

fyang2223:
resolves #429

### Problem

When using `truncate()` function in Athena partition keys for string columns, the SQL generated is invalid because `truncate()` [only works on decimal places](https://trino.io/docs/current/functions/math.html#truncate) in Athena/Trino. String columns must use `substr()` instead, and for numeric columns, we need to apply division. The `get_partition_batches` macro was generating invalid SQL like `truncate(str_col, 1)` which would cause query failures.

### Solution

- Added logic to detect truncate partition keys and convert `truncate(col, N)` to `substr(col, 1, N)` for non-numeric columns, and `floor(col / N)` for numeric columns.
- Overrode `is_integer()` and `is_numeric()` in `AthenaColumn` to properly recognize Athena dtype values (`int`, `bigint`, `decimal(P,S)`) 
- Updated the `get_partition_batches` macro to always fetch column info from the temporary unpartitioned Athena Iceberg relation.

### Checklist


fyang2223:
The PR currently involves changes to the function signature of `format_partition_keys` and `format_one_partition_key`, which could be a breaking change. Alternatively, instead of putting the column datatype logic inside the python functions, we could extract it to the caller, the get_partition_batches macro. Adding the following excerpt does the trick, and would satisfy the 'no interface changes' condition:

```jinja
    ...
    {%- set athena_partitions_limit = config.get('partitions_limit', 100) | int -%}

    {# --- Start new code below --- #}

    {# Get column info from relation (for truncate transformation) #}
    {%- set relation_columns = adapter.get_columns_in_relation(sql) -%}

    {# Transform truncate() partition keys based on column types #}
    {%- set transformed_partitioned_by = [] -%}
    {%- for partition_key in partitioned_by -%}
        {%- set truncate_match = modules.re.search('truncate\\((.+?),\\s*(\\d+)\\)', partition_key.lower()) -%}
        {%- if truncate_match -%}
            {%- set col_name = truncate_match.group(1) -%}
            {%- set width = truncate_match.group(2) -%}

            {# Find the column type #}
            {%- set column = None -%}
            {%- for col in relation_columns -%}
                {%- if col.name.lower() == col_name.lower() -%}
                    {%- set column = col -%}
                {%- endif -%}
            {%- endfor -%}

            {# Transform based on column type #}
            {%- if column and not column.is_numeric() -%}
                {%- do transformed_partitioned_by.append('substr(' ~ col_name ~ ', 1, ' ~ width ~ ')') -%}
            {%- else -%}
                {%- do transformed_partitioned_by.append('floor(' ~ col_name ~ ' / ' ~ width ~ ')') -%}
            {%- endif -%}
        {%- else -%}
            {# No transformation needed, keep original #}
            {%- do transformed_partitioned_by.append(partition_key) -%}
        {%- endif -%}
    {%- endfor -%}

    {%- set partitioned_keys = adapter.format_partition_keys(transformed_partitioned_by) -%}

    {# --- End new code above --- #}

    {% do log('PARTITIONED KEYS: ' ~ partitioned_keys) %}
    ...
```

colin-k-rogers:
@fyang2223 thanks for the PR can you add a changie entry summarizing the PR?

fyang2223:
Thanks @colin-rogers-dbt, I've added the changie entry

colin-k-rogers:
@fyang2223 this change looks good, can we update our functional tests to include the repro case in the linked issue?
```
{{
    config(
        materialized = 'table',
        table_type = 'iceberg',
        force_batch = true,
        partitioned_by = ['truncate(string_partition,2)', 'month(date_partition)']
    )
}}
```
You should be able to add it as a test fixture here: https://github.com/dbt-labs/dbt-adapters/blob/8c2272169b466f09120e946743012847c3de964c/dbt-athena/tests/functional/adapter/test_incremental_microbatch.py#L18

fyang2223:
Thanks for the guidance @colin-rogers-dbt, I've made the necessary modifications

PR #1439, opened 2025-11-15
athena: extend Lake Formation tagging and data filters

adavoudi:
### Problem

* `dbt-athena` is missing several Lake Formation capabilities that exist in the `dbt-glue` adapter (database-level tags, better data cell filter controls, more flexible behavior for tag/filter removal).
* Current LF tag behavior is effectively “authoritative” and can break inherited tags and external tag management.
* Data cell filters don’t support `all_rows`, column include/exclude patterns, or a non-destructive mode.
* There are existing unit-test warnings around catalog filtering in `AthenaAdapter`.
* Sample mode tests using naive datetimes cause an integration test failure in `dbt-athena/tests/functional/adapter/test_sample_mode.py:40`.

> Many of the concepts and shapes in this PR are borrowed from, or closely aligned with, the implementation and docs in [aws-samples/dbt-glue](https://github.com/aws-samples/dbt-glue)

### Solution

* **Lake Formation tags**
  * Extend `lf_tags_config` with:
    * `drop_existing` (default `false`) to choose between additive vs authoritative tag management.
    * `tags_database` for database-level tags.
    * `inherited_tags` (clarified) to protect inherited keys from removal when `drop_existing: true`.
  * Update tag processing to:
    * Optionally remove tags on database/table/columns when `drop_existing: true` (respecting `inherited_tags`).
    * Only add/update tags when `drop_existing: false`.
  * Update README to document the new options, clarify database vs table vs column behavior, and note that:
    * dbt manages LF tag associations only.
    * All LF tags must exist in Lake Formation ahead of time.

* **Data cell filters (`lf_grants`)**
  * Extend `data_cell_filters` with:
    * `drop_existing` (default `false`) to control whether filters missing from config are deleted.
    * Per-filter options:
      * `row_filter` **or** `all_rows: true`.
      * `column_names` **or** `excluded_column_names` (omit both for all columns).
  * Add validation so invalid `row_filter` / `all_rows` combinations fail fast.
  * Improve update logic to only update filters when row/column semantics actually change.
  * Document the new options and provide a fuller example (row predicate + column scoping).

* **Adapter / tests**

  * `dbt-athena/src/dbt/adapters/athena/impl.py`:
    * Add `_CATALOG_TEXT_COLUMNS` and `_catalog_filter_table` using `_catalog_filter_schemas` to address unit-test warnings around catalog filtering and text-only columns.
  * Sample mode tests:
    * Update `now` to `datetime.datetime.now(datetime.timezone.utc)` in:
      * `dbt-athena/tests/functional/adapter/test_sample_mode.py`
      * `dbt-bigquery/tests/functional/adapter/test_sample_mode.py`
    * This fixes the integration test error in `dbt-athena/tests/functional/adapter/test_sample_mode.py:40` caused by naive vs timezone-aware datetime comparisons.
  * Extend Lake Formation unit tests to cover:
    * The adjusted tag response shape.
    * `FilterConfig` validation, API representation, and update detection.

### Checklist


adavoudi:
I just signed the CLA

Issue #343, opened 2024-11-07
[Bug] dbt-core and dbt-adapters version mismatch


guenp:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

After running `pip install dbt-core dbt-adapters dbt-duckdb`:

```
$ pip list | grep dbt
dbt-adapters              1.7.0
dbt-common                1.10.0
dbt-core                  1.8.8
dbt-duckdb                1.9.0
dbt-extractor             0.5.1
dbt-semantic-interfaces   0.5.1

$ dbt --version                                                                                                                   
Core:
  - installed: 1.8.8
  - latest:    1.8.8 - Up to date!

Plugins:
  - duckdb: 1.9.0 - Not compatible!
  ```

~`dbt-core` complains when the adapter version does not match the `dbt-core` version (see [line](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/version.py#L134)). Because we released `dbt-duckdb` 1.9.0 before `dbt-adapters` 1.9.0 got released that's an error that shows up.~ `dbt-adapters` 1.9.0 was just released! 🙌 

Another issue we're seeing is that `dbt-core` 1.8.8 depends on on `dbt-adapters` <1.7.1 (see [line](https://github.com/dbt-labs/dbt-core/blob/v1.8.8/core/setup.py#L79)) but the latest version is 1.8.0 (which just came out last week)
It seems like the solution to this would be for `dbt-core` to release a new version 1.9.0

### Expected Behavior

No errors or version conflicts and should install the latest versions:

```
$ pip list | grep dbt
dbt-adapters              1.9.0
dbt-common                1.12.0
dbt-core                  1.8.8
dbt-duckdb                1.9.0
dbt-extractor             0.5.1
dbt-semantic-interfaces   0.7.4

$ dbt --version
Core:
  - installed: 1.8.8
  - latest:    1.8.8 - Up to date!

Plugins:
  - duckdb: 1.9.0 - Up to date!
```

### Steps To Reproduce

Run `pip install dbt-core dbt-adapters dbt-duckdb` in a new Python environment and run

```
$ pip show | grep dbt
$ dbt --version
```

### Relevant log output

```shell
N/A
```

### Environment

```markdown
- OS:N/A
- Python:N/A
- dbt-adapters:1.8.8
```

### Additional Context

N/A

joellabes:
[...]

Issue #226, opened 2023-09-12
[CT-3112] [Feature] Pass model configuration to execute()


SoumayaMauthoorMOJ:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Have the ability to pass in model, snapshot, seed and other configuration to the [cursor()](https://github.com/dbt-labs/dbt-core/blob/5182e3c40c9aadcafdb77e7fdcd6091ae0277444/core/dbt/adapters/sql/connections.py#L134) method, in addition to credentials saved to `profiles.yml`.

### Who will this benefit?

This has been raised by at least two adapters:

1. dbt-athena: https://github.com/dbt-athena/dbt-athena/issues/377
2. dbt-trino: https://github.com/starburstdata/dbt-trino/pull/323#discussion_r1261282679

PR #1455, opened 2025-11-19
Add 'execution_project' option to model configuration for BigQuery adapter

bordway-shop:
related to [#343](https://github.com/dbt-labs/dbt-bigquery/issues/343) / [#226](https://github.com/dbt-labs/dbt-adapters/issues/226)

### Problem
BigQuery model execution is defined at the `profiles.yml`. This PR introduces a per-resource `execution_project` configuration that lets each model explicitly choose which GCP project to use for query execution at run time. This can be useful for when the model selector is used for multiple models; perhaps one model in that grouping is a known, heavy model that needs to utilize a BigQuery project with more slots available. This allows the user to have a more fine-grained control over which BigQuery project a specific model is run in.

The default behavior is not changed and the execution project specified in the `profiles.yml` is used by default.

This solution provides a similar result as the [Snowflake connector configuration](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#configuring-virtual-warehouses)

### Solution
The `BigQueryConnectionManager` now maintains a cache of BigQuery clients keyed by execution_project, protected by a lock. Each dbt thread records which execution project it’s currently using so we can restore it later.

`pre_model_hook` reads config.execution_project, which can be either a plain string or a `{target-name: project-id}` mapping. If the configuration specifies a different project than the profile’s default, the adapter calls switch_execution_project, which:

1. Either reuses a cached client for that project or creates one on demand then
2. Swaps the thread’s connection handle to point at the new client.
3. `post_model_hook` then restores the previous `execution_project` returned by `pre_model_hook`, ensuring subsequent models don’t inherit the override.

Cached clients are marked so they aren’t closed prematurely, and `cleanup_all` now properly closes everything in the cache.

### Checklist


bordway-shop:
Closing this PR in favor of reservation-level settings which were merged in [here](https://github.com/dbt-labs/dbt-adapters/pull/1843) and gives similar functionality

PR #1468, opened 2025-11-23
Warn when underscores in Snowflake account names are replaced with hyphens

### LINKED ISSUE #1467: [Feature] Add warning when Snowflake account names with underscores are automatically normalized to hyphens

### Describe the feature

The Snowflake adapter automatically replaces underscores (`_`) with hyphens (`-`) in the account field of the snowflake credentials. This behavior is mentioned in issues #1143 and #687, but currently the transformation happens silently.

I ran into this on my team as well. Our Snowflake account name includes an underscore, and even though the correct long term fix is to update our account name, I think it would be nice to warn when underscores are detected and replaced to improve debugging.

### Who will this benefit?

This will benefit users whose Snowflake account names contain underscores (`_`). Instead of the normalization happening silently, they will receive a clear warning indicating that their account name was modified. This makes it easier to understand and debug connection issues that may arise from the transformed URL.

Yes I would be happy to contribute this and I think the solution is straight forward.

### Anything else?

_No response_

nishan-soni:
resolves #1467 

### Problem

This PR adds a warning when underscores (`_`) in the Snowflake account field are automatically replaced with hyphens (`-`) to form a valid account URL. Previously, this normalization happened silently, which could make debugging connection failures difficult.

### Solution

Emit an `AdapterEventWarning` if any underscores in the account name were substituted.

## Testing
Manual tests to confirm the warning was logged.

### Checklist

### DIFF
--- a/.github/workflows/codeowners-check.yml
+++ b/.github/workflows/codeowners-check.yml
@@ -1,0 +1,22 @@
+name: CODEOWNERS Coverage Check
+permissions:
+  contents: read
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  validate-codeowners:
+    if: github.actor != 'dependabot[bot]'
+    runs-on: depot-ubuntu-24.04
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # actions/checkout@v6
+      - uses: mszostok/codeowners-validator@7f3f5e28c6d7b8dfae5731e54ce2272ca384592f #v0.7.4
+        env:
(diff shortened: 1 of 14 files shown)


colin-k-rogers:
If someone has set their project to error on warnings we could cause their project to unexpectedly fail. If the purpose is to make it easier to debug could we just emit a debug log instead?

nishan-soni:
> If someone has set their project to error on warnings we could cause their project to unexpectedly fail. If the purpose is to make it easier to debug could we just emit a debug log instead?

I can switch this to a debug log 👍

colin-k-rogers:
Thanks, appreciate the change

nishan-soni:
Do we need to re-run the CI pipeline 😁

PR #1502, opened 2025-12-16
record BigQueryAdapter is_replaceable, describe_table, record BaseAdapter submit_python_job

MichelleArk:
resolves #N/A

### Problem

Several commonly used, BigQuery-specific adapter methods do not have recording structures available.

### Solution

Add them!

### Checklist

### DIFF
--- a/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
@@ -76,7 +76,6 @@ from dbt.adapters.bigquery.python_submissions import (
 )
 from dbt.adapters.bigquery.record.record_types import (
     BigQueryAdapterDescribeRelationRecord,
-    BigQueryAdapterGetBqTableRecord,
     BigQueryAdapterIsReplaceableRecord,
 )
 from dbt.adapters.bigquery.relation import BigQueryRelation
@@ -1105,12 +1104,6 @@ class BigQueryAdapter(BaseAdapter):
         return opts
 
     @available.parse(lambda *a, **k: True)
-    @record_function(
-        BigQueryAdapterGetBqTableRecord,
-        method=True,
-        index_on_thread_id=True,
-        id_field_name="thread_id",
-    )
     def get_bq_table(self, relation: BigQueryRelation) -> Optional[BigQueryTable]:
         try:
             table = self.connections.get_bq_table(
--- a/dbt-bigquery/src/dbt/adapters/bigquery/record/record_types.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/record/record_types.py
@@ -1,44 +1,9 @@
 import dataclasses
 from typing import Optional, List, Union
 
-from google.cloud.bigquery import Table as BigQueryTable
-
 from dbt_common.record import Record, Recorder
 from dbt.adapters.bigquery.relation import BigQueryRelation
 from dbt.adapters.bigquery.relation_configs import PartitionConfig
-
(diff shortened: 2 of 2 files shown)


dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -573,6 +578,12 @@ def _clusters_match(table, conf_cluster) -> bool:
         return table.clustering_fields == conf_cluster
 
     @available.parse(lambda *a, **k: True)
+    @record_function(
+        BigQueryAdapterIsReplaceableRecord,
+        method=True,
+        index_on_thread_id=True,
+        id_field_name="thread_id",
+    )
```
xuliangs, line 587: Do you think we could replace the `self.connections.get_bq_table` to `self. get_bq_table`, so that remote request gets recorded?

In this case, fusion can replay the remote request, exercise the rest of the its implementation and verify the result against the recording of this method

dbt-bigquery/src/dbt/adapters/bigquery/record/record_types.py
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+
+@dataclasses.dataclass
+class BigQueryAdapterGetBqTableResult:
+    return_val: Optional["BigQueryTable"]
+
+    def _to_dict(self):
+        return {
+            "return_val": self.return_val.to_dict() if self.return_val else None,
```
MichelleArk: Does the `BigQueryTable` return_val actually have a `to_dict` method available?

From my quick test, it does not:
```python
>>> from google.cloud.bigquery import Table as BigQueryTable
>>> BigQueryTable.to_dict
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: type object 'Table' has no attribute 'to_dict'
```

MichelleArk: I think we'll need to introduce a more custom implementation here to successfully serialize the BigQueryTable. Tested recording building a model defined as: 
```sql
select 1 as id
{% do adapter.get_bq_table(this) %}
```

recording.json contains:
```json
    {
        "type": "RecordingError",
        "record_type": "BigQueryAdapterGetBqTableRecord",
        "error": "'Table' object has no attribute 'to_dict'"
    },
```

xuliangs: Good catch! I ended up not recording the `get_bq_table`, and this method is not seen used in any internal macro. I'll remove these types.

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
```diff
@@ -101,7 +101,7 @@ def __init__(self, parsed_model: Dict, credentials: BigQueryCredentials) -> None
                 "Need to supply dataproc_cluster_name in profile or config to submit python job with cluster submission method"
             )
 
-    def submit(self, compiled_code: str) -> Job:
+    def submit(self, compiled_code: str) -> PythonSubmissionResult:
```
colin-k-rogers, line 104: If we're updating this for bigquery do we also need to update this for snowflake & databricks?

xuliangs: Great question! we don't need it for Snowflake since the Python models submission is via `execute` and validated by the existing recording there. We decided not to worry about Databricks conformance, but I put it in the shared module so we could add them when we need to in the future

PR #1487, opened 2025-12-03
Preserve collation when altering a column size

colin-rogers-dbt:
resolves #1369 
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#

### Problem

Because we currently ignore column collation if we run with `on_schema_change=true` we will try to remove column collation which Snowflake will trigger an exception on.

### Solution

In the fullness of time perhaps we will have a proper way of enabling users to manage collation on their various models and schemas but alas today is not that day. Instead this PR merely aims to preserve the collation so we don't break snowflake's rules and try to remove collation on an already existing table. 

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #1510, opened 2025-12-18
add expanded_data_type field on Column to more safely handle non-standard data type cases

colin-rogers-dbt:
resolves #

### Problem
So in #1487 I was committed the cardinal sin of being too cavalier with changing the `data_type` field. Specifically doing this breaks users when they try  use `data_type` in a cast function like:

```
cast({{ col_nmame }} as {{ col.data_type }}) as {{ col.quoted }} 
```

### Solution

Instead as the incomparable @joellabes has pointed out to me we could add _a new_ field. While I hesitate at adding yet another expansions to the Column interface I think the long term risk with messing with `data_type` is just too big.  

### Checklist

### DIFF
--- a/dbt-snowflake/src/dbt/adapters/snowflake/column.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/column.py
@@ -51,13 +51,10 @@ class SnowflakeColumn(Column):
     @property
     def expanded_data_type(self) -> str:
         """Include collation clause for string types where present."""
-        base_type = super().data_type
-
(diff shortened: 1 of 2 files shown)


gavinmackinlay:
@colin-rogers-dbt I see this is merged but has this been resolved? A number of our tier 1 models are still failing

colin-k-rogers:
@gavinmackinlay this change has been released in `1.10.6` and on 'latest' dbt Platform

PR #1078, opened 2025-05-09
Drop target table before cloning

jeremyyeo:
resolves #1077 
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#

### Problem
BQ requires actual dropping of the table before cloning (`create or replace ...` is not sufficient) if the partition spec changes.

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #1525, opened 2026-01-02
Set grant_config before calling apply_grants

### LINKED ISSUE #1503: [Regression] Cloning nodes removes grants in v1.10
### Is this a regression?

- [x] I believe this is a regression in functionality
- [x] I have searched the existing issues, and I could not find an existing issue for this regression

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [x] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

After upgrading from `dbt-core` and `dbt-bigquery` version 1.9 to version 1.10, we've observed that cloning nodes removes grants configured on the node in the target schema.

### Expected/Previous Behavior

On `dbt-core` and `dbt-bigquery` version 1.9, cloning nodes did not affect grants configured on the node in the target schema.

### Steps To Reproduce

1. In a `profiles.yml` file, configure two targets, `staging` and `production`.
1. Install `dbt-core` and `dbt-bigquery` at latest patch of version 1.9
1. Grant the BigQuery Data Viewer role on a model to a principal, in the model's schema file
1. Build model to `production` target
1. Confirm that grant was applied
1. Build model to `staging` target, specifying a `--target-path`
1. Clone model from `staging` target to `production` target
  a. Pass `--target-path` value from previous step to `--defer-state`
  b. Include `--full-refresh` flag
1. Observe that grant on production table is no longer present

### Relevant log output

```shell

```

### Environment

```markdown
- OS: macOS Tahoe 26.1
- Python: 3.12.4
- dbt-adapters (working version): 1.22.0
- dbt-adapters (regression version): 1.22.0
- dbt-core (working version): 1.9.10
- dbt-core (regression version): 1.10.16
- bigquery (working version): 1.9.2
- bigquery (regression version): 1.10.3
```

### Additional Context

_No response_

katieclaiborne-duet:
resolves #1503
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#N/A

### Problem

In BigQuery, cloning a model with `grants` defined will remove those grants from the target table.

### Solution

Define `grant_config`, so that `apply_grants` isn't always a no-op.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,25 @@ This repository is a monorepo containing the following packages:
 
 Please refer to each of these packages for more specific information.
 
+## Releases
+
+All of our packages are merged off of main except for `dbt-adapters` and `dbt-tests-adapter`. Therefore merging a pull request to main does not automatically put it in the queue for the next release. To do so please add the 'promote to stable' label to the PR once it's been merged.
+
+The reason we do this is to allow us to patch the previous minor version with updates (i.e. what's in stable) as needed while preparing what's on main (the next minor release) to be ready for release.
+
+### Upcoming Minor Releases
+
+The following milestones track features that we're planning for the next minor version release of each adapter:
+
+- [dbt-athena v1.11.0](https://github.com/dbt-labs/dbt-adapters/milestone/3) (2 PRs)
+- [dbt-bigquery v1.12.0](https://github.com/dbt-labs/dbt-adapters/milestone/4) (31 PRs)
+- [dbt-postgres v1.11.0](https://github.com/dbt-labs/dbt-adapters/milestone/5) (6 PRs)
+- [dbt-redshift v1.11.0](https://github.com/dbt-labs/dbt-adapters/milestone/2) (24 PRs)
+- [dbt-snowflake v1.12.0](https://github.com/dbt-labs/dbt-adapters/milestone/6) (23 PRs)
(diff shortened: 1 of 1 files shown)


katieclaiborne-duet:
I've identified https://github.com/dbt-labs/dbt-adapters/pull/1078 as the source of the regression! Moving back to draft to develop a more targeted fix.

katieclaiborne-duet:
Take 3: I think [4f8dcd9](https://github.com/dbt-labs/dbt-adapters/pull/1525/commits/4f8dcd90cadb02d5cd910af98d50804e4abd82d6) was really all I needed to do after all. It's fine if we always drop the target table in BigQuery clone operations, as long as we [apply grants](https://github.com/dbt-labs/dbt-adapters/blob/0462dbfe277323bd6ac581fae80adbe0bc0dfbd5/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/clone/clone.sql#L45) afterward.

Prior to #1078, grants weren't being applied, but it didn't matter, because they were preserved in the clone operation (create / replace). Since #1078, grants still aren't being applied, and it does matter, because they're destroyed in the clone operation (drop, create / replace).

I'll simplify the changes here to only correct for the missing apply grants behavior, rather than also introducing conditional drops during cloning.

PR #1527, opened 2026-01-05
Avoid dropping collation when updating column size

colin-rogers-dbt:
resolves #

### Problem
Previous fix preserved the collation for the alter when both source/target had collation set but didn't handle the scenarios where the collation between the two was different

### Solution
~~Override `can_expand_to` which is already set up to gate running alter (invoked [here](https://github.com/dbt-labs/dbt-adapters/blob/ab44f46a39e2271250651ccd117b871829316d38/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/column_helpers.sql#L44)) for columns during `on_schema_change`.~~

Overriding `can_expand_to` became difficult to reason about and wouldn't support still updating the columns instead it's much cleaner to dispatch `diff_column_data_types` to dbt-snowflake so we can add collation logic there. 

### Checklist

### DIFF
--- a/dbt-snowflake/tests/functional/adapter/incremental/test_incremental_collation.py
+++ b/dbt-snowflake/tests/functional/adapter/incremental/test_incremental_collation.py
@@ -144,6 +144,15 @@ select * from {{ source('test', 'stg_has_collation') }}
 
 
 class TestIncrementalCollationPreservedOnSchemaChange:
+    """Test that collation is preserved during incremental runs with schema changes.
+    Tests the following scenarios:
+    - Incremental run with schema change on a column with no collation to a column with collation
+    - Incremental run with schema change on a column with different collation than source
+    - Incremental run with schema change on a column with collation to a column with no collation
+    To do this we simulate a schema change by creating the tables ahead of time and specifying the collation on the source / target tables.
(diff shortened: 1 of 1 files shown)


dbt-snowflake/src/dbt/adapters/snowflake/column.py
```diff
@@ -56,6 +56,31 @@ def expanded_data_type(self) -> str:
...
+        if not self.is_string() or not other_column.is_string():
+            return False
+
+        other_collation = getattr(other_column, "collation", None)
+
+        # If the target column already has a collation but the source column doesn't,
+        # we should not consider this a type change that warrants dropping collation.
+        # Treat it as compatible so schema change logic won't emit an ALTER COLUMN ... SET DATA TYPE.
+        if other_collation and not self.collation:
+            return True
```
emmyoop: What if this is true and `other_column.string_size() < self.string_size()`, should we still return True?

colin-k-rogers: 🤦 nope, updated the logic, this should be the inverse and or'd with `self.collation and not other_collation`

Issue #1496, opened 2025-12-12
[Feature] Spark 4 support in dbt-spark adapter — roadmap or contribution guidance?


PigNatovsky:
### Is this your first time submitting a feature request?

- [x] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [x] I have searched the existing issues, and I could not find an existing issue for this feature
- [x] I am requesting a straightforward extension of existing dbt functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Hello dbt maintainers!

First of all, thank you for your work on dbt-adapters and the dbt-spark adapter.

I would like to ask whether support for Apache Spark 4 is currently planned for the dbt-spark adapter.

If so:

Is there an existing issue, RFC, or discussion thread that tracks this work and can be followed?

If not:

What would be the recommended next steps to initiate the process of adding Spark 4 support?

Are there specific design considerations, constraints, or prior decisions around Spark version support that contributors should be aware of?

Would starting with a proposal or exploratory issue be the preferred approach?

I’ve reviewed the dbt open-source contribution expectations and understand that maintainers may not be able to commit to timelines or roadmap guarantees. My goal here is primarily to understand the current state and the most appropriate way to collaborate or contribute in line with the project’s processes.

Thanks in advance for any guidance.

### Describe alternatives you've considered

Lack of alternative here.

### Who will this benefit?

Every Spark 4 user :)

### Are you interested in contributing this feature?

Sure!

### Anything else?

Have a nice day!

colin-k-rogers:
Closing — Spark 4 support was added by PR #1537 (merged 2026-01-29).

PR #1537, opened 2026-01-09
Bump pyspark version to support Spark v4.x

thmshmm:
resolves #

#1496

### Problem

pyspark v4.x can not be used with the spark adapter.

### Solution

By increasing the version constraint of pyspark, version 4 can be used as well.

I am currently testing in our environment.

So far I've tested with thrift and session mode. Mainly focusing on session mode.

### Checklist


colin-k-rogers:
Looks like some of our integration tests are failing here with exceptions like:
```
[DEBUG] py4j.clientserver: Answer received: !ys[CAST_INVALID_INPUT] The value 'None' of the type "STRING" cannot be cast to "TIMESTAMP" because it is malformed. Correct the value as per the syntax, or change its target type. Use `try_cast` to tolerate malformed input and return NULL instead. SQLSTATE: 22018\n== SQL (line 3, position 113) ==\n...e),cast('2018-01-01' as date)),(cast('None' as timestamp),cast('None' as date),cast('None...\n                                   ^^^^^^^^^^^^^^^^^^^^^^^^^\n
```

thmshmm:
Looking into it

thmshmm:
@colin-rogers-dbt, I think what we can do is to run the tests with disabled ANSI compliance, which was the default in Spark 3 and changed in Spark 4. V4 added the try_cast to operate similar to what Spark 3 did, return NULL on errors. Changing this now would make it incompatible with v3. Any thought?

colin-k-rogers:
> I think what we can do is to run the tests with disabled ANSI compliance, which was the default in Spark 3 and changed in Spark 4. 

@thmshmm I think we'll need to update application code to set `spark.sql.ansi.enabled` to false for all versions and then do a more extensive follow up to add Behavior Flag gating ansi compliance + update all of our templating logic to be ansi compliant.

thmshmm:
We can ensure the parameter is set in `server_side_parameters` at `__post_init__` of the `SparkCredentials`. The creds are passed to each of the connection methods.

Do you want to split the second part into a dedicated issue then?

colin-k-rogers:
Sorry for the delayed response @thmshmm:
> We can ensure the parameter is set in server_side_parameters at __post_init__ of the SparkCredentials. The creds are passed to each of the connection methods.

I think this is reasonable step, at some point we can make this user configurable with a default of false 

> Do you want to split the second part into a dedicated issue then?
yes, don't need to block this on that

thmshmm:
hey @colin-rogers-dbt, np. updated the code based on what we discussed. could you please trigger the integration tests again.

colin-k-rogers:
> hey @colin-rogers-dbt, np. updated the code based on what we discussed. could you please trigger the integration tests again.

Thanks @thmshmm, will get this merged and released!

PR #1538, opened 2026-01-09
Skip unnecessary transaction statements if autocommit is enabled

colin-k-rogers:
## Summary

This PR adds an optional optimization to skip unnecessary transaction statements when using autocommit mode, gated behind a behavior flag for safe rollout.

### Problem

With `autocommit=True` (the default since dbt-redshift 1.5), dbt was still sending explicit `BEGIN`/`COMMIT`/`ROLLBACK` statements. These are unnecessary when autocommit is enabled since each statement is automatically committed by the driver, adding extra round trips to Redshift.

### Solution

Added a new behavior flag `redshift_skip_autocommit_transaction_statements` that, when enabled, skips sending transaction management statements when autocommit is enabled.

**Key changes:**

1. **Behavior flag** - New flag `redshift_skip_autocommit_transaction_statements` (default: `false`)
   - Defined in `impl.py` with `_behavior_flags` property
   - Passed to ConnectionManager via callback for access in transaction methods

2. **Transaction management optimization** - When the flag is enabled AND autocommit is true:
   - `begin()` skips sending `BEGIN` but still tracks `transaction_open = True`
   - `commit()` skips sending `COMMIT` but still tracks `transaction_open = False`  
   - `rollback_if_open()` skips rollback but resets `transaction_open = False`

3. **Framework state consistency** - The `transaction_open` flag is always properly managed to maintain compatibility with dbt's transaction tracking, even when actual statements are skipped.

#### Testing 
In running against a 1000 model test project it saved 2min: 1.11 takes 20min while this branch takes 18min.

### Usage

Users can enable the optimization in their `dbt_project.yml`:

```yaml
flags:
  redshift_skip_autocommit_transaction_statements: true
```

### Backward Compatibility

- **Default behavior unchanged** - Flag defaults to `false`, so existing behavior is preserved
- **`autocommit=False` unaffected** - Explicit transactions still work as before
- **`autocommit=True` without flag** - Still sends BEGIN/COMMIT (current behavior)
- **`autocommit=True` with flag** - Skips unnecessary transaction statements

## Test plan

- [x] All 75 unit tests pass
- [x] Tests for behavior flag enabled (statements skipped)
- [x] Tests for behavior flag disabled (statements sent even with autocommit)
- [x] Tests for autocommit=False (statements always sent)
- [ ] Functional testing against Redshift cluster with flag enabled
- [ ] Verify no BEGIN/COMMIT in query logs when flag is enabled


colin-k-rogers:
## Follow-up Fix: SVV_ALL_COLUMNS instead of SHOW COLUMNS

After functional testing, discovered that `SHOW COLUMNS FROM TABLE` doesn't work for materialized views. Updated `get_columns_in_relation` macro to use `SVV_ALL_COLUMNS` instead, which works uniformly for:
- Tables
- Views
- Materialized views

Still maintaining the late-binding view fallback using `pg_get_late_binding_view_cols()` since late-binding views are not included in `SVV_ALL_COLUMNS`.

The `describe_materialized_view` macro continues to use `SVV_REDSHIFT_COLUMNS` because it provides the `distkey` and `sortkey` columns needed for materialized view configuration detection.

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -487,6 +517,62 @@ def get_response(cls, cursor: redshift_connector.Cursor) -> AdapterResponse:
...
+
+        When autocommit is enabled and the behavior flag is set, we skip sending
+        the BEGIN statement since each statement is automatically committed.
+        However, we still set transaction_open = True to maintain framework state
+        consistency.
+        """
+        connection = self.get_thread_connection()
+
+        if self._should_skip_transaction_statements():
+            # With autocommit + behavior flag, skip BEGIN but still track transaction state
```
QMalcolm: Glad this comment exists, because I was otherwise I definitely would be confused about why we're not setting the connection as open when we don't skip transaction statements.

NIT: That said, this could be written to not need the comment by just doing
```python
if self._should_skip_transaction_statements():
  super.begin()

connection.transaction_open = True
```
I don't believe there'd be any downside to setting `transaction_open = True` twice. Additionally means there is only one return statement (well technically there'd no longer be any explicit return statement, only the implicit). Overall it's just more a style thing and you can ignore this.

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -487,6 +517,62 @@ def get_response(cls, cursor: redshift_connector.Cursor) -> AdapterResponse:
...
+
+        When autocommit is enabled and the behavior flag is set, we skip sending
+        the COMMIT statement since each statement is automatically committed.
+        However, we still set transaction_open = False to maintain framework state
+        consistency.
+        """
+        connection = self.get_thread_connection()
+
+        if self._should_skip_transaction_statements():
+            # With autocommit + behavior flag, skip COMMIT but still track transaction state
```
QMalcolm: Ditto to this part

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -487,6 +517,62 @@ def get_response(cls, cursor: redshift_connector.Cursor) -> AdapterResponse:
...
+
+        When autocommit is enabled and the behavior flag is set, there's no
+        transaction to rollback since each statement is independently committed.
+        We still reset the transaction_open flag to maintain framework state
+        consistency.
+        """
+        connection = self.get_thread_connection()
+
+        if self._should_skip_transaction_statements():
+            # With autocommit + behavior flag, nothing to rollback but reset transaction state
```
QMalcolm: and ditto to this part

QMalcolm (approved):
Approved! Some style nits, but it's just my preference to reduce return statements, nothing that actually needs to be addressed

wusanny:
Hi @colin-rogers-dbt, I have a quick question regarding the `redshift_skip_autocommit_transaction_statements` flag.

From the PR description here, it looked like the default was intended to be `False`, with users opting into the optimization by enabling it in their dbt_project.yml.

Screenshot 2026-03-05 at 7 20 15 pm

However, based on what I’m seeing in the changes here:
https://github.com/dbt-labs/dbt-adapters/pull/1538/changes#diff-a2a74d0a1b2eb523bb95724b8913d20d76c4332a92da87a63294e12cd8a2fbc9R34

and the docs here:
https://docs.getdbt.com/reference/global-configs/redshift-changes#redshift_skip_autocommit_transaction_statements-flag

it looks like the default may now be `True`, meaning users who want to preserve the legacy behavior would need to explicitly set the flag to `False`.

Just wanted to double check that I’m interpreting this correctly. Would you mind clarifying?

Thanks in advance!

barthon-b:
@colin-rogers-dbt, Wusanny's concern seems to be valid. Would you be able to comment?

image


The concern here is that preventing the transaction control statements from being emitted may break the previous assumption that hooks (by default) are in a single transaction with the models. Even autocommit=True mode I believe does not commit if a transaction has been opened explicitly, like the one around the hooks and models. This is not documented clearly on the flag's docs page.

PR #1558, opened 2026-01-17
Optimize Athena partition deletions for insert_overwrite strategy

### LINKED ISSUE #1125: [Feature] Optimize Athena insert overwrite deletions by using batch operations

### Describe the feature

Insert overwrite of dbt-athena deletes overlapping partitions one by one. This includes all these API calls that are made for a single entry at a time:
* Get partition metadata (to get the s3 location)
* List and delete files under the partition s3 prefix
* Drop partition

All of these can be migrated to batch operations, reducing overhead caused by individual API calls.

**In my practical test that had 450 partitions to delete, the total duration for the deletion of overlapping partitions was reduced _from 11 minutes to 1 minute 30 seconds_**.

I don't see any clean alternative, apart from using something else than `table_type=hive` with `incremental_strategy=insert_overwrite`.

### Who will this benefit?

Anyone that is writing more than a few partitions per run.

Yes, I have a working prototype

### Anything else?

First of all, this loop (which is called as part of insert ovewrite):

https://github.com/dbt-labs/dbt-adapters/blob/4b3fc1e9c11a6b454672e02bca5af186dc5caf96/dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/helpers.sql#L116-L118

can be changed to:

```sql
  {%- do adapter.clean_up_partitions(target_relation, partitions) -%}
```

and then this function singature:

https://github.com/dbt-labs/dbt-adapters/blob/4b3fc1e9c11a6b454672e02bca5af186dc5caf96/dbt-athena/src/dbt/adapters/athena/impl.py#L404

is respectively changed to:

```python
    def clean_up_partitions(self, relation: AthenaRelation, where_conditions: List[str]) -> None:
```

Then the `clean_up_partitions` function can be optimised as follows::
* `get_partitions` is called in chunks with multiple partition conditions with expression max length 2048 characters
  * In my test case with a single partitioned column one call could fit 66 partitions.
* `objects.filter(Prefix=prefix).delete()` can be changed to:
  1. calling `s3_bucket.objects.filter(Prefix=prefix)` for each prefix
  2. calling `delete_objects` for the listed s3 objects across all partitions in chunks of max 1000 objects
* `delete_partition` can be replaced with `batch_delete_partition` in chunks of 25 partitions

Links:
* https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-multiple-objects.html
* https://docs.aws.amazon.com/glue/latest/webapi/API_BatchDeletePartition.html#API_BatchDeletePartition_RequestSyntax

juhoautio-rovio:
resolves #1125

docs: "N/A"

## Problem

The dbt-athena adapter currently deletes metadata and files of overlapping partitions by making individual API calls for each partition.

This naive approach results in extraneous API calls for incremental models that process more than a few partitions. For example, a model with 150 partitions requires 300+ API calls (150+ to Glue, 150+ to S3), causing partition deletion to dominate the total execution time. In real-world testing, `clean_up_partitions` consumed 13.5 minutes out of a 15-minute total runtime, making partition deletion the primary performance bottleneck.

## Solution

Significantly improves performance of `insert_overwrite` incremental materializations by optimizing S3 and Glue API usage when deleting partitions.

### Key Changes

1. Batch S3 deletions
   - New `bulk_delete_from_s3()` method batches up to 1000 objects per API call
2. Batch Glue partition deletions
   - Uses `batch_delete_partition` API (25 partitions per call)

### Performance Impact

API call reduction - for example, if there are 150 partitions to handle:
- Before: 150 S3 calls + 150 Glue calls = 300 API calls
- After: ~1 S3 call + ~6 Glue calls = ~7 API calls (~98% reduction)

Real-world execution time (150-partition incremental model):
- Before: 13m 30s in clean_up_partitions
- After: 1m 33s in clean_up_partitions
- Result: 10x faster partition deletion

### Testing

- Unit test with 1100 partitions verifying batch processing logic
- Integration test with 150 partitions verifying expression chunking

#### Manual validation

Tested in a dbt project by patching dbt-athena adapter in the project venv and running:

```shell
dbt run -s my_daily_incremental_model --vars '{"start_date": "2024-01-10", "end_date": "2025-05-11"}'
```

- Before patching: total 15m 0s, clean_up_partitions 13 m 30s
- After: total 3m 33s, clean_up_partitions 1m 33s

In other words, the `clean_up_partitions` part was ~10x faster.

#### Question about a slow integration test

The new `test_partition_chunking.py` takes approximately 2+ minutes to run. Is that too much?

Reason for the long duration is creating and processing 150 partitions with real AWS API calls. The Glue GetPartitions API has a 2048 character limit for the Expression parameter. This test creates enough partitions to trigger code path that splits to have more than a single chunk. There's no way to trigger that scenario without creating many partitions or modifying the runtime code.

Some options to consider:
1. Keep the test as is for maximum confidence, despite the long runtime.
2. Patch the constant `GET_PARTITIONS_API_EXPRESSION_MAX_LENGTH` (default 2048) to use a smaller value during the test, so that a minimal number of partitions are needed to trigger chunking.
3. Make the `GET_PARTITIONS_API_EXPRESSION_MAX_LENGTH` configurable and set a smaller value during the test. This would not be needed in actual usage though, so it would be adding noise to the set of available configurations.
4. Remove this test. Cover the chunking logic only in unit test instead of an integration test.

### Future Improvements

Consider parallelizing S3 API calls for additional performance gains.

## Checklist

   - The `clean_up_partitions` signature change from `str` to `Union[str, List[str]]` was specified by @colin-rogers-dbt in https://github.com/dbt-labs/dbt-adapters/issues/1125#issuecomment-3181125230
   - New `bulk_delete_from_s3()` method is marked `@available` for potential future use; the annotation can be removed if not desired for the public API


juhoautio-rovio:
Addressed the copilot review comments (with relevant test cases added).

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -415,24 +434,79 @@ def clean_up_partitions(self, relation: AthenaRelation, where_condition: str) ->
...
+            for glue_batch in get_chunks(
+                partitions, AthenaAdapter.BATCH_DELETE_PARTITION_API_LIMIT
+            ):
+                response = glue_client.batch_delete_partition(
+                    CatalogId=catalog_id,
+                    DatabaseName=relation.schema,
+                    TableName=relation.identifier,
+                    PartitionsToDelete=[{"Values": p["Values"]} for p in glue_batch],
+                )
+                if response.get("Errors"):
```
colin-k-rogers: Are there any retry-able errors? Thinking of concurrency/networking issues where it would make sense to just try again

colin-k-rogers: also nit but I think you can simplify this code a bit with a walrus operator
```suggestion
                if errors := response.get("Errors"):
```

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -538,6 +612,69 @@ def delete_from_s3(self, s3_path: str) -> None:
         else:
             LOGGER.debug("S3 path does not exist")
 
+    @available
```
colin-k-rogers: does this need to be `available`? (i.e. do we think this needs to be exposed in the jinja context?)

juhoautio-rovio: It doesn't need to. I had this question in the PR description:

> New `bulk_delete_from_s3()` method is marked `@available` for potential future use; the annotation can be removed if not desired for the public API

I can remove the annotation.

Edit: done in https://github.com/dbt-labs/dbt-adapters/pull/1558/commits/7438e47de5963c2e2108c2eb85e9f234762c22b1

dbt-athena/src/dbt/adapters/athena/impl.py
juhoautio-rovio: > Are there any retry-able errors? Thinking of concurrency/networking issues where it would make sense to just try again

The `glue_client` is already configured with botocore's standard retry mode via `get_boto3_config(num_retries=creds.effective_num_retries)` in `dbt-athena/src/dbt/adapters/athena/config.py`, which automatically retries on transient errors (throttling, networking, 5xx) up to the user-configured `num_retries` attempts. So concurrency/networking issues are already covered transparently at the client level.

juhoautio-rovio: > also nit but I think you can simplify this code a bit with a walrus operator

-> https://github.com/dbt-labs/dbt-adapters/pull/1558/commits/0cac99866c4032f259dde1402516512ebbbdf514

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -538,6 +612,69 @@ def delete_from_s3(self, s3_path: str) -> None:
...
+                    )
+                    if "Errors" in response:
+                        for err in response["Errors"]:
+                            LOGGER.error(
+                                f"Failed to delete S3 object: Key='{err['Key']}', "
+                                f"Code='{err['Code']}', Message='{err['Message']}', "
+                                f"Bucket='{bucket_name}'"
+                            )
+                        raise DbtRuntimeError(
+                            f"Failed to delete {len(response['Errors'])} object(s) from S3 bucket '{bucket_name}'"
```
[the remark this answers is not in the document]
juhoautio-rovio: -> https://github.com/dbt-labs/dbt-adapters/pull/1558/commits/bf5658ff81307c860faa60652cbdc0455b3b5597

dbt-athena/tests/unit/test_adapter.py
```diff
@@ -386,6 +417,224 @@ def test_clean_up_partitions_will_work(self, dbt_debug_caplog, mock_aws_service)
...
+        # Verify all S3 objects were deleted
+        objs = s3.list_objects_v2(Bucket=BUCKET)
+        assert objs.get("KeyCount", 0) == 0
+
+        # Verify all Glue partitions were deleted
+        partitions_after = glue.get_partitions(
+            DatabaseName=DATABASE_NAME, TableName=table_name
+        ).get("Partitions", [])
+        assert len(partitions_after) == 0
+
```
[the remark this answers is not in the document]
juhoautio-rovio: I included a question about this exact point in the PR description – the section title is **Question about a slow integration test**.

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -415,24 +434,79 @@ def clean_up_partitions(self, relation: AthenaRelation, where_condition: str) ->
...
+                ):
+                    raise DbtRuntimeError(
+                        f"Partition condition exceeds the Glue API expression limit of "
+                        f"{AthenaAdapter.GET_PARTITIONS_API_EXPRESSION_MAX_LENGTH} characters: "
+                        f"'{condition_with_brackets[:100]}...'"
+                    )
+                if current_chunk and (
+                    len(join_or_conditions(current_chunk + [condition_with_brackets]))
+                    > AthenaAdapter.GET_PARTITIONS_API_EXPRESSION_MAX_LENGTH
+                ):
```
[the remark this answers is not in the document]
juhoautio-rovio: In practice `where_conditions` is a list of partition filter conditions (e.g. `dt = '2022-01-01'`). These are already pre-chunked from the caller, and the number of conditions per chunk is bounded by the 2048-char Glue limit — so at most a handful of conditions fit in one chunk. The "quadratic" behavior is real in theory but completely negligible here. I'd say it's not worth fixing.

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -415,24 +434,79 @@ def clean_up_partitions(self, relation: AthenaRelation, where_condition: str) ->
...
+                expression = join_or_conditions(expression_chunk)
+                partition_params = {
+                    "CatalogId": catalog_id,
+                    "DatabaseName": relation.schema,
+                    "TableName": relation.identifier,
+                    "Expression": expression,
+                    "ExcludeColumnSchema": True,
+                }
+                for page in paginator.paginate(**partition_params):
+                    yield from page.get("Partitions", [])
```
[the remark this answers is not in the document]
juhoautio-rovio: -> https://github.com/dbt-labs/dbt-adapters/pull/1558/commits/4bdf91f351b273c34555be6090a4df651ab720f8

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -415,24 +434,79 @@ def clean_up_partitions(self, relation: AthenaRelation, where_condition: str) ->
...
+
+        def delete_partition_chunk(partitions):
+            self.bulk_delete_from_s3([p["StorageDescriptor"]["Location"] for p in partitions])
```
colin-k-rogers, line 483: Should we move the bulk_delete_from_s3 until _after_ the glue partitions are deleted? Just concerned that if we fail to delete the glue partition after deleting the s3 data that leaves the catalog in an incorrect state.

colin-k-rogers (commented):
On testing options: I don't think we need a specific integration test for chunking, ideally this is validated as part of a test of `insert_overwrite`

One question on the order of operations here but this looks just about good to ship

dbt-athena/src/dbt/adapters/athena/impl.py
juhoautio-rovio: The ordering is pre-existing behavior from `main` — this PR doesn't change it.

This PR only changes efficiency (batching), not functional guarantees.

The current strategy derives S3 paths to delete from `partition["StorageDescriptor"]["Location"]` in Glue partition metadata. Then new files are written with `INSERT INTO` (Athena SQL, which appends to the target path — it doesn't replace).

Simply flipping to Glue-first ordering would make things worse: if the job is interrupted after Glue deletion but before S3 deletion, the S3 data is permanently orphaned since a retry won't find the partition in Glue anymore.

Actually this makes us realize that the taken approach even has a correctness issue that can happen if a job fails in the middle of its work when writing (after the cleanup phase): if a previous run wrote to S3 but failed before registering the partition in Glue, the next run finds nothing in Glue and silently skips S3 cleanup — leaving stale files mixed with new ones, causing data duplication as the end state.

This PR is not changing the strategy now, to not alter existing behavior. Fixing the consistency issues properly would be a bigger change.

A proper fix for that could be like this:
1. Drop the Glue partition metadata first (assuming this is desired for "consistency")
2. Delete the S3 locations from that metadata as best effort (if the job fails here some stale data may remain)
4. Delete from S3 based on the partition paths determined for the current write — not relying on partition metadata

juhoautio-rovio:
> On testing options: I don't think we need a specific integration test for chunking, ideally this is validated as part of a test of insert_overwrite

Did you mean modifying some of the existing integration tests (tests/functional/adapter/something) or just making sure that unit tests have the desired coverage?

If you would still like to cover this in an integration test (some existing one, right?), then would you like to optimize the speed by mocking (as proposed in PR desc)? I suppose an extra 2 minutes is too much for this :)

juhoautio-rovio:
I checked for different options for testing. Maybe this could help with the decision:

**What the new functional test covers that unit tests don't:** Unit tests mock AWS responses and verify chunking logic in isolation. The functional test is the only place that verifies chunking integrates correctly
end-to-end — that after real Glue API calls with expression chunking, the resulting data has no duplicates or missing rows.

**Existing insert_overwrite functional tests:** `test_unique_tmp_table_suffix.py` is the closest match — it uses `insert_overwrite` with `partitioned_by=['date_column']` and runs incrementally multiple times, but
 only writes one partition per run (a single date value). `test_partitions.py` has 212 records across many partitions but never does an incremental insert_overwrite run. None of the existing tests exercise the
expression-length chunking path.

**Folding chunking into the existing test:** `GET_PARTITIONS_API_EXPRESSION_MAX_LENGTH = 2048` is a class constant on `AthenaAdapter`. Monkeypatching it to e.g. 60 would force chunking with
just 3–4 partitions (each `(date_column='2023-01-01')` expression is ~28 chars). `test_unique_tmp_table_suffix.py` could be extended to write a few dates instead of one, monkeypatch the limit, and assert no
duplicates — covering the chunking path without 150 partitions. The test would become marginally slower: Athena query startup and result polling dominate the runtime, so a handful of extra partitions adds
negligible time compared to the current `test_partition_chunking.py` which takes 2+ minutes purely due to creating 150 partitions. The same optimization for test duration can be of course applied also in the `test_partition_chunking.py` itself, if you would like to keep these tests more isolated and case specific.

dbt-athena/src/dbt/adapters/athena/impl.py
juhoautio-rovio: Reproducing the correctness issue based on current main branch: https://github.com/dbt-labs/dbt-adapters/pull/1895

colin-k-rogers: fair enough, definitely out of scope for this PR

colin-k-rogers:
>Folding chunking into the existing test: GET_PARTITIONS_API_EXPRESSION_MAX_LENGTH = 2048 is a class constant on AthenaAdapter. Monkeypatching it to e.g. 60 would force chunking with
just 3–4 partitions (each (date_column='2023-01-01') expression is ~28 chars). test_unique_tmp_table_suffix.py could be extended to write a few dates instead of one, monkeypatch the limit, and assert no

@juhoautio-rovio if we can do this that would be great

colin-k-rogers (approved):

PR #1590, opened 2026-01-31
not use token

ChenyuLInx:
just submodule without token


ChenyuLInx:
https://github.com/dbt-labs/dbt-adapters/actions/runs/21535815216/job/62061196453
works, merging

PR #1594, opened 2026-02-04
Configure notebook execution job security labels

jialuoo:
When running dbt-bigframes job using cross-project SA with impersonation, vertex AI enforces a security boundary to prevent unauthorized data access and the API triggers a safety block.

To bypass this, the `notebook_runtime_out_of_org_warning` label should be set to `ack`. This serves as a programmatic acknowledgement of the security configuration, allowing the job to proceed. Or it will get a bad request error in cross-project SA scenario. 

It won't affect any current workflows.

resolves #

### Checklist

### DIFF
--- a/dbt-bigquery/.changes/unreleased/Features-20260204-010759.yaml
+++ b/dbt-bigquery/.changes/unreleased/Features-20260204-010759.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Configure notebook execution job security labels
+body: Configure notebook execution job security labels to support cross-project runtime templates
 time: 2026-02-04T01:07:59.265166919Z
 custom:
     Author: jialuoo


colin-k-rogers:
Thanks, can we add a unit test and update our functional tests to include this new configuration?

jialuoo:
> Thanks, can we add a unit test and update our functional tests to include this new configuration?

I previously used dbt-bigquery/tests/functional/adapter/test_python_model.py for dbt-bigframes testing, but those tests are designed for single-project environments. Currently, this framework doesn't support the cross-project scenarios. I can only test the functionalities locally and manually (already done that).

aahel:
> > Thanks, can we add a unit test and update our functional tests to include this new configuration?
> 
> I previously used dbt-bigquery/tests/functional/adapter/test_python_model.py for dbt-bigframes testing, but those tests are designed for single-project environments. Currently, this framework doesn't support the cross-project scenarios. I can only test the functionalities locally and manually (already done that).

Can we add a unit test ? 
Also if we are not adding functional tests can we add some code comments so we remember that this is not covered by functional testing.

jialuoo:
> framework doesn't support the cross-project scenarios

I added a TODO since current framework doesn't support tests for the cross-project scenarios. But it's a good idea to support that to increase test coverage.

aahel:
> > framework doesn't support the cross-project scenarios
> 
> I added a TODO since current framework doesn't support tests for the cross-project scenarios. But it's a good idea to support that to increase test coverage.

Thanks, We should also add the unit test to assert this label is getting set

jialuoo:
> > > framework doesn't support the cross-project scenarios
> > 
> > 
> > I added a TODO since current framework doesn't support tests for the cross-project scenarios. But it's a good idea to support that to increase test coverage.
> 
> Thanks, We should also add the unit test to assert this label is getting set

There currently isn't a unit test file for python_submissions.py to cover all components. In discussions with Mike and Stephen, we noted that the existing Python (Dataproc) model also lacks unit tests,. So we rely on extensive manual testing prior to check-in. Our existing functional tests—found in dbt-bigquery/tests/functional/adapter/test_python_model.py—are currently limited to single-project scenarios. While I agree we should have unit tests for all components, which requires some thoughts and is likely out of scope for this PR.

By the way, I have very limited permissions for this repository so far and am unable to assign reviewers to the PR. Could you assist with that? Also, please apply the necessary label to trigger the functional tests, as they are currently blocked. Thanks.

dbt-bigquery/.changes/unreleased/Features-20260204-010759.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Configure notebook execution job security labels
```
colin-k-rogers:
```suggestion
body: Configure notebook execution job security labels to support cross-project runtime templates
```

dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
```diff
@@ -332,6 +332,15 @@ def _config_notebook_job(
             aiplatform_v1.NotebookExecutionJob.GcsNotebookSource(uri=self._gcs_path)
         )
 
+        # Explicit acknowledge the 'out-of-org' warning required by Vertex AI.
+        # This label is needed for cross-project runtime templates to bypass GCP
+        # security policy blocks that would otherwise prevent job creation.
+        # TODO: add a function test for it.
+        security_ack_label = {
```
colin-k-rogers, line 339: Do we think there are users who want to ensure/enforce that we don't reference templates across projects?

jialuoo: That’s a fair point. To date, we haven’t received any user feedback or requests for that. 

This label is required to bypass safety blocks triggered when a user explicitly provides a Service Account from another project. Since specifying a cross-project SA is an intentional configuration choice, I think we can programmatically provide this acknowledgment to ensure the job proceeds. For users operating within a single project, applying the label is just redundant but carries no risk. 

Given that the current configuration is already complex, I suggest we avoid adding another parameter to toggle this acknowledgment unless a specific need for it arises in the future.

Issue #716, opened 2024-06-17
[Feature] transient dynamic tables


homerunner-mqp:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt-snowflake functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Could we implement the "transient" table config to dynamic tables? Follows same pattern as for tables.
https://docs.snowflake.com/en/sql-reference/sql/create-dynamic-table

### Describe alternatives you've considered

None.

### Who will this benefit?

For cases of dynamic tables as output from a data warehouse, it could decrease cost.

### Are you interested in contributing this feature?

Possibly

wahlrkr:
+1 For this feature. We just noticed that it is possible https://docs.snowflake.com/en/user-guide/dynamic-tables-create#use-transient-dynamic-tables-to-reduce-storage-cost in "raw sql" but the dynamic_table materialisation in dbt does not respect the table `transient`-flag, even though dbt docs have a reference of why `transient` is useful https://docs.getdbt.com/reference/resource-configs/snowflake-configs#transient-tables 🙂

nlaaksonen:
We're heavily relying on dynamic tables and this feature would be very useful for us to save on storage costs.

PR #1603, opened 2026-02-06
feat(snowflake): add transient support for dynamic tables

igorbelianski-cyber:
## Summary

- Add support for the `transient` configuration option on Snowflake dynamic tables
- Dynamic tables can now be created as transient (no fail-safe period) by setting `transient: true` in the model config
- Add behavior flag `snowflake_default_transient_dynamic_tables` to control default behavior (defaults to `False` for backward compatibility)
- Query `SHOW TABLES` to retrieve transient status (not available in `SHOW DYNAMIC TABLES`)
- Detect transient changes and trigger full table recreation (ALTER not supported for transient property)

Resolves: #716 

## Checklist

- [x] Add `transient` field to `SnowflakeDynamicTableConfig`
- [x] Add behavior flag `snowflake_default_transient_dynamic_tables`
- [x] Update `describe_dynamic_table` to query `SHOW TABLES` for transient status
- [x] Add `SnowflakeDynamicTableTransientConfigChange` for change detection
- [x] Update CREATE and REPLACE macros to include `transient` keyword
- [x] Add unit tests for transient configuration
- [x] Add functional tests for transient behavior and behavior flag

## Test plan

- [x] Unit tests pass (`tests/unit/test_dynamic_table_config.py`)
- [x] Functional tests pass (`tests/functional/relation_tests/dynamic_table_tests/`)
- [x] Test creating transient dynamic table with `transient: true`
- [x] Test creating non-transient dynamic table with `transient: false`
- [x] Test changing transient status triggers full refresh
- [x] Test behavior flag controls default when `transient` not specified


dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -550,6 +566,26 @@ def describe_dynamic_table(self, relation: SnowflakeRelation) -> Dict[str, Any]:
 
         return {"dynamic_table": dt_table.select(base_columns)}
 
+    @available
```
colin-k-rogers: While this will definitely work it paints us into a corner: if folks start writing their own custom jinja that relies on this then we're stuck maintaining this very specific method forever. 

Some ideas here:
1. We can make this more generic (i.e `get_show_tables_metadata` or something) and have it return a dict or agate table of the results then the caller can get just the info they need
2. We don't make the available in the jinja context and include it in the `describe_dynamic_table` logic. That might looks like:
```
if relation.config.transient is not None:
     dt_table = dt_table.add_column(
    'transient', 
    dataType=agate.Boolean(), 
    compute=lambda row: self._query_dynamic_table_transient_status(relatgion)
)
```

igorbelianski-cyber: Nice catch i didn't realize this function becomes part of the public interface (definitely undesirable). 
I like the second approach better. It will translate nicely when/if we get TRANSIENT as a part of the DT definition .

PR #1608, opened 2026-02-09
Add autocommit support to Postgres adapter

gmr:
## Summary
- Adds `autocommit` credential field (default `False`) to `PostgresCredentials`, allowing users to opt in via `profiles.yml`
- Sets `handle.autocommit = True` on native psycopg2 connections when enabled, eliminating "idle in transaction" states between statements
- Overrides `begin()`, `commit()`, and `rollback_if_open()` to skip unnecessary transaction statements when autocommit is active
- Adds `postgres_skip_autocommit_transaction_statements` behavior flag as an escape hatch

## Problem
Long-running "idle in transaction" states are a serious operational problem with dbt and PostgreSQL. By default, psycopg2 wraps every statement in an implicit transaction that must be explicitly committed. Between statements, the connection sits in an "idle in transaction" state, which:

- **Holds locks** that block other queries and DDL operations
- **Prevents vacuum and autovacuum** from reclaiming dead tuples, leading to table bloat
- **Consumes connection slots** with connections that are doing no useful work
- **Can trigger `idle_in_transaction_session_timeout`**, killing dbt runs unexpectedly

In production environments with frequent dbt runs, this causes cascading problems: tables bloat, queries slow down, and connections pile up. We've experienced significant operational issues from this behavior at scale, where long-running dbt models hold transactions open for minutes between statements, blocking critical maintenance operations and degrading database performance across the cluster.

Enabling psycopg2's `ISOLATION_LEVEL_AUTOCOMMIT` mode eliminates this entirely by committing each statement immediately.

## Solution
Follows the same pattern established by the Redshift adapter in b9ebd240 (`Skip unnecessary transaction statements if autocommit is enabled (#1538)`). When `autocommit: true` is set in `profiles.yml`, psycopg2's autocommit mode is enabled, which commits each statement immediately and avoids the "idle in transaction" state.

Key design decisions:
- **Default `False`**: Preserves existing behavior — no change for current users. Users opt in by setting `autocommit: true` in `profiles.yml`
- **Behavior flag escape hatch**: `postgres_skip_autocommit_transaction_statements` (default `True`) lets users disable transaction skipping without disabling autocommit itself
- **Record/replay safe**: autocommit is set on the native psycopg2 handle before the record/replay wrapper
- **`close()` path is safe**: `handle.rollback()` on an autocommit psycopg2 connection is a no-op

## Test plan
- [x] All existing Postgres unit tests pass
- [x] New `TestAutocommitBehavior` tests cover:
  - `begin()` / `commit()` / `rollback_if_open()` are no-ops with autocommit + behavior flag
  - `begin()` / `commit()` send statements when autocommit is off
  - `begin()` / `commit()` send statements when autocommit is on but behavior flag is off
  - autocommit is set on the psycopg2 connection handle
  - autocommit defaults to `False`

### DIFF
--- a/dbt-postgres/.changes/unreleased/Features-20260209-212600.yaml
+++ b/dbt-postgres/.changes/unreleased/Features-20260209-212600.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add autocommit support to eliminate idle-in-transaction states
+time: 2026-02-09T21:26:00.000000-05:00
+custom:
+  Author: gmr
+  Issue: "1608"
--- a/dbt-postgres/tests/unit/test_connection.py
+++ b/dbt-postgres/tests/unit/test_connection.py
@@ -386,16 +386,18 @@ class TestAutocommitBehavior(TestCase):
 
         connection = MagicMock()
         connection.state = "closed"
-        connection.credentials = PostgresCredentials.from_dict({
-            "type": "postgres",
-            "dbname": "postgres",
-            "user": "root",
-            "host": "localhost",
-            "pass": "password",
-            "port": 5432,
-            "schema": "public",
-            "autocommit": True,
-        })
+        connection.credentials = PostgresCredentials.from_dict(
+            {
+                "type": "postgres",
+                "dbname": "postgres",
+                "user": "root",
+                "host": "localhost",
+                "pass": "password",
+                "port": 5432,
+                "schema": "public",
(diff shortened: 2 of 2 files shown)


gmr:
CLA submitted

colin-k-rogers:
@cla-bot check

colin-rogers-dbt (approved):
Great idea, thanks!

PR #1632, opened 2026-02-11
Adding friendlier error message when dist is not string for redshift

### LINKED ISSUE #651: [CT-1519] [Bug] Better error message for misspecified dist key yaml config
### Is this a new bug in dbt-redshift?

- [X] I believe this is a new bug in dbt-redshift

### Current Behavior

If you set the dist key in a `model.yml` file like

```yaml
version: 2
models:
  - name: a_real_table
    config:
      sort:
        - primary_key
        - foreign_key
      dist:
        - primary_key
```

dbt will throw the following error

`list has no attribute strip`

Which is generic and non descriptive of what the user should fix. The issue is that since dist keys can only be single valud properties it should be specified as `dist: primary_key`.

### Expected Behavior

A better message that isn't exposing the underlying python machinery and indicates what the user should do to fix the issue e.g. Single valued properties cannot be specified as a list

### Steps To Reproduce

See Current Behaviour. If you set up a model like that in any dbt project you will reproduce the error

### Relevant log output

_No response_

### Environment

```markdown
- OS: Mac
- Python: 3.9
- dbt-core:
- dbt-redshift: 1.3.0
```

### Additional Context

_No response_

tauhid621:
resolves #651 

### Problem

This PR adds a type check to the redshift adapters.sql `dist` macro and raises a better error if the user tries to set the single valued property as non string. The previous error was not very helpful: `list object' has no attribute 'strip'`

A [fix](https://github.com/dbt-labs/dbt-redshift/pull/226) was earlier created by @dluftspring in the archived `dbt-redshift` repo. This PR is similar and checks for non string value instead of just list.

### Solution

`dist` is expected to be a string. We check if `dist` is not string and then raise an error message with details. 

### Checklist

#### Before
Screenshot 2026-02-11 at 12 37 11 PM

#### After
Screenshot 2026-02-11 at 12 46 40 PM

### DIFF
--- a/dbt-redshift/.changes/unreleased/Fixes-20260211-110926.yaml
+++ b/dbt-redshift/.changes/unreleased/Fixes-20260211-110926.yaml
@@ -2,5 +2,5 @@ kind: Fixes
 body: Adding friendlier error message when dist is not string for redshift
 time: 2026-02-11T11:09:26.802148+05:30
 custom:
-    Author: tauhid621
+    Author: tauhid621 dluftspring
     Issue: "651"


dbt-redshift/.changes/unreleased/Fixes-20260211-110926.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Adding friendlier error message when dist is not string for redshift
+time: 2026-02-11T11:09:26.802148+05:30
+custom:
+    Author: tauhid621
```
colin-k-rogers:
```suggestion
    Author: tauhid621 dluftspring
```

PR #1635, opened 2026-02-11
Redshift: Use native alter column type for Varchar size change

### LINKED ISSUE #649: [CT-2166] [Feature] Use ALTER TABLE to expand column / field length

- [X] I am requesting a straightforward extension of existing dbt-redshift functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

When there is a change in a field type and the new field size is "bigger" than the previous, DBT should ALTER the column not recreate on a new column.

ALTER TABLE schema.table ALTER COLUMN field TYPE varchar(1024);
instead of creating a new field, copying the data, renaming, and dropping the old column.
This is a notably better approach on Redshift because of the overhead it generates and the time spent later to run VACUUM on a big table with an update on all rows.

It includes another step, but would do the same, when possible when reducing field size:
- Check if the old values fit on the new size (Select max(len(field)) from schema.table;)
- If it fits, then ALTER TABLE
- if not, then ERROR

### Who will this benefit?

Redshift users who have changes to their schema

_No response_

### Anything else?

_No response_

tauhid621:
resolves #649 

### Problem

When a dbt model’s column type changes (e.g. varchar(100) → varchar(1024)), dbt applies that change to existing tables. The dbt-redshift adapter did not override alter_column_type, so it always used the default implementation: add a temporary column with the new type, copy data, drop the old column (cascade), and rename the new one.

On Redshift that approach is expensive: it rewrites the whole table, increases storage during the change, and can trigger long VACUUMs on large tables. 
Redshift supports in-place size changes for VARCHAR and VARBYTE, which is metadata-only for expanding length and avoids the add/copy/drop/rename overhead.

### Solution
use Redshift’s native ALTER COLUMN TYPE when the type is `VARCHAR`, `CHARACTER VARYING`, or `VARBYTE` (column size changes only). For any other type change, the macro falls back to the default add-temp-column → copy → drop → rename flow.
Handling `dist` and `sort` key changes are out of scope and would still error at runtime(existing behaviour).

**Something strange:** Decreasing the column size was not recognised by dbt for some reason.

### Checklist

#### Before
Screenshot 2026-02-12 at 12 51 14 PM

#### After -> VARCHAR
Screenshot 2026-02-12 at 12 55 52 PM

#### After -> non varchar(old flow)
Screenshot 2026-02-12 at 1 04 33 PM

### DIFF
--- a/dbt-redshift/.changes/unreleased/Features-20260226-172322.yaml
+++ b/dbt-redshift/.changes/unreleased/Features-20260226-172322.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Migrate get_show_grant_sql to use SHOW GRANTS ON TABLE
+time: 2026-02-26T17:23:22.316316+05:30
+custom:
+    Author: tauhid621
+    Issue: "1677"
--- a/dbt-redshift/src/dbt/adapters/redshift/impl.py
+++ b/dbt-redshift/src/dbt/adapters/redshift/impl.py
@@ -189,6 +189,29 @@ class RedshiftAdapter(SQLAdapter):
             column_types=[agate.Text(), agate.Text(), agate.Text(), agate.Text()],
         )
 
+    def standardize_grants_dict(self, grants_table: "agate.Table") -> dict:
+        """Translate the result of `show grants` to match the grants config format.
+
+        When ``redshift_use_show_apis`` is enabled, ``SHOW GRANTS ON TABLE``
+        returns columns ``identity_name`` and ``privilege_type`` (uppercase).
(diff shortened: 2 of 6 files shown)


MichelleArk:
Have you seen the tests in https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-redshift/tests/functional/adapter/incremental/test_incremental_on_schema_change.py? I think it'd be an appropriate place to add testing around the new varchar type change behavior.

tauhid621:
@MichelleArk Thanks for pointing out. I have added tests around varchar and int.

PR #1636, opened 2026-02-11
feat(athena): Fix a bucket partitioning error against many partitions

dtaniwaki:
### Problem

When dbt-athena's ICEBERG_TOO_MANY_OPEN_PARTITIONS fallback triggers batch inserts with bucket partitioning, the
get_partition_batches macro generates WHERE clauses that exceed Athena's 256KB query size limit.

The root cause is that the macro duplicates the full IN (...) clause for every partition expression within each OR
condition:

-- Each OR condition contains the full IN clause → query explodes
```sql
WHERE (date='01-01' AND col IN (...6650 vals...))
    OR (date='01-02' AND col IN (...6650 vals...))
    OR ...
```

With high-cardinality bucket columns (e.g., 850K distinct values across 128 buckets), each bucket's IN clause is ~160KB,
and duplicating it across hundreds of partitions quickly exceeds the limit.

### Solution

Restructure the batch generation from OR-of-ANDs to AND-of-ORs, eliminating IN clause duplication:

-- IN clause appears only once per batch
```sql
WHERE (date='01-01' OR date='01-02' OR ... OR date='01-100')
  AND col IN (...chunk_of_values...)
```

Specific changes to get_partition_batches.sql:

- Partition batching: Group non-bucket partition conditions into batches of athena_partitions_limit (default 100), joined
by OR
- IN clause chunking: Split each bucket's IN clause values into chunks of athena_partitions_limit, consistent with
partition batching. Users can tune partitions_limit to control both partition batch size and IN clause chunk size.
- AND structure: Combine as (partition_batch) AND col IN (chunk) — each final batch contains the IN clause exactly once
- Empty partition filtering: Filter out empty partition expressions (occurs when all partition columns are bucket columns)
  to avoid generating invalid WHERE () AND col IN (...)
- Bucket-only case: When there are no non-bucket partition columns, emit only the bucket IN condition
- Non-bucketed path: Unchanged (batched by athena_partitions_limit as before)

Batch count example (850K distinct values, 128 buckets, 365 dates, partitions_limit=100):

- Each bucket: ~6,640 values → ceil(6640/100) = 67 IN clause chunks
- Partition batches: ceil(365/100) = 4
- Final batches: 128 buckets × 67 chunks × 4 partition batches = 34,304
- Each batch query is small and well within Athena's 256KB limit

<details><summary>Query Examples</summary>

#### Scenario

- Partition columns: date_col, user_id (user_id is the bucket column)
- Date partitions: 100 (2024-01-01 to 2024-04-10)
- Bucket 0 user_id values: 6,640
- athena_partitions_limit: 100

#### Before (single batch query)

```sql
INSERT INTO target (cols)
SELECT cols FROM tmp
WHERE (date_col='2024-01-01' AND user_id IN ('val1', 'val2', ..., 'val6640'))
    OR (date_col='2024-01-02' AND user_id IN ('val1', 'val2', ..., 'val6640'))
    OR ...
    OR (date_col='2024-04-10' AND user_id IN ('val1', 'val2', ..., 'val6640'))
```

The same IN clause (~160KB) is repeated for all 100 partitions, resulting in ~16MB — far exceeding Athena's 256KB query
size limit.

#### After (batched)

Partition conditions and IN clause values are batched independently and combined with AND:

Batch 1:
```sql
INSERT INTO target (cols)
SELECT cols FROM tmp
WHERE (date_col='2024-01-01' OR date_col='2024-01-02' OR ... OR date_col='2024-04-10')
  AND user_id IN ('val1', 'val2', ..., 'val100')
```

Batch 2:
```sql
INSERT INTO target (cols)
SELECT cols FROM tmp
WHERE (date_col='2024-01-01' OR date_col='2024-01-02' OR ... OR date_col='2024-04-10')
  AND user_id IN ('val101', 'val102', ..., 'val200')
```

... (repeats through batch 67)

Each batch query size: partition conditions ~2.5KB + IN clause ~4KB = ~10KB (well within 256KB).

#### Equivalence

By the distributive law of Boolean algebra (A∧C) ∨ (B∧C) = (A∨B) ∧ C, the results are identical before and after. The
bucket hash is determined solely by the user_id value and is independent of the date partition, so the same IN clause
values apply across all partitions.

</details>

### Checklist

### DIFF
--- a/dbt-bigquery/.changes/unreleased/Fixes-20260213-153225.yaml
+++ b/dbt-bigquery/.changes/unreleased/Fixes-20260213-153225.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Change log level for BigQuery job link from debug to info for better visibility.
+time: 2026-02-13T15:32:25.585766+05:30
+custom:
+    Author: aahel
+    Issue: "544"
--- a/dbt-bigquery/src/dbt/adapters/bigquery/connections.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/connections.py
@@ -609,9 +609,7 @@ class BigQueryConnectionManager(BaseConnectionManager):
             and query_job.job_id is not None
             and query_job.project is not None
         ):
-            logger.debug(
-                self._bq_job_link(query_job.location, query_job.project, query_job.job_id)
-            )
+            logger.info(self._bq_job_link(query_job.location, query_job.project, query_job.job_id))
 
         pre = time.perf_counter()
         try:
--- a/dbt-redshift/.changes/unreleased/Fixes-20260213-125020.yaml
+++ b/dbt-redshift/.changes/unreleased/Fixes-20260213-125020.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix dbt clone failing when persist_docs adds column comments to late binding views
+time: 2026-02-13T12:50:20.410997+05:30
+custom:
+    Author: tauhid621
+    Issue: "622"
--- a/dbt-redshift/src/dbt/include/redshift/macros/adapters.sql
+++ b/dbt-redshift/src/dbt/include/redshift/macros/adapters.sql
@@ -272,7 +272,7 @@
   {% endif %}
 
   {# Override: do not set column comments for LBVs #}
-  {% set is_lbv = config.get('materialized') == 'view' and config.get('bind') == false %}
+  {% set is_lbv = relation.type == 'view' and config.get('bind') == false %}
   {% if for_columns and config.persist_column_docs() and model.columns and not is_lbv %}
     {% do run_query(alter_column_comment(relation, model.columns)) %}
   {% endif %}


dbt-athena/tests/unit/test_get_partition_batches.py
```diff
@@ -0,0 +1,313 @@
...
+    {%- endfor -%}
+{%- else -%}
+    {%- set non_empty_partitions = ns_partitions | select | list -%}
+    {%- for i in range(0, non_empty_partitions | length, athena_partitions_limit) -%}
+        {%- set batch = non_empty_partitions[i:i + athena_partitions_limit] -%}
+        {%- do partitions_batches.append(batch | join(' or ')) -%}
+    {%- endfor -%}
+{%- endif -%}
+{{ partitions_batches | tojson }}
+"""
```
iconara: How can we ensure that this won't drift from the actual implementation? This test is only testing the duplicated code here, correct?

Would it be possible to move the implementation to `impl.py` and call out to it from the macro?

iconara (commented):
Not a maintainer, just a concerned party here. I applaud adding test coverage for the macro, but I worry about the code duplication. Is there another way to get the implementation under test than duplicating the code?

dbt-athena/tests/unit/test_get_partition_batches.py
nicor88: I've the same view of @iconara - there is indeed a potential drift issue, in which this functional test could diverge from the actual implementation. Moving part of the logic to a python implementation exposing a function could help but I'm not a bit fan of this approach. 

Did you consider to read the conent `get_partition_batches.sql` in your tests? that could be a sort of win win situation.

dtaniwaki: Thank you for your comment. Let me fix it. 👍

dtaniwaki: Fixed in 4da713c1d0e01282256a5262772b0e4171a6d09b

QMalcolm (changes requested):
Man good thing we have CI 😅 Missed that this is missing the changelog 🫠 Could you please add a changelog 🙈 

If this isn't something you've done before you can do
1. `cd dbt-athena` (you need to be in the dbt-athena directory)
2. `changie new` this will start an interactive changie workflow in your terminal
  a. select `Fixes`
  b. add a message you feel is appropriate
  c. supply your github username
  d. for the issue number, feel free to use the number of this PR `1636`

dtaniwaki:
@QMalcolm 
Thank you for your review! I added a changelog. Would you check the PR again?

nicor88 (approved):
Looks good to me, great work.

PR #1639, opened 2026-02-12
add model level job execution timeout

### LINKED ISSUE #575: [Feature] Make `job_execution_timeout_seconds` configurable at the model level

- [X] I am requesting a straightforward extension of existing dbt-bigquery functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Today, `job_execution_timeout_seconds` is set at the connection level in dbt-bigquery

It would be great to be able to set/overwrite it at the model level as well. Looking at [this code](https://github.com/dbt-labs/dbt-bigquery/blob/de5ff46ca816bb32cc45a5a693a203f3e2d8a47a/dbt/adapters/bigquery/connections.py#L740), it looks that we provide the `timeout` for every SQL call (instead of setting it at the client level).

We could update the [`raw_execute()`](https://github.com/dbt-labs/dbt-bigquery/blob/de5ff46ca816bb32cc45a5a693a203f3e2d8a47a/dbt/adapters/bigquery/connections.py#L446-L452) method to also accept an Optional timeout set at the model level and not at the connection level.

The workaround to have different models with different timeouts today would be to:
- create multiple dbt projects
- or, have the `job_execution_timeout_seconds` depend on some var or env_var and splitting the models with different timeouts in different dbt executions

I can't think of a workaround today where we would allow running models with different timeouts in a single dbt execution.

### Who will this benefit?

People using dbt and BigQuery and having models with large differences in processing time

In that case, some large models can be set with a high timeout and small ones with a smaller timeout.

If need be

### Anything else?

_No response_

aahel:
resolves #575

### Problem

Today, job_execution_timeout_seconds can only be set at the profile/connection level in profiles.yml. Users with models that have vastly different execution times (e.g., 10-second lookups vs 30-minute transformations) must either use a single timeout for all models or split them across separate dbt executions. There is no way to set different timeouts for different models within a single dbt run.

### Solution

This PR makes job_execution_timeout_seconds configurable at the model level, with the model-level value taking precedence over the profile-level default. The implementation follows the same pattern Snowflake uses for per-model warehouse switching via pre_model_hook/post_model_hook.

Changes:

impl.py: Added job_execution_timeout_seconds to BigqueryConfig so dbt recognizes it as a valid model config. Added pre_model_hook and post_model_hook on BigQueryAdapter to set/clear per-model timeout before and after materialization.
connections.py: Added thread-safe per-thread timeout storage on BigQueryConnectionManager (keyed by thread ID, same pattern as jobs_by_thread). Modified _query_and_results() and copy_bq_table() to check for model-level timeout first, falling back to credentials-level timeout. This ensures both SQL execution and partition copying during incremental materializations respect the model-level timeout.

Usage:

-- In a model file:
{{ config(job_execution_timeout_seconds=600) }}
SELECT ...

Or in schema.yml:
models:
  - name: my_long_running_model
    config:
      job_execution_timeout_seconds: 600

### Checklist

### DIFF
--- a/dbt-redshift/.changes/unreleased/Features-20260216-175539.yaml
+++ b/dbt-redshift/.changes/unreleased/Features-20260216-175539.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Adds support for Redshift session `query_group` so dbt runs can be tagged for WLM and query logging
+time: 2026-02-16T17:55:39.941825+05:30
+custom:
+    Author: tauhid621 trouze
+    Issue: "648"
--- a/dbt-redshift/src/dbt/adapters/redshift/connections.py
+++ b/dbt-redshift/src/dbt/adapters/redshift/connections.py
@@ -186,6 +186,9 @@ class RedshiftCredentials(Credentials):
     tcp_keepalive_interval: Optional[int] = None
     tcp_keepalive_count: Optional[int] = None
 
+    # Query group for WLM and query logging (appears in STL_QUERY, SVL_QLOG, etc.)
+    query_group: Optional[str] = None
+
     _ALIASES = {"dbname": "database", "pass": "password"}
 
     @property
@@ -222,6 +225,7 @@ class RedshiftCredentials(Credentials):
             "tcp_keepalive_idle",
             "tcp_keepalive_interval",
             "tcp_keepalive_count",
+            "query_group",
         )
 
     @property
@@ -448,6 +452,9 @@ def get_connection_method(
             c.autocommit = True
         if credentials.role:
             c.cursor().execute(f"set role {credentials.role}")
+        if credentials.query_group:
+            value = str(credentials.query_group).replace("'", "''")
+            c.cursor().execute(f"SET query_group TO '{value}'")
         return c
 
     return connect
--- a/dbt-redshift/src/dbt/adapters/redshift/impl.py
+++ b/dbt-redshift/src/dbt/adapters/redshift/impl.py
@@ -5,7 +5,7 @@ from multiprocessing.context import SpawnContext
 import agate
 from dbt_common.behavior_flags import BehaviorFlag
 from dbt_common.contracts.constraints import ConstraintType
-from typing import List, Optional, Set, Any, Dict, Type
+from typing import List, Optional, Set, Any, Dict, Type, Mapping
 from collections import namedtuple
 from dbt.adapters.base import PythonJobHelper
 from dbt.adapters.base.impl import AdapterConfig, ConstraintSupport
@@ -61,6 +61,7 @@ class RedshiftConfig(AdapterConfig):
     bind: Optional[bool] = None
     backup: Optional[bool] = True
     auto_refresh: Optional[bool] = False
+    query_group: Optional[str] = None
 
 
 class RedshiftAdapter(SQLAdapter):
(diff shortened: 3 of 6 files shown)


dbt-bigquery/src/dbt/adapters/bigquery/connections.py
```diff
@@ -591,7 +607,9 @@ def _query_and_results(
         limit: Optional[int] = None,
     ):
         client: Client = conn.handle
-        timeout = self._retry.create_job_execution_timeout()
+        # Use model-level timeout if set, otherwise fall back to credentials-level
+        model_timeout = self.get_model_timeout()
+        timeout = model_timeout or self._retry.create_job_execution_timeout()
```
colin-k-rogers: I think a simpler approach that's more idiomatic to how we approach this framework would be to put `job_timeout_ms` in the `job_params` and then only set it explicitly on `query_job_config` if it's not been set. 
That would remove the need to have the setter/getter methods on the connection

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -252,6 +254,16 @@ def rename_relation(
         client.copy_table(from_table_ref, to_table_ref)
         client.delete_table(from_table_ref)
 
+    def pre_model_hook(self, config: Mapping[str, Any]) -> Optional[float]:
```
colin-k-rogers, line 268: This is a novel usage of the `pre_model_hook` but it's a little bit counter intuitive since usually a `pre_model_hook` actually makes a call to the target platform.

dbt-bigquery/src/dbt/adapters/bigquery/connections.py
aahel: done

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
colin-k-rogers: Can you add a code comment/doc string to the pre/post function explaining why we we're doing this:
```suggestion
    def pre_model_hook(self, config: Mapping[str, Any]) -> Optional[float]:
    """ Override the connection's query execution timeout based on the model config
    """
```

MichelleArk:
Does this config only apply to models, or does it also work for seeds & snapshots?

aahel:
> Does this config only apply to models, or does it also work for seeds & snapshots?

[pre_model_hook](https://github.com/dbt-labs/dbt-adapters/pull/1639/changes#diff-6e486fc7d7bef61175f198c776b9bbb73919681166a210f270ec002db67778d6R257) is called for other resource types in dbt-core as well, not just models:

Models — ModelRunner.execute() ([run.py:345](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/run.py#L345))
Seeds/Snapshots — SeedRunner/SnapshotRunner extend ModelRunner, inherit the same hook call
Data tests — TestRunner.execute_data_test() ([test.py:145](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/test.py#L145))
Unit tests — TestRunner.execute_unit_test() ([test.py:223](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/test.py#L223))

Models, snapshots, data tests, unit tests, and clones all execute SQL through the same raw_execute → _query_and_results code path, so the timeout behavior is identical for all of these resource types. The only limitation is the seed CSV upload step (_write_file_to_table), which is a separate non-SQL operation that only respects the profile-level timeout currently ( do we need to fix this here ? . we dont have access to conn object there.)

So job_execution_timeout_seconds works as a model-level config for seeds, snapshots, tests etc, not just models. However, the scope of this PR should was intended for model-level config as described in the original issue (#575).

MichelleArk:
That's good to know! I'd expect that to be the case for seeds and snapshots but it's neat that the config would also apply to data/unit tests. As long as there's a valid use case for supporting the execution timeout across more resources (I think there is!) - I don't see the need to special case only support this for models. 

What do you think @graciegoheen ?

graciegoheen:
@MichelleArk I agree - I don't think models are special here (compared to other resources like snapshots and seeds). I think there'd be value in allowing folks to configure a timeout value for those resource types as well, though models are likely the most important given those typically take the longest to execute in the warehouse.

PR #1649, opened 2026-02-16
Adds support for Redshift session `query_group`

### LINKED ISSUE #648: [ADAP-381] Enable tagging of query_group in configs

- [X] I am requesting a straightforward extension of existing dbt-redshift functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

Much akin to dbt-snowflake's implementation of [query tagging](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#query-tags), I think it'd be great to allow users to set a [query_group](https://docs.aws.amazon.com/redshift/latest/dg/r_query_group.html) for their models. I'd imagine the implementation of this would be quite straightforward as the functionality is quite similar to that of Snowflake's, but hoping to get input there.

Not aware that there are any alternatives.

### Who will this benefit?

Anybody who uses the Redshift query logs to perform analysis on their queries.

Certainly! Feels like it could be a good first issue.

### Anything else?

_No response_

tauhid621:
resolves #648

### Problem
Adds support for Redshift session `query_group` so dbt runs can be tagged for WLM and query logging (e.g. `STL_QUERY`, `SVL_QLOG`), similar to Snowflake’s `query_tag`.

### Solution
Solution is similar to `query_tag` [for snowflake](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#query-tags). We support query_group at profile and model level.

- Connection/profile level: We run `SET query_group TO 'query_group'` when the connection opens.
- Model level: For each model materialization, we call a macro which set's a new query group and revert/reset it at the end of the materialization. 

### Checklist

#### project level query group
Screenshot 2026-02-16 at 5 36 27 PM

#### project + model level query_group
Screenshot 2026-02-16 at 5 38 59 PM

### DIFF
--- a/dbt-redshift/src/dbt/adapters/redshift/impl.py
+++ b/dbt-redshift/src/dbt/adapters/redshift/impl.py
@@ -263,20 +263,20 @@ class RedshiftAdapter(SQLAdapter):
 
     def pre_model_hook(self, config: Mapping[str, Any]) -> Optional[str]:
         default_query_group = self.config.credentials.query_group
-        query_group = config.get("query_group", default_query_group)
+        model_query_group = config.get("query_group")
 
-        if query_group == default_query_group or query_group is None:
+        if model_query_group == default_query_group or model_query_group is None:
             return None
-        self._set_query_group(query_group)
+        self._set_query_group(model_query_group)
         return None
 
     def post_model_hook(self, config: Mapping[str, Any], context: Optional[str]) -> None:
         default_query_group = self.config.credentials.query_group
-        query_group = config.get("query_group", default_query_group)
+        model_query_group = config.get("query_group")
 
-        if query_group == default_query_group:
+        if model_query_group == default_query_group:
             return None
-        elif default_query_group is None and query_group is not None:
+        elif default_query_group is None and model_query_group is not None:
             self._unset_query_group()
         else:
             self._set_query_group(default_query_group)


colin-k-rogers:
On reading this I think we should think of this less like Snowflake's `query_tag` feature and more like Snowflake's `warehouse` session parameter. Therefore we should implement this as pre/post model hooks. 
We can crib from that here: https://github.com/dbt-labs/dbt-adapters/blob/5bff73aed42c9a6c43cca4eec1a401525b0f5b98/dbt-snowflake/src/dbt/adapters/snowflake/impl.py#L175

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -639,6 +643,12 @@ def open(cls, connection):
 
         if backend_pid := cls._get_backend_pid(open_connection):
             open_connection.backend_pid = backend_pid
+
+        if getattr(credentials, "query_group", None):
+            value = credentials.query_group.replace("'", "''")
+            with open_connection.handle.cursor() as cursor:
+                cursor.execute(f"SET query_group TO '{value}'")
```
colin-k-rogers: Looks like we're duplicating this query template in a number of places, if we're going with this approach we should centralize that to one macro

tauhid621: I have modified the code to use pre and `post_model_hook`.
Also, moved setting profile level query group at `connect()` where we set the role as well.

tauhid621:
@colin-rogers-dbt Thanks for the suggestion on using pre and post model hooks. I've implemented the same and the solution is much cleaner.

colin-rogers-dbt (approved):

PR #1654, opened 2026-02-17
Fix insert_overwrite with integer range partitions

### LINKED ISSUE #605: [CT-2051] [Bug] New insert_overwrite Bigquery partitioning with integer keys can create huge temporary array variables, exceeding BQ limits
### Describe the bug
When doing an incremental update of an integer-partitioned bigquery table with the new insert_overwrite merge strategy then DBT calculates which partitions it should replace. In the process it can generate a huge BigQuery ARRAY value.

The cause is that DBT does not take the "interval" parameter of the partitioning specification into account. The generated SQL code selects "array_agg(distinct PARTITION_FIELD)" when calculating the partitions to replace. This selects ALL distinct partition field values of the incremental update, even if these values are actually in the same partition. This causes a potentially huge array to be created. If there is enough data in the table then this will even cause an error because BQ's maximum array size (100 MB) is exceeded.

Note that this bug is not triggered by time partitioned tables because for those all partition fields are dates and the partition size is always one day (i.e. there is only one valid value per partition).

### Steps To Reproduce
1) Generate a model with:
```sql
config(
    materialized = 'incremental',
    unique_key = 'partkey',

    partition_by = {
        "field": "partkey",
        "data_type": "int64",
    
        "range": {
            "start": 0,
            "end": 100000,
            "interval": 1000
        }
    },
   incremental_strategy = 'insert_overwrite'
  )
```

2) Run the model once with empty source data

3) Add 1000 records in the source data with values for partkey from 0 to 999

4) Generate the incremental update code for the model and look at the values in the dbt_partitions_for_replacement internal variable.

5) You will see that it contains all 1000 values for partkey from 0 to 999, even though those are all inside the same single partition.

### Expected behavior
DBT should ensure that no huge temporary array variable is generated.
The dbt_partitions_for_replacement array should have at most as many elements as the number of partitions being updated. In my opinion the way to go would be to store only the starting values of each partition in the array and then modify the merge clause to use a range for each partition.

### Screenshots and log output
N/A

### System information
**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [x] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
installed version: 0.16.0
   latest version: 0.16.0

Up to date!
```

**The operating system you're using:**

macOS

**The output of `python --version`:**
Python 3.7.4

aahel:
resolves #605 

### Problem
When using `insert_overwrite` incremental strategy on integer-range-partitioned BigQuery tables with `interval > 1`, dbt generates an excessively large `dbt_partitions_for_replacement` array that can exceed BigQuery's 100MB scripting limit.

The root cause is that `array_agg(distinct PARTITION_FIELD)` collects every distinct field value, even when multiple values belong to the same partition. For example, with `interval=1000` and values 0–999, the array contains 1000 elements instead of just 1 (since they all map to partition [0, 1000)).

This is also a correctness bug: the merge predicate `PARTITION_FIELD in unnest(dbt_partitions_for_replacement)` only deletes rows whose field values exactly match incoming data, rather than deleting all rows in the affected partitions. This leaves stale data behind after an incremental run.

### Solution
Extend `render_wrapped()` in `PartitionConfig` to normalize int64 range partition field values to their partition start boundary using the formula: `value - MOD(value - range_start, range_interval)`.

For time-based partitions, `render_wrapped()` already handled normalization via `timestamp_trunc()`/`date_trunc()`. The int64 range case was simply missing. For all other partition types, behavior is unchanged.

This ensures both the replacement array and the merge predicate consistently use partition-normalized values, so:

- The array contains at most one entry per affected partition (fixes the size limit issue)
- All rows in affected partitions are correctly deleted during overwrite (fixes the stale data issue)

**Before fix** (interval=10, values 1–6 all in partition [0,10)):

| id | id_bucket |
|----|-----------|
|  1 |         1 |  ← stale, not deleted
|  2 |         2 |  ← stale, not deleted
|  3 |         3 |  ← stale, not deleted
| 10 |         4 |
| 20 |         5 |
| 30 |         6 |

**After fix:**

| id | id_bucket |
|----|-----------|
| 10 |         4 |
| 20 |         5 |
| 30 |         6 |

### Checklist

### DIFF
--- a/dbt-bigquery/.changes/unreleased/Fixes-20260213-134345.yaml
+++ b/dbt-bigquery/.changes/unreleased/Fixes-20260213-134345.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix BigFrames notebook execution with OAuth service account credentials by detecting identity from credential object properties
+time: 2026-02-13T13:43:45.687605+05:30
+custom:
+    Author: aahel
+    Issue: "1231"
--- a/dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/python_submissions.py
@@ -27,6 +27,7 @@ from dbt.adapters.bigquery.credentials import (
     create_google_credentials,
     DataprocBatchConfig,
 )
+from google.auth.impersonated_credentials import Credentials as ImpersonatedCredentials
 from dbt.adapters.bigquery.retry import RetryFactory
 from dbt.adapters.events.logging import AdapterLogger
 from dbt_common.exceptions import DbtRuntimeError
@@ -53,8 +54,6 @@ _DEFAULT_BIGFRAMES_TIMEOUT = 60 * 60
 # Time interval in seconds between successive polling attempts to check the
 # notebook job's status in BigFrames mode.
 _COLAB_POLL_INTERVAL = 30
-# Suffix used by service accounts.
-_SERVICE_ACCOUNT_SUFFIX = "iam.gserviceaccount.com"
 
 
 class _BigQueryPythonHelper(PythonJobHelper):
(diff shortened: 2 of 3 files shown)


dbt-bigquery/src/dbt/adapters/bigquery/relation_configs/_partition.py
```diff
@@ -66,8 +66,26 @@ def render(self, alias: Optional[str] = None):
...
+        For int64 range partitions, normalizes values to their partition start
+        boundary using: value - MOD(value - range_start, range_interval).
+        This prevents generating excessively large arrays of distinct values
+        when computing partitions for replacement in insert_overwrite.
+        """
+        # int64 range partitions: normalize to partition start boundary
+        if self.data_type == "int64" and self.range is not None:
+            column: str = self.field
+            if alias:
+                column = f"{alias}.{column}"
```
colin-k-rogers: We might need to properly `render` this (i.e. handle quoting / special characters in the column name)
```suggestion
                column = self.render(f"{alias}.{column}")
```

aahel: fixed this. https://github.com/dbt-labs/dbt-adapters/pull/1654/changes/f7171665951cd501aa4ae3b92f28cec45072584e

render func handles the 81-83
```
    def render(self, alias: Optional[str] = None):
        column: str = (
            self.field if not self.time_ingestion_partitioning else self.time_partitioning_field()
        )
        if alias:
            column = f"{alias}.{column}"
```

PR #1657, opened 2026-02-18
feat(athena): Add STS AssumeRole support for cross-account access

dtaniwaki:
### Problem

The Athena adapter currently supports AWS authentication via access keys, profiles, and session tokens, but not STS AssumeRole. In AWS multi-account environments, cross-account access through role assumption is a very common pattern, and without native support, users have to manually obtain temporary credentials outside of dbt before each run.

### Solution

This PR adds optional AssumeRole support by introducing four new profile fields:

```yaml
assume_role_arn: arn:aws:iam::123456789012:role/DbtRole
assume_role_external_id: my-external-id        # optional
assume_role_session_name: my-dbt-session        # optional, default: "dbt-athena"
assume_role_duration_seconds: 3600              # optional
```

When assume_role_arn is set, the session layer calls STS AssumeRole and creates a new boto3 session with the temporary credentials. Since all AWS client creation (Athena, Glue, S3, LakeFormation, etc.) flows through get_boto3_session() / get_boto3_session_from_credentials(), the change is fully transparent to the rest of the codebase. When assume_role_arn is not set, behavior is completely unchanged.

### Checklist


dbt-athena/src/dbt/adapters/athena/connections.py
```diff
@@ -61,6 +61,10 @@ class AthenaCredentials(Credentials):
     aws_access_key_id: Optional[str] = None
     aws_secret_access_key: Optional[str] = None
     aws_session_token: Optional[str] = None
+    assume_role_arn: Optional[str] = None
+    assume_role_external_id: Optional[str] = None
+    assume_role_session_name: Optional[str] = None
+    assume_role_duration_seconds: Optional[int] = None
```
nicor88: did you consider to add a default value? e.g. 3600 seconds

dbt-athena/src/dbt/adapters/athena/connections.py
```diff
@@ -61,6 +61,10 @@ class AthenaCredentials(Credentials):
     aws_access_key_id: Optional[str] = None
     aws_secret_access_key: Optional[str] = None
     aws_session_token: Optional[str] = None
+    assume_role_arn: Optional[str] = None
+    assume_role_external_id: Optional[str] = None
+    assume_role_session_name: Optional[str] = None
```
nicor88: maybe consider to add a default value, which means that the only mandatory value to use this feature is the assume_role_arn eventually.

dbt-athena/src/dbt/adapters/athena/session.py
```diff
@@ -23,25 +24,136 @@
...
+        "role_arn": credentials.assume_role_arn,
+        "external_id": getattr(credentials, "assume_role_external_id", None),
+        "session_name": getattr(credentials, "assume_role_session_name", None),
+        "duration": getattr(credentials, "assume_role_duration_seconds", None),
+        "region": getattr(credentials, "region_name", None),
+    }
+    return json.dumps(parts, sort_keys=True)
+
+
+def _evict_expired_cache_entries() -> None:
```
nicor88: it feels that you are reinventing the wheel here, check lru cache https://docs.python.org/3/library/functools.html and consider adding some ttl (https://stackoverflow.com/questions/31771286/python-in-memory-cache-with-time-to-live) IMHO nicer implementation than this, and definitely more pythonic.

dbt-athena/src/dbt/adapters/athena/session.py
```diff
@@ -23,25 +24,136 @@
...
+
+    with _assume_role_cache_lock:
+        cached = _assume_role_cache.get(cache_key)
+        if cached is not None:
+            session, expiration = cached
+            if datetime.now(timezone.utc) < expiration - timedelta(seconds=_EXPIRY_BUFFER_SECONDS):
+                LOGGER.debug(f"Using cached assumed role session for: {role_arn}")
+                return session
+
+        _evict_expired_cache_entries()
```
nicor88: as written above, consider the usage of a functools cache decorator to cache the session with some extra ttl (add the ttl parameter to the configuration, as I believe that it needs to be tweak based on the role assumption time)

nicor88 (commented):
The implementation looks about right - the only section that you must revisit is how you cache the session, which must be more pythonic.

dbt-athena/src/dbt/adapters/athena/session.py
dtaniwaki: @nicor88  I reimplemented the assume-role feature with lru_cache. Would you review it again?

dbt-athena/src/dbt/adapters/athena/connections.py
```diff
@@ -99,6 +103,7 @@ def _connection_keys(self) -> Tuple[str, ...]:
             "poll_interval",
             "aws_profile_name",
             "aws_access_key_id",
+            "assume_role_arn",
```
nicor88, line 106: I believe that you need to add also the other `assume_role_*` params.

dbt-athena/src/dbt/adapters/athena/session.py
```diff
@@ -23,25 +25,94 @@
...
+    # Increments every ttl seconds, causing lru_cache to treat each period as a distinct call
+    ttl_bucket = int(time.time() / ttl)
+    return _get_assume_role_session(base_session, key, ttl_bucket)
+
+
+@lru_cache(maxsize=1)
+def _get_assume_role_session(
+    base_session: boto3.session.Session,
+    key: _AssumeRoleParams,
+    _ttl_hash: int,
```
nicor88: nit: add a comment mentioning that `_ttl_hash` is an artificial value used to force the cache refresh.

dbt-athena/src/dbt/adapters/athena/session.py
```diff
@@ -23,25 +25,94 @@
...
+    assume_role_duration_seconds: int
+    region_name: str
+
+
+def _assume_role_session(
+    base_session: boto3.session.Session,
+    credentials: Any,
+) -> boto3.session.Session:
+    duration = credentials.assume_role_duration_seconds
+    if not (900 <= duration <= 43200):
```
nicor88, line 47: could you add a comment pointing to some docs mentioning that the duration must be in those intervals?

nicor88 (approved):
Changes looks good to me.
@colin-rogers-dbt could you and tea have a look?

nicor88 (approved):

nicor88:
@dtaniwaki I see some little issues on the pre-commits checks, mostly linting on flake8:
> dbt-athena/src/dbt/adapters/athena/session.py:4:1: F401 'datetime.datetime' imported but unused
dbt-athena/src/dbt/adapters/athena/session.py:4:1: F401 'datetime.timedelta' imported but unused
dbt-athena/src/dbt/adapters/athena/session.py:4:1: F401 'datetime.timezone' imported but unused
dbt-athena/src/dbt/adapters/athena/session.py:7:1: F401 'typing.Tuple' imported but unused

Could you please fix that?

dtaniwaki:
@nicor88 I fixed the lint issues.

nicor88 (approved):
Looks good to me, but it's on @colin-rogers-dbt and team to merge

dtaniwaki:
@colin-rogers-dbt would you review the PR?

dtaniwaki:
@colin-rogers-dbt Could you tell me what else I can do to get this this PR merged?

colin-k-rogers (approved):

colin-rogers-dbt:
> @colin-rogers-dbt Could you tell me what else I can do to get this this PR merged?

Sorry for the delay, thanks for the PR! Will get this merged.

dbt-athena/src/dbt/adapters/athena/connections.py
```diff
@@ -99,6 +103,12 @@ def _connection_keys(self) -> Tuple[str, ...]:
...
+            # external_id is not a secret; it is a shared condition value to prevent confused deputy attacks.
+            # See: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html
+            "assume_role_external_id",
```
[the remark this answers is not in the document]
dtaniwaki: The External ID is not treated as a sensitive value in AWS Athena as described in the referenced URL.

dbt-athena/src/dbt/adapters/athena/session.py
```diff
@@ -23,25 +24,94 @@
...
+    base_session = boto3.session.Session(
         aws_access_key_id=credentials.aws_access_key_id,
         aws_secret_access_key=credentials.aws_secret_access_key,
         aws_session_token=credentials.aws_session_token,
         region_name=credentials.region_name,
         profile_name=credentials.aws_profile_name,
     )
+    if credentials.assume_role_arn:
+        return _assume_role_session(base_session, credentials)
+    return base_session
```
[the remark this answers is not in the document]
dtaniwaki: Good point about the connection-level session. The TTL cache works for the impl.py callers (get_boto3_session_from_credentials) since those create short-lived clients on each operation. However, connections.py passes the session to pyathena.connect() once at connection open, so it won't benefit from cache rotation.

That said, in practice dbt manages connections per-thread and individual model executions rarely exceed the STS session lifetime (default 3600s). Switching to botocore.credentials.RefreshableCredentials would be the ideal fix, but it's a larger change given the pyathena session-passing structure. Tracking that as a future improvement — out of scope for this PR.

dbt-athena/src/dbt/adapters/athena/session.py
```diff
@@ -23,25 +24,94 @@
...
+    ttl_bucket = int(time.time() / ttl)
+    return _get_assume_role_session(base_session, key, ttl_bucket)
+
+
+@lru_cache(maxsize=1)
+def _get_assume_role_session(
+    base_session: boto3.session.Session,
+    key: _AssumeRoleParams,
+    _ttl_hash: int,  # artificial value that changes every ttl seconds to force cache invalidation
+) -> boto3.session.Session:
```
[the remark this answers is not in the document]
dtaniwaki: The original implementation assumed `base_session` would be the same object reused across calls — the intent was to build the STS call on top of a shared session. In practice, `get_boto3_session`/`get_boto3_session_from_credentials` create a new `boto3.session.Session(...)` each invocation, so the object identity changed and the cache never hit.

Fixed in 34f999c3: consolidated `_AssumeRoleParams` + `base_session` into a single hashable `_AssumeRoleKey` NamedTuple, and moved base session creation inside the cached function. Added a test that verifies the cache actually hits across multiple `get_boto3_session_from_credentials` calls.

@colin-rogers-dbt FYI — this was a real caching bug, now fixed with test coverage.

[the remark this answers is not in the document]
dtaniwaki: Reconsidered — in a dbt run there is effectively only one set of credentials throughout the process, so the base session is always logically the same. The impact is limited to redundant STS calls, not incorrect behavior. Reverted the fix (34f999c3) in https://github.com/dbt-labs/dbt-adapters/pull/1657/commits/749a947f0c39debb5c800d68891c518f1c2ea7fc and will track this as a minor optimization for a future PR.

@colin-rogers-dbt Would you trigger the CI again if it's not a problem?

dbt-athena/src/dbt/adapters/athena/session.py
```diff
@@ -23,25 +24,94 @@
...
+    LOGGER.debug(f"Assuming role: {key.assume_role_arn}")
+    sts_client = base_session.client("sts")
```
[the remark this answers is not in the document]
dtaniwaki: @colin-rogers-dbt I applied this change suggestion as having the same client settings is useful for debugging. Would you run the CI again?

dtaniwaki:
@colin-rogers-dbt The mock should've been updated. I fixed it. Would you run the CI again?

dtaniwaki:
Also, the suggestion from Copilot was not great. We should not reuse the hardcoded value for another purpose. I fixed it too.

PR #1659, opened 2026-02-18
added support for `IS NOT DISTINCT FROM`

### LINKED ISSUE #1663: [CT-2137] [Feature] Cross-database implementation of is_distinct_from

### Describe the feature

If I had a dollar for every time I had done 

```sql
select a, b, 
  a = b or (a is null and b is null) as are_the_same
from table
```
or 
```sql
select a, b, 
  coalesce(a, -1) = coalesce(b, -1) as are_the_same --very dangerous! do not do!
from table
```
,
I would have a lot of money. 

Turns out that there is a solution for this in the sql standard! 

image

We should add this as a cross-db macro - a speedy Google looks like most of our key adapters support it natively [except for Redshift](https://popsql.com/learn-sql/redshift/how-to-compare-two-values-when-one-is-null-in-redshift)

Doing nothing. 

### Who will this benefit?

This seems useful for two reasons: 
- By creating and documenting it, more people might stumble over it
- I would have given my left leg for this when building the first version of the metrics package, when I was trying to do joins that were potentially null 

Yep! But it'll be my first cross-adapter contribution so I'll need some handholding

### Anything else?

_No response_

sriramr98:
resolves #1663

### Problem

The current implementation of `equals` macro is not good for the following reasons
1. Every expression is evaluated twice resulting in unnecessary computation.
2. Engine's cannot recognise this as just checking for equals which might result in un-optimised index lookups.

Most warehouses support `IS [NOT] DISTINCT FROM` clause which provides the following benefits

1. In the default implementation, expressions are evaluated twice. Using IS NOT DISTINCT FROM will only evalate each expression once
2. Engines can recognise this as a single comparison operator and produce better plans (e.g. hash joins on nullable keys). The CASE version may not be recognised as a simple equality predicate, potentially preventing join optimisation or index usage

### Solution

Added adapter specific `equals` macro using `IS NOT DISTINCT FROM` clause in supported warehouses.

### Checklist

### DIFF
--- a/dbt-athena/src/dbt/include/athena/macros/utils/equals.sql
+++ b/dbt-athena/src/dbt/include/athena/macros/utils/equals.sql
@@ -1,3 +1,7 @@
 {% macro athena__equals(expr1, expr2) -%}
+{%- if adapter.behavior.enable_truthy_nulls_equals_macro.no_warn %}
     ({{ expr1 }} IS NOT DISTINCT FROM {{ expr2 }})
+{%- else -%}
+    ({{ expr1 }} = {{ expr2 }})
+{%- endif %}
 {%- endmacro %}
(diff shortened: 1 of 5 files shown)


sriramr98:
Temporarily removing `dbt-bigquery` tag since it takes too long for each run. Once the CI fixes are done, I'll re-enable it before the final merge.

PR #1665, opened 2026-02-24
Fix seed concurrency using intermediate tables

will-sargent-dbtlabs:
resolves #<INSERT_ISSUE_NUMBER_HERE>
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#<INSERT_DOCS_ISSUE_NUMBER_OR_NA>

### Problem

When multiple dbt jobs run `dbt seed` concurrently against the same target seed, there is a race condition that can lead to compilation errors or duplicate records. 

Currently, the `seed` materialization executes by:
1. Dropping or truncating the target table.
2. Creating the table (if it doesn't exist/was dropped).
3. Inserting the records directly into the target table in batches.

If `Job A` has created the table and is currently running `INSERT` batches, and `Job B` concurrently starts running `dbt seed` for the same table, `Job B` will drop the table out from under `Job A`. This results in `Job A` throwing a database compilation error (e.g., `Table does not exist or not authorized`). Alternatively, if both run their truncates at the exact same time before moving on to inserts, they might both insert their batches simultaneously, leading to duplicate rows in the final table. 

### Solution

This PR refactors the seed materialization to use an intermediate table, bringing its concurrency safety up to par with the standard `table` materialization. 

The seed process now executes as follows:
1. **Load via Intermediate:** The CSV data is loaded (batched inserts) into an intermediate/temporary relation (`<seed_name>__dbt_tmp`). 
2. **Atomic Swap/Insert:** 
   - **For non-full-refresh:** We run a single atomic command to truncate the target table and `INSERT INTO <target> SELECT * FROM <intermediate>`, guaranteeing no window of time where the table is dropped, preserving existing grants.
   - **For full-refresh:** We mirror the `table` materialization logic by renaming the target relation to a backup relation, renaming the intermediate relation to the target, and finally dropping the backup relation.

**Implementation Details:**
- Updated the global seed helper macros (`create_csv_table`, `reset_csv_table`, `load_csv_rows`) to accept an optional `relation` parameter.
- Updated overriding seed helper macros across all core adapters (`BigQuery`, `Spark`, `Snowflake`, `Redshift`, `Athena`) to use this `relation` parameter instead of defaulting to `this.render()`.
- Updated the main `seed` materialization in `global_project` to utilize the `make_intermediate_relation()` macro to create and manage the temp table.
- Added `TestSeedConcurrent` in `dbt-tests-adapter` to verify that concurrent executions of `dbt seed` via a `ThreadPoolExecutor` complete successfully without throwing `Table not found` errors or duplicating records.

### Adapter-specific fixes

During integration testing, several adapter-specific issues were discovered and resolved:

#### BigQuery
- **No `TRUNCATE TABLE` support:** BigQuery does not support `TRUNCATE TABLE`. Introduced the dispatchable `seed_can_use_truncate()` macro (default `true`) and a `bigquery__seed_can_use_truncate()` override returning `false`, so the global `seed.sql` takes the drop-and-recreate path instead.
- **Async `rename_relation` causes "Not found: Table" errors:** BigQuery's `rename_relation` implementation uses an asynchronous `copy_table` operation followed by `delete_table`. When the seed materialization immediately called `seed_finalize()` (to reapply table options like labels and expiration) after a rename, the target table might not yet exist. Fix: added `bigquery__seed_can_use_relation_rename()` returning `false`, forcing BigQuery to use the synchronous `seed_create_target_from_intermediate` (CTAS) path instead of rename.
- **Table options lost after intermediate swap:** When the intermediate table replaced the target, BigQuery-specific table options (labels, expiration, description) were lost. Introduced the dispatchable `seed_finalize()` macro and a `bigquery__seed_finalize()` implementation that reapplies these options to the target relation after the swap.

#### Athena
- **Backtick quoting rejected by Presto engine:** `athena__seed_create_target_from_intermediate` was using `intermediate_relation.render_hive()` (backtick-quoted) in `SELECT` statements, but Athena's Presto SQL engine requires double-quoted identifiers. Fixed by using `intermediate_relation.render()` for Presto-compatible DML while keeping `.render_hive()` for Hive DDL.
- **No `TRUNCATE TABLE` support:** Added `athena__seed_can_use_truncate()` returning `false`.

#### Postgres / Redshift
- **Connection pool exhaustion under concurrency:** `BaseSeedConcurrent`'s `setUp` fixture executed raw SQL that was incompatible with some adapters and exhausted connection pools. Removed the fixture and made `test_concurrent_seeds` resilient by catching exceptions from concurrent runs and asserting only the success of a final sequential run.

### Checklist


will-sargent-dbtlabs:
The failing test is getting cancelled via GHA rather than failing. Not sure what to do about this.  Otherwise, this seems ready to consider @dbt-labs/adapters

will-sargent-dbtlabs:
closing as stale for now

PR #1684, opened 2026-02-27
Better handling of invalid columns in models during persist_docs

### LINKED ISSUE #1690: [CT-1024] [Bug] Nonexistent column in yml file causes "can't execute an empty query"

### Current Behavior

When I run dbt run  on table.sql and table.yml has a column and a test defined on a nonexistent column I get this error:

"can't execute an empty query"

### Expected Behavior

An error message stating "Non-existent column in table.yml. Please remove the column "column1" from the yml file. "

### Steps To Reproduce

with a table1.sql file with a simple select and a few ctes and 
with a table1.yml file with a non-existent column in table1.sql final select

```
run dbt run --select table1.sql
```

### Relevant log output

```shell
22:46:04  Database Error in model table1 (models/mart/tbl/table1.sql)
22:46:04    can't execute an empty query
22:46:04    compiled SQL at target/run/mlp_dbt/models/tbl/fct/table1.sql
```

### Environment

```markdown
- OS: docker python:3.9-slim
- Python: python:3.9-slim
- dbt:1.2
```

redshift

### Additional Context

_No response_

sriramr98:
resolves dbt-labs/dbt-adapters#1690

### Problem

When we have a model with `persist_docs` enabled  with an invalid column, dbt fails with an error `can't execute an empty query`. 

This happens because we try to use `ALTER COLUMN` to add comment on the column since it doesn't exist in the model.

### Solution

We will gracefully handle this by warning users about the invalid column but letting the model build successfully since this only happens when we use `persist_docs` to document the column.

Screenshot 2026-02-27 at 8 25 52 PM

NOTE: The warning isn't very legible since there's no indicator that it's a warning. This is being handled in a separate [PR](https://github.com/dbt-labs/dbt-core/pull/12555) to improve warning messages

### Checklist


dbt-adapters/src/dbt/include/global_project/macros/adapters/persist_docs.sql
```diff
@@ -22,12 +22,36 @@
   {{ return(adapter.dispatch('persist_docs', 'dbt')(relation, model, for_relation, for_columns)) }}
 {% endmacro %}
 
+{#-- Validates documented columns against the actual database columns. Warns about any columns in column_dict that don't exist in existing_column_names. Returns a filtered column_dict containing only columns that exist. --#}
+{% macro validate_doc_columns(relation, column_dict, existing_column_names) %}
```
colin-k-rogers, line 26: Can we move this to use the dispatch + default pattern?

dbt-spark/dagger/run_dbt_spark_tests.py
```diff
@@ -9,6 +9,11 @@
 from dotenv import find_dotenv, load_dotenv
 
 PG_PORT = 5432
+# Resolve absolute paths to sibling packages (Dagger blocks relative ".." paths)
```
colin-k-rogers, line 12: these seem unrelated to the PR's focus?

dbt-athena/src/dbt/include/athena/macros/adapters/persist_docs.sql
```diff
@@ -1,6 +1,10 @@
 {% macro athena__persist_docs(relation, model, for_relation=true, for_columns=true) -%}
   {% set persist_relation_docs = for_relation and config.persist_relation_docs()%}
   {% set persist_column_docs = for_columns and config.persist_column_docs() and model.columns %}
+  {% if persist_column_docs %}
+    {% set existing_columns = adapter.get_columns_in_relation(relation) | map(attribute="name") | list %}
+    {% do validate_doc_columns(relation, model.columns, existing_columns) %}
```
colin-k-rogers, line 6: Why doesn't this follow the same pattern as the others and `{% set filtered_columns =...`?

dbt-postgres/src/dbt/include/postgres/macros/adapters.sql
```diff
@@ -206,8 +206,7 @@
 
 
 {% macro postgres__alter_column_comment(relation, column_dict) %}
-  {% set existing_columns = adapter.get_columns_in_relation(relation) | map(attribute="name") | list %}
```
colin-k-rogers, line 209: this will break anyone with a custom materialization that relies on this current behavior. If we were to cache the `get_columns_in_relation` we could keep this as is with out requiring a second call

dbt-athena/src/dbt/include/athena/macros/adapters/persist_docs.sql
sriramr98: We don't need to send in `filtered_columns` since it's being [handled](https://github.com/dbt-labs/dbt-adapters/blob/89cf3badd5ff8687addf50dedeb4be0a1c81e949/dbt-athena/src/dbt/adapters/athena/impl.py#L1084) in `persist_docs_to_glue` already.

We still call `validate_doc_columns` to warn the user. Any invalid columns are handled in `persist_docs_to_glue` method

dbt-adapters/src/dbt/include/global_project/macros/adapters/persist_docs.sql
sriramr98: We can. I just didn't see any reason someone would want to override this since it's pure logic with no interaction with adapters etc. If you think we should make it configurable, I will move this to using dispatch + default pattern.

dbt-spark/dagger/run_dbt_spark_tests.py
sriramr98: Yes, it's not related to the PR feature, but is required to test the changes in the PR reliably in our CI.

In the current setup, `dbt-adapters` and `dbt-tests-adapters` gets installed from the main branch. However, it won't work when running CI tests since we would have PRs like this one where there are changes in `dbt-adapters` and `dbt-test-adapter` that's required in-order to execute tests properly. I could raise a separate PR for this, but the change was small enough and was required in this PR to successfully run tests, I pushed it here.

Do you want me to raise a separate PR for this?

colin-k-rogers (approved):

dbt-spark/dagger/run_dbt_spark_tests.py
colin-k-rogers: let's leave it here, thanks for clarifying!

dbt-adapters/src/dbt/include/global_project/macros/adapters/persist_docs.sql
```diff
@@ -22,12 +22,36 @@
...
+  {% set existing_lower = existing_column_names | map("lower") | list %}
```
xuliangs, line 27: Do we need to consider quoting config here before deciding whether we can do case-insensitive comparison?

dbt-adapters/src/dbt/include/global_project/macros/adapters/persist_docs.sql
```diff
@@ -22,12 +22,36 @@
...
+{% endmacro %}
+
 {% macro default__persist_docs(relation, model, for_relation, for_columns) -%}
   {% if for_relation and config.persist_relation_docs() and model.description %}
     {% do run_query(alter_relation_comment(relation, model.description)) %}
   {% endif %}
 
   {% if for_columns and config.persist_column_docs() and model.columns %}
-    {% do run_query(alter_column_comment(relation, model.columns)) %}
+    {% set existing_columns = adapter.get_columns_in_relation(relation) | map(attribute="name") | list %}
```
xuliangs, line 50: I think this will introduce a redundant `get_columns_in_relation` for snowflake.
Snowflake uses `default__persist_docs` -> `alter_column_comment` -> `snowflake__alter_column_comment` which inside uses `get_columns_in_relation` already

sriramr98: Yes @xuliangs . I discussed this with @colin-rogers-dbt when this PR was in play. Here's a TLDR;

1. We can move the `get_columns_in_relation` outside `snowflake__alter_column_comment` but that'll break the contract of `alter_column_comment`. We will need to pass in `existing_columns` from outside `alter_column_comment`. If any customer has custom macro overriding `alter_column_comment`, it'll be problematic. We had [this](https://github.com/dbt-labs/dbt-adapters/pull/1684#discussion_r2938062345) discussion regarding postgres, but the same applies here.

2. We looked at the performance overhead and decided that for the adapters which had the extra `get_columns_in_relation` , the latency cost isn't that high and is only applicable here when `persist_docs` is enabled, so we were okay with the extra metadata call.

xuliangs: I didn't mean that we should move `get_columns_in_relation ` outside `snowflake__alter_column_comment `. Instead, we could either leave the `default__persist_docs` at it was, had the fix go into `redshift__persist_docs`, and `postgres__persist_docs` etc or we may create a override for snowflake without this extra call

xuliangs: > We looked at the performance overhead and decided that for the adapters which had the extra get_columns_in_relation , the latency cost isn't that high and is only applicable here when persist_docs is enabled, so we were okay with the extra metadata call.

The issue that the PR is supposed to fix was for Redshift, Postgres, there was no problem for Snowflake, so the fix should be scoped to them as well. Besides we should always be diligent to avoid any unnecessary work, not to mention that this is one extra network call.

dbt-adapters/src/dbt/include/global_project/macros/adapters/persist_docs.sql
jasonlin45: ^ yes this is a bug @sriramr98 https://docs.snowflake.com/en/sql-reference/identifiers-syntax

By default in Snowflake column identifiers are case sensitive when quoted.
fine for dbx and bq. redshift is configurable too.

PR #1694, opened 2026-03-02
Redshift: Support cross-database create_schema and drop_schema, deprecate ra3_node config

### LINKED ISSUE #1693: Migrate redshift__create_schema and redshift__drop_schema to perform operations on another database
### Describe the feature

Migrate `redshift__create_schema` and `redshift__drop_schema` to perform operations on another database

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

tauhid621:
resolves #1693, #1630

### Problem

In redshift, `create_schema` and `drop_schema` currently delegates to the postgres macros. Those macros call `adapter.verify_database()` which blocks cross-database operations unless `ra3_node` is set, and they render SQL with `include(database=False)`, which strips the database qualifier — so the schema is always created or dropped in the connection database, not the one the relation references.

Additionally, `ra3_node` and `redshift_use_show_apis` serve the same purpose (enabling cross-database support) but are checked independently in different code paths.

### Solution

- Updated `redshift__create_schema` and `redshift__drop_schema` macros to bypass the postgres implementations when redshift_use_show_apis is enabled. The new paths use `relation.without_identifier()` which includes the database qualifier, so DDL targets the correct database.
- Introduced `use_show_apis()` as a single method on the adapter that returns True when either redshift_use_show_apis or ra3_node is set. All Python and Jinja code paths now go through this method, making `ra3_node` an alias for the behavior flag.
- Added a deprecation warning at adapter startup when `ra3_node` is set.

### Checklist

#### Deprecation warning
Screenshot 2026-03-02 at 3 06 00 PM

#### Create schema getting called with database
Screenshot 2026-03-02 at 3 08 58 PM


dbt-redshift/src/dbt/adapters/redshift/impl.py
```diff
@@ -94,6 +94,12 @@ def __init__(self, config, mp_context: SpawnContext) -> None:
             lambda: self.behavior.redshift_skip_autocommit_transaction_statements.no_warn
         )
 
+        if self.config.credentials.ra3_node:
+            logger.warning(
```
colin-k-rogers: let's make this an info level for now, a lot of folks have their project set to error on warn

tauhid621: Thanks for the suggestion. I have added info level logging.

PR #1704, opened 2026-03-03
perf(athena): cache `_get_data_catalog()` result to avoid repeated STS calls

dtaniwaki:
resolves #

### Problem

`_get_data_catalog("awsdatacatalog")` is called from 12 places in `impl.py` (e.g. `list_relations_without_caching`, `get_columns_in_relation`, `swap_table`, `persist_docs_to_glue`, …). For the `awsdatacatalog` case it always calls `sts.get_caller_identity()` to look up the AWS account ID, even though that value never changes within a run.

On a small project this adds up to ~6–8 STS round-trips; on larger projects it scales roughly as `S + M × 8` where S is the number of schemas and M is the number of models.

### Solution

Add `@lru_cache(maxsize=32)` to `_get_data_catalog`. The cache key is `(self, database)`, so each adapter instance caches results independently per catalog name. `maxsize=32` is large enough to cover any realistic number of distinct catalogs while bounding memory usage.

The same caching benefits custom/federated catalog lookups, which call `athena.get_data_catalog` on every invocation.

### Checklist

### DIFF
--- a/dbt-athena/.changes/unreleased/Features-20260304-143653.yaml
+++ b/dbt-athena/.changes/unreleased/Features-20260304-143653.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Cache `_get_data_catalog()` result to avoid repeated STS calls
+time: 2026-03-04T14:36:53.558771+09:00
+custom:
+    Author: dtaniwaki
+    Issue: "1704"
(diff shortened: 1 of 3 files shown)


dtaniwaki:
@colin-rogers-dbt I think this avoids API throttling with less risk. Would you review it?

PR #1710, opened 2026-03-04
feat(snowflake): add scheduler support for dynamic tables 

igorbelianski-cyber:
Add support for the Snowflake dynamic table `scheduler` property, which controls whether the internal refresh scheduler is ENABLE or DISABLE. This is distinct from `scheduling_state` (ACTIVE/SUSPENDED), which is a Snowflake-managed runtime status.

Key changes:

- Add `Scheduler` enum (ENABLE/DISABLE) and make `target_lag` optional in `SnowflakeDynamicTableConfig` to support scheduler=DISABLE without a target lag specification.

- Apply implicit scheduler defaults in both `parse_relation_config` and `parse_relation_results`: when scheduler is not explicitly set, default to ENABLE if target_lag is present, DISABLE otherwise. This aligns the Python-side change detection logic with the Jinja macro defaults.

- Add scheduler change detection in `dynamic_table_config_changeset` and prevent target_lag from being included in ALTER when its new value is None (avoids generating `target_lag = 'None'` in SQL).

- Update CREATE, REPLACE, and ALTER macros to render `scheduler` in DDL with correct ENABLE/DISABLE defaults. Guard ALTER target_lag rendering to skip when context is None.

- Update materialization to issue `ALTER DYNAMIC TABLE ... REFRESH` when scheduler is DISABLE (explicit or implicit via missing target_lag), since Snowflake won't auto-refresh in that case.

- Read the `scheduler` column from SHOW DYNAMIC TABLES when available.

- Fix FileSystemLoader in test_alter_relation_comment_macro.py to use an absolute path, avoiding TemplateNotFound errors from different CWDs.

- Add comprehensive unit tests for scheduler config parsing, changeset detection, and change detection logic, plus functional tests covering scheduler in CREATE/ALTER/REPLACE DDL and refresh behavior.

Made-with: Cursor

resolves #

### Checklist


dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -129,6 +137,13 @@ def parse_relation_config(cls, relation_config: RelationConfig) -> Dict[str, Any
         if initialize := relation_config.config.extra.get("initialize"):  # type:ignore
             config_dict["initialize"] = initialize.upper()
 
+        if scheduler := relation_config.config.extra.get("scheduler"):  # type:ignore
+            config_dict["scheduler"] = scheduler.upper()
+        elif config_dict.get("target_lag"):
+            config_dict["scheduler"] = "ENABLE"
```
colin-k-rogers: Can we use the enum values?

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -163,15 +178,21 @@ def parse_relation_results(cls, relation_results: RelationResults) -> Dict[str,
         else:
             cluster_by = None
 
+        scheduler = dynamic_table.get("scheduler")
+        target_lag = dynamic_table.get("target_lag")
+        if scheduler is None:
+            scheduler = "ENABLE" if target_lag else "DISABLE"
```
colin-k-rogers: Same as above

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -58,11 +64,12 @@ class SnowflakeDynamicTableConfig(SnowflakeRelationConfigBase):
     schema_name: str
     database_name: str
     query: str
-    target_lag: str
     snowflake_warehouse: str
+    target_lag: Optional[str] = None
```
colin-k-rogers, line 69: `target_lag` is still required if `scheduler` is enabled right? We should raise a validation exception in that case.

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -33,6 +33,11 @@ def default(cls) -> Self:
         return cls("ON_CREATE")
 
 
+class Scheduler(StrEnum):
```
colin-k-rogers, line 37: I wonder if there's actually three states:
1. DISABLE - we don't want the MV to be updated
2. ENABLE - require target_lag (default state today
3. MANUAL - dbt runs the refresh manually on every run

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
igorbelianski-cyber: target lag is not mandatory any more  
    with scheduler = DISABLED    , target_lag will not be allowed ( we treat dynamic tables as a unit of incrementalization and completely disconnected  from the  dynamic  table built in scheduler) 
    with scheduler = ENABLED  it is  target lag is still mandatory  
    
PS 
i was debating  whether DBT needs to validate it as opposed allowing snowflake to produce appropriate precise error ( implementing it here seem like extra validation effort that doesn't add a lot of value and make chnages (if  snowflake adds some new modes hard to version)

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
igorbelianski-cyber: There are 2 states: 
  ENABLE ( default like  Dynamic TAbles /MV ) work today subject to be scheduled by Snowflake directly or part of the downstream schedule 
  DISABLED ( only manual refreshes allowed , Dynamic table scheduler will never touch it. directly or through upstream dependency).  => DT's   
  
  i do not see a meaningful scenario to justify 3rd mode  where we want to disable Dynamic tables from both DBT refreshes and Snowflake Dynamic table scheduler. 
   ( i guess  target_lag = downstream  on all DTs in the pipeline achieves similar effect functionally , but i am not seeing meaningful scenario for that case)   
   
   moreover 2 modes provide really simple syntax for users ( they just need to specify warehouse,  absence of target lag implies disabling Dynmic table scheduler and DBT running refreshes)

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
igorbelianski-cyber: done

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
igorbelianski-cyber: done

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
colin-k-rogers: Right I'm saying we _should_ do that validation here since not validating it means the user only gets the exception when dbt executes the model as opposed to when the user calls `dbt compile`. Ultimately `target_lag=None` and `scheduler='ENABLED'` isn't a legal model config and we should call that out.

igorbelianski-cyber: done

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/dynamic_table.sql
```diff
@@ -15,6 +15,16 @@
...
 
+    {%- set dynamic_table = target_relation.from_config(config.model) -%}
+    {% set scheduler_is_disabled = (dynamic_table.scheduler is not none and dynamic_table.scheduler | upper == 'DISABLE') %}
+    {% set scheduler_defaults_to_disabled = (dynamic_table.scheduler is none and dynamic_table.target_lag is none) %}
+    {% set needs_refresh = scheduler_is_disabled or scheduler_defaults_to_disabled %}
+    {% if build_sql != '' and needs_refresh %}
+        {% call statement(name="refresh") %}
+            {{ snowflake__refresh_dynamic_table(target_relation) }}
+        {% endcall %}
+    {% endif %}
```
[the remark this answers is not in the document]
igorbelianski-cyber: fixed

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/dynamic_table.sql
```diff
@@ -15,6 +15,16 @@
         {{ dynamic_table_execute_build_sql(build_sql, existing_relation, target_relation) }}
     {% endif %}
 
+    {%- set dynamic_table = target_relation.from_config(config.model) -%}
+    {% set scheduler_is_disabled = (dynamic_table.scheduler is not none and dynamic_table.scheduler | upper == 'DISABLE') %}
+    {% set scheduler_defaults_to_disabled = (dynamic_table.scheduler is none and dynamic_table.target_lag is none) %}
+    {% set needs_refresh = scheduler_is_disabled or scheduler_defaults_to_disabled %}
+    {% if build_sql != '' and needs_refresh %}
+        {% call statement(name="refresh") %}
```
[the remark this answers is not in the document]
igorbelianski-cyber: fixed

colin-k-rogers (approved):

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/alter.sql
```diff
@@ -17,20 +17,23 @@
...
+        {%- set has_set_changes = target_lag or snowflake_warehouse or (snowflake_initialization_warehouse and snowflake_initialization_warehouse.context) or scheduler or (immutable_where and immutable_where.context) -%}
 
         {#- Handle SET operations -#}
         {% if has_set_changes %}
         alter dynamic table {{ existing_relation }} set
-            {% if target_lag %}target_lag = '{{ target_lag.context }}'{% endif %}
+            {% if target_lag and target_lag.context %}target_lag = '{{ target_lag.context }}'{% endif %}
             {% if snowflake_warehouse %}warehouse = {{ snowflake_warehouse.context }}{% endif %}
             {% if snowflake_initialization_warehouse and snowflake_initialization_warehouse.context %}initialization_warehouse = {{ snowflake_initialization_warehouse.context }}{% endif %}
+            {% if scheduler %}scheduler = '{{ scheduler.context }}'{% endif %}
```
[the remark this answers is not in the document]
igorbelianski-cyber: it works both quotes and not quoted

PR #1711, opened 2026-03-04
fix(athena): partition tmp table in incremental to reduce batch scan cost

### LINKED ISSUE #1744: [Bug] `safe_create_table_as` skips partition handling for tmp tables, causing O(N) scan cost in batch incremental
### Is this a new bug?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [x] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

In `safe_create_table_as`, the `temporary=True` branch always creates the tmp table (`__dbt_tmp`) **without partitioning** by calling `create_table_as(..., skip_partitioning=True)` directly, bypassing `run_query_with_partitions_limit_catching` entirely.

```sql
-- Simplified before-state
{%- if temporary -%}
  {%- do run_query(create_table_as(temporary, relation, compiled_code, language, true)) -%}
  -- skip_partitioning=True: ignores partitioned_by config, no TOO_MANY_OPEN_PARTITIONS handling
{%- else -%}
  {%- set result = adapter.run_query_with_partitions_limit_catching(...) -%}
  ...
{%- endif -%}
```

For models with `partitioned_by` configured and using `insert_overwrite` incremental strategy, `batch_incremental_insert` reads from `__dbt_tmp` once per partition batch:

```sql
INSERT INTO {{ target }}
SELECT * FROM {{ tmp }}
WHERE partition_col IN (...)
```

Because `__dbt_tmp` has no partitions, every batch performs a **full scan** of the tmp table. With N batches, total Athena scan cost is **(N + 1) × full scan**.

Additionally, if the source query yields many distinct partition values, the unguarded CTAS fails immediately with `TOO_MANY_OPEN_PARTITIONS` instead of falling back to `create_table_as_with_partitions`.

### Expected Behavior

The tmp table should be created **with partitions** when `partitioned_by` is configured, so that:

- Each batch in `batch_incremental_insert` benefits from partition pruning, reducing total scan cost to approximately **2 × full scan** regardless of N.
- `TOO_MANY_OPEN_PARTITIONS` is handled consistently with the non-temporary path (fallback to batch creation).

Models without `partitioned_by` should be unaffected — they produce an unpartitioned CTAS as before.

### Steps To Reproduce

1. Configure an Athena model with:
   ```yaml
   config:
     materialized: incremental
     incremental_strategy: insert_overwrite
     partitioned_by: [date_col]
   ```
2. Run `dbt run` with a large dataset that yields many distinct partition values.
3. Observe Athena query history — each batch INSERT scans the full `__dbt_tmp` table instead of a single partition slice.

### Relevant log output

```shell

```

### Environment

```markdown
- OS: macOS
- Python: 3.12.8
- dbt-adapters: 1.22.8
- dbt-athena: 1.9.4
```

### Additional Context

_No response_

dtaniwaki:
resolves #1744

Thank you for maintaining this project! I'd appreciate your review on this fix to the Athena adapter's incremental materialization.

### Problem

For incremental models with many partitions (e.g. `bucket(col, 256)`), batch processing is extremely slow due to redundant full table scans.

**Before this PR**, the flow for `force_batch=True` was:

| Step | Operation | Queries |
|------|-----------|---------|
| 1 | CTAS: source → `__tmp_not_partitioned` (no partitions) | 1 |
| 2 | Batch INSERT: `__tmp_not_partitioned` → `__dbt_tmp` (partitioned) | N |
| 3 | Batch INSERT: `__dbt_tmp` → target | N |
| | **Total** | **2N+1** |

Each of the N batch INSERTs in step 2 performs a **full scan** of the unpartitioned tmp table, making total scan cost O(N) × table size.

#### Commit 1: Partition the tmp table (for insert_overwrite / merge)

Unified the `temporary=True` and `temporary=False` code paths in `safe_create_table_as` so that tmp tables now go through `run_query_with_partitions_limit_catching` and fall back to `create_table_as_with_partitions` on `TOO_MANY_OPEN_PARTITIONS`.

With this change, `__dbt_tmp` is created **with partitions** (when `partitioned_by` is configured), so each batch in step 3 benefits from partition pruning. Reduces scan cost for `insert_overwrite` and `merge` strategies.

Models without `partitioned_by` are unaffected — they produce an unpartitioned CTAS as before.

#### Commit 2: Skip partitioning entirely for Iceberg append strategy

For **Iceberg** `append` strategy, partitioning the tmp table is **unnecessary** because the final INSERT into the target can be done in a single query.

**Key insight:** `TOO_MANY_OPEN_PARTITIONS` is a **CTAS-specific** limitation in Athena. `INSERT INTO` an existing Iceberg table does not hit this limit — the Trino engine can sort/shuffle data by partition key and write partitions sequentially without keeping all partition writers open simultaneously. Iceberg's metadata commit only appends manifest entries, avoiding per-partition Glue Catalog registration.

This means:
- **CTAS** (creating `__dbt_tmp`): needs batch splitting when partitions > 100
- **INSERT INTO existing Iceberg table** (tmp → target): can always be done in a single query

**After this PR**, the flow for Iceberg append is:

| Step | Operation | Queries |
|------|-----------|---------|
| 1 | CTAS: source → `__dbt_tmp` (skip_partitioning=True) | 1 |
| 2 | INSERT: `__dbt_tmp` → target | 1 |
| | **Total** | **2** |

Falls back to batch INSERT if `TOO_MANY_OPEN_PARTITIONS` is unexpectedly raised.

**Hive tables** continue to use the standard path (`safe_create_table_as` + `force_batch`) since Hive `INSERT INTO` may still be subject to partition limits.

**Real production impact:**
- Table: `fitbit_heart_rate` with `bucket(uid_hash, 256)` + `day(import_time_ts6)`
- Before: 500+ batch queries × 82 GB scan each → 5+ hours, CodeBuild timeout
- After: 2 queries total

### Checklist

### DIFF
--- a/dbt-athena/.changes/unreleased/Fixes-20260311-141407.yaml
+++ b/dbt-athena/.changes/unreleased/Fixes-20260311-141407.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Partition tmp table in incremental to reduce batch scan cost
+time: 2026-03-11T14:14:07.932048+09:00
+custom:
+    Author: dtaniwaki
+    Issue: "1744"
--- a/dbt-athena/src/dbt/include/athena/macros/materializations/models/table/create_table_as.sql
+++ b/dbt-athena/src/dbt/include/athena/macros/materializations/models/table/create_table_as.sql
@@ -207,16 +207,11 @@
       {%- do create_table_as_with_partitions(temporary, relation, compiled_code, language) -%}
       {%- set query_result = relation ~ ' with many partitions created' -%}
     {%- else -%}
-        {%- if temporary -%}
-          {%- do run_query(create_table_as(temporary, relation, compiled_code, language, true)) -%}
-          {%- set compiled_code_result = relation ~ ' as temporary relation without partitioning created' -%}
-        {%- else -%}
-          {%- set compiled_code_result = adapter.run_query_with_partitions_limit_catching(create_table_as(temporary, relation, compiled_code)) -%}
-          {%- do log('COMPILED CODE RESULT: ' ~ compiled_code_result) -%}
-          {%- if compiled_code_result == 'TOO_MANY_OPEN_PARTITIONS' -%}
+        {%- set compiled_code_result = adapter.run_query_with_partitions_limit_catching(create_table_as(temporary, relation, compiled_code)) -%}
+        {%- do log('COMPILED CODE RESULT: ' ~ compiled_code_result) -%}
+        {%- if compiled_code_result == 'TOO_MANY_OPEN_PARTITIONS' -%}
             {%- do create_table_as_with_partitions(temporary, relation, compiled_code, language) -%}
             {%- set compiled_code_result = relation ~ ' with many partitions created' -%}
-          {%- endif -%}
         {%- endif -%}
     {%- endif -%}
     {{ return(compiled_code_result) }}
--- a/dbt-athena/tests/unit/test_safe_create_table_as.py
+++ b/dbt-athena/tests/unit/test_safe_create_table_as.py
@@ -1,0 +1,165 @@
+"""
+Unit tests for the safe_create_table_as macro in create_table_as.sql.
+
+Tests the full macro end-to-end using jinja2.FileSystemLoader with stubbed
+dbt context, following the pattern used in test_get_partition_batches.py.
+"""
+
+import os
+from unittest import mock
+
+import jinja2
+
+_TABLE_DIR = os.path.normpath(
+    os.path.join(
+        os.path.dirname(__file__),
+        os.pardir,
+        os.pardir,
+        "src",
+        "dbt",
+        "include",
+        "athena",
+        "macros",
+        "materializations",
+        "models",
+        "table",
+    )
+)
+
+
+def _render_safe_create_table_as(temporary, run_query_result="success"):
+    """Load and render safe_create_table_as with stubbed dbt context.
+
+    Args:
+        temporary: Whether the relation is a tmp table.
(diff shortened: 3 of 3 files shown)


nicor88:
@dtaniwaki, what you proposed here is against the original implementation decision - when we create a temporary table, we always write unpartitioned data, to then finally write to the final target location in a partitioned table.
We are aware that the first unpartititioned write lead to a full scan - this is because as you wrote, we want to bypass the `TOO_MANY_OPEN_PARTITIONS` entirely - therefore, I discurage to proceed with your approach as it can lead to other hidden issues.

dtaniwaki:
@nicor88 I see. Then, how can we avoid massive full scan queries against huge data with massive number of partitions? I mistakenly created a model of this situation and waisted lots of money...

nicor88:
@dtaniwaki how about introducing the approach that you suggested, but being able to control it via a "config" flag? The flag can be false by default, and properly documented, but then in your case you can set to "true".

doing so, we avoid any regretion, and you and other users are covered for those edge cases - I believe that that's the best compromise. Think about a good descriptive name for such configuration flag.

dtaniwaki:
Closing this PR. The approach of partitioning the tmp table in `create_table_as_with_partitions` introduced more complexity than expected and caused severe performance regressions in certain configurations (e.g. Iceberg append with `bucket(col, 256)`). The core assumption that `INSERT INTO` existing Iceberg tables avoids `TOO_MANY_OPEN_PARTITIONS` has not been verified against official documentation. I plan to revisit this with a different approach after further investigation.

PR #1717, opened 2026-03-05
fix(athena): force Hive table type for tmp staging table in create_table_as_with_partitions

dtaniwaki:
resolves #

Thanks for maintaining this project! I'd appreciate a review of this bug fix for Iceberg models using `create_table_as_with_partitions`.

### Problem

When an Iceberg model exceeds the partition limit and falls back to `create_table_as_with_partitions`, the macro creates a temporary staging table (`__tmp_not_partitioned`) by calling `create_table_as()`, which inherits the model's `table_type='iceberg'` config.

Iceberg requires an **empty S3 location** when creating a table. So when `num_iceberg_retries > 0` triggers a retry, the data written by the previous attempt remains in S3, causing the next attempt to fail with:

```
ICEBERG_FILESYSTEM_ERROR: Cannot create a table on a non-empty location
```

This puts the model in an unrecoverable retry loop — each retry fails for the same reason, leaving the tmp table and its S3 data behind.

### Solution

Force the tmp staging table to use `table_type='hive'` regardless of the model's `table_type` config.

Hive is appropriate here for two reasons:

1. **Correctness**: Hive external tables have no restriction on the S3 location state, so retries succeed even if previous data was not fully cleaned up.
2. **Performance**: The tmp staging table is short-lived and read sequentially during batch inserts. Hive avoids the overhead of Iceberg snapshot and manifest file operations on a table that is immediately dropped.

The S3 location is explicitly cleaned via `adapter.delete_from_s3()` before creation, preserving retry-safety.

```sql
-- Before: inherits table_type from model config (may be 'iceberg')
{%- do run_query(create_table_as(True, tmp_relation, compiled_code)) -%}

-- After: explicitly use Hive for the tmp staging table
create table {{ tmp_relation }}
with (
  table_type='hive',
  is_external=true,
  external_location='{{ tmp_location }}',
  format='{{ format }}'
)
as {{ compiled_code }}
```

### Checklist

### DIFF
--- a/dbt-athena/tests/unit/test_create_hive_staging_table_as.py
+++ b/dbt-athena/tests/unit/test_create_hive_staging_table_as.py
@@ -60,7 +60,9 @@ def _run_macro(table_type="hive", work_group_output_location_enforced=False):
     """Call athena__create_hive_staging_table_as and return (sql, mock_adapter)."""
     mock_adapter = mock.Mock()
     mock_adapter.generate_s3_location.return_value = _TMP_LOCATION
-    mock_adapter.is_work_group_output_location_enforced.return_value = work_group_output_location_enforced
+    mock_adapter.is_work_group_output_location_enforced.return_value = (
+        work_group_output_location_enforced
+    )
 
     config_data = {
         "table_type": table_type,


dtaniwaki:
I found an issue of timestamp precision between TIMESTAMP(3) of Hive and TIMESTAMP(6) of Iceberg in this approach.

Issue #1480, opened 2025-12-02
[Bug] dbt-snowflake doesn't handle lowercase table names from `SHOW OBJECTS`


dataders:
### Is this a new bug?

- [x] I believe this is a new bug
- [x] I have searched the existing issues, and I could not find an existing issue for this bug

### Which packages are affected?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [x] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

via @ameyakhedekar ([community slack thread](https://getdbt.slack.com/archives/C07P3K1NN7N/p1762881266448349))

if you use a Catalog-Linked Database (CLD) attached to Unity Catalog, you might encounter this error

```
Compilation Error in model dbt found an approximate match. Please delete ...  
```

the working hypothesis is that, for time immemorial, Snowflake's `SHOW OBJECTS` would return case-insensitive tables names which means dbt receives upper-cased identifiers (provided there was now account-level override).

However, tables in the CLD pointing to Unity Catalog return case-sensitive identifiers by default

We suspect this is what is causing the approximate match

### Expected Behavior

dbt-snowflake should account for potential case-sensitivity in the results of `SHOW OBJECTS` as a result of external catalog identifiers

### Steps To Reproduce

try to materialize a table to a CLD that points to Unity Catalog

### Relevant log output

```shell

```

### Environment

```markdown
- OS:
- Python:
- dbt-adapters:
- <adapter>:
```

jtcohen6:
The logic for raising an "approximate match" error is tricky:
https://github.com/dbt-labs/dbt-adapters/blob/8129e465b856f5f147594a0153ec631076020587/dbt-adapters/src/dbt/adapters/base/relation.py#L180-L193

From what I remember (and from rereading the code), when dbt-core looks up a relation in the cache:
- If project-level `quoting` is turned on for this field (database/schema/identifier), then the identifiers must match exactly to be a "cache hit"
- If project-level `quoting` is turned off (default on Snowflake), and the identifiers aren't the same, but they differ _only_ in the presence of **the quote character**, then dbt-core will raise this "approximate match" error

So I'm wondering if this error crops up in cases where:
[...]

PR #1724, opened 2026-03-06
Fix 1439/dbt snowflake doesn't handle lowercase table names

ameyakhedekar:
## Summary

resolves: https://github.com/dbt-labs/dbt-adapters/issues/1480

This PR adds support for **Databricks Unity Catalog** (and other non-CTAS catalogs) when using Iceberg REST catalog linked databases in dbt-snowflake, and fixes the `ApproximateMatchError` that occurs when CLD identifiers are lowercase.

---

### Issue 1: Unity Catalog doesn't support CTAS

**Problem**
Table creation fails for Unity Catalog because Snowflake's `CREATE ICEBERG TABLE ... AS SELECT` (CTAS) is not supported for Unity-backed catalog linked databases.

**Root Cause**
The only non-CTAS code path was hardcoded to check for `catalog_linked_database_type == 'glue'` in the `snowflake__create_table_as` macro. There was no generic way for other catalogs (Unity, or future ones) to opt into the same insert-into flow.

**Fix**
- Added `ctas_not_supported` adapter property (default `false`) that explicitly routes table creation through an insert-into macro instead of CTAS
- Renamed `snowflake__create_table_iceberg_rest_with_glue` → `snowflake__create_insert_into_table_iceberg_rest` to reflect its generic purpose
- The Glue type check is preserved for backwards compatibility

---

### Issue 2: ApproximateMatchError for lowercase CLD identifiers

**Problem**
dbt fails with `ApproximateMatchError` when a catalog linked database uses lowercase identifiers:

```
When searching for a relation, dbt found an approximate match.
Instead of: "my_cld"."my_schema"."my_table"
Did you mean one of these?
  - "MY_CLD"."MY_SCHEMA"."MY_TABLE"
```

**Root Cause**

The issue is due to the current implementation in two places:

1. **`_parse_list_relations_result()`** — Hardcoded `quote_policy = True` for all relations, causing case-sensitive comparison. The uppercased identifier doesn't match the lowercase one from the CLD.
2. **`_make_match_kwargs()`** — Always uppercased identifiers when `quoting=False`, regardless of catalog type.

Together, these cause a mismatch: dbt uppercases the identifiers it's looking for, but the actual CLD identifiers are lowercase and case-sensitively compared.

**Fix**
- Added `preserve_identifier_case` adapter property (default not set) that, when `true`:
  - Skips uppercasing in `_make_match_kwargs()` so identifiers stay lowercase
  - Sets `quote_policy` to `False` in `_parse_list_relations_result()` so comparisons are case-insensitive
- Added `get_catalog_integration_by_database()` to `CatalogIntegrationClient` to look up a catalog integration by its linked database name (case-insensitive), enabling per-database behavior in the adapter

---

### Configuration

Two new properties can be set in `adapter_properties` for Iceberg REST catalog integrations:

```yaml
catalogs:
  - name: unity_iceberg_rest_catalog
    active_write_integration: unity_integration
    write_integrations:
      - name: unity_integration
        catalog_type: iceberg_rest
        table_format: iceberg
        adapter_properties:
          catalog_linked_database: my_unity_db
          catalog_linked_database_type: unity
          ctas_not_supported: true            # use insert-into instead of CTAS
          preserve_identifier_case: true      # don't uppercase identifiers
```

| Property | Default | Description |
|----------|---------|-------------|
| `ctas_not_supported` | `false` | Routes table creation through an insert-into macro instead of CTAS. Required for catalogs like Unity that don't support CTAS. |
| `preserve_identifier_case` | not set | Keeps identifiers in their original case instead of uppercasing. Prevents `ApproximateMatchError` when CLD identifiers are lowercase. |

---

### What Changed

**dbt-adapters**
- `catalogs/_client.py` — Added `get_catalog_integration_by_database()` to look up a catalog integration by its linked database name (case-insensitive)

**dbt-snowflake (Python)**
- `catalogs/_iceberg_rest.py` — Added `ctas_not_supported` and `preserve_identifier_case` to the integration class; `ctas_not_supported` is also plumbed through to the relation dataclass for macro use
- `impl.py` — Added `_preserve_identifier_case()` method; modified `_make_match_kwargs()` to skip uppercasing and `_parse_list_relations_result()` to skip force-quoting when `preserve_identifier_case` is enabled

**dbt-snowflake (Macros)**
- `create.sql` — Renamed `snowflake__create_table_iceberg_rest_with_glue` → `snowflake__create_insert_into_table_iceberg_rest`; routing now checks `ctas_not_supported` flag (in addition to Glue type check for backwards compatibility); added model contract enforcement support

**Tests**
- Added Unity Catalog functional test class in `test_iceberg_partition_by.py`
- Added 13 new unit tests across 4 test classes:
  - `ctas_not_supported` defaults and plumbing
  - `preserve_identifier_case` storage and lookup
  - `_make_match_kwargs` case preservation
  - `_parse_list_relations_result` quote policy

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,25 @@ This repository is a monorepo containing the following packages:
 
 Please refer to each of these packages for more specific information.
 
+## Releases
+
+All of our packages are merged off of main except for `dbt-adapters` and `dbt-tests-adapter`. Therefore merging a pull request to main does not automatically put it in the queue for the next release. To do so please add the 'promote to stable' label to the PR once it's been merged.
+
+The reason we do this is to allow us to patch the previous minor version with updates (i.e. what's in stable) as needed while preparing what's on main (the next minor release) to be ready for release.
+
+### Upcoming Minor Releases
+
+The following milestones track features that we're planning for the next minor version release of each adapter:
+
+- [dbt-athena v1.11.0](https://github.com/dbt-labs/dbt-adapters/milestone/3) (2 PRs)
+- [dbt-bigquery v1.12.0](https://github.com/dbt-labs/dbt-adapters/milestone/4) (31 PRs)
+- [dbt-postgres v1.11.0](https://github.com/dbt-labs/dbt-adapters/milestone/5) (6 PRs)
+- [dbt-redshift v1.11.0](https://github.com/dbt-labs/dbt-adapters/milestone/2) (24 PRs)
+- [dbt-snowflake v1.12.0](https://github.com/dbt-labs/dbt-adapters/milestone/6) (23 PRs)
+- [dbt-spark v1.11.0](https://github.com/dbt-labs/dbt-adapters/milestone/7) (2 PRs)
+
+**Note:** PRs in these milestones may have been merged to `main` but not yet been promoted to the `stable` branch for patch releases in the current minor version.
+
 # Getting started
 
 ## Install dbt
@@ -55,7 +74,3 @@ Please refer to each of these packages for more specific information.
 # Contributors ✨
 
 Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/docs/en/emoji-key)):
-
-<a href="https://github.com/dbt-labs/dbt-adapters/graphs/contributors">
-    <img src="https://contrib.rocks/image?repo=dbt-labs/dbt-adapters" />
-</a>
--- a/dbt-adapters/.changes/unreleased/Fixes-20260125-094319.yaml
+++ b/dbt-adapters/.changes/unreleased/Fixes-20260125-094319.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Set grant_config before applying grants
+time: 2026-01-25T09:43:19.962275-05:00
+custom:
+    Author: katieclaiborne-duet
+    Issue: "1503"
--- a/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/clone/clone.sql
+++ b/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/clone/clone.sql
@@ -22,6 +22,7 @@
   -- Otherwise, this will be a view
 
   {% set can_clone_table = can_clone_table() %}
+  {%- set grant_config = config.get('grants') -%}
 
   {%- if other_existing_relation and other_existing_relation.type == 'table' and can_clone_table -%}
 
--- a/dbt-bigquery/.changes/unreleased/Fixes-20260125-095315.yaml
+++ b/dbt-bigquery/.changes/unreleased/Fixes-20260125-095315.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Add testing for grant preservation during clone operations
+time: 2026-01-25T09:53:15.698657-05:00
+custom:
+    Author: katieclaiborne-duet
+    Issue: "1503"
--- a/dbt-bigquery/tests/functional/adapter/dbt_clone/test_dbt_clone.py
+++ b/dbt-bigquery/tests/functional/adapter/dbt_clone/test_dbt_clone.py
@@ -1,9 +1,16 @@
+from copy import deepcopy
+
 import pytest
+
+from dbt.tests.adapter.dbt_clone import fixtures
 from dbt.tests.adapter.dbt_clone.test_dbt_clone import (
+    BaseClone,
     BaseClonePossible,
     BaseCloneSameSourceAndTarget,
     BaseCloneNotPossible,
 )
+from dbt.tests.adapter.grants.base_grants import BaseGrants
+from dbt.tests.util import run_dbt, get_connection
 
 
 class TestBigQueryClonePossible(BaseClonePossible):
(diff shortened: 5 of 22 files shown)


dbt-snowflake/tests/functional/iceberg/test_iceberg_partition_by.py
```diff
@@ -305,3 +344,43 @@ def test_partition_by_glue_iceberg_rest_catalog(self, project, setup_class):
         run_dbt(["run", "--log-level", "debug"])
         iceberg_sql = get_cleaned_model_ddl_from_file("glue_table.sql")
         assert 'partition by ("order_date")' in iceberg_sql
+
+
+class TestPartitionByIcebergRestUnityCatalog(TestPartitionByIcebergRestGlueCatalog):
```
aahel: this test is failing in ci.

 `TestPartitionByIcebergRestUnityCatalog` inherits from `TestPartitionByIcebergRestGlueCatalog`, which causes it to also run  the inherited `test_partition_by_glue_iceberg_rest_catalog` method. That test expects `glue_table.sql`, but the Unity class only provides unity_table.sql.                                                                                           
                                                                                                                           
  Fix: Change the parent class to TestIcebergPartitionBy and add the missing fixtures:

ameyakhedekar: @aahel Thanks for the tip. 

@aahel @dataders 
I have made the updates, and the tests are passing on local.
Github-actions removed the label. So the checks are not running here. I need your help to add the label back.

dbt-adapters/src/dbt/adapters/catalogs/_client.py
```diff
@@ -45,6 +45,25 @@ def get(self, name: str) -> CatalogIntegration:
...
+        Find a catalog integration by its associated catalog_linked_database name.
+
+        Args:
+            database: The database name to search for
+
+        Returns:
+            The CatalogIntegration if found, None otherwise
+        """
+        for integration in self.__catalog_integrations.values():
+            if hasattr(integration, "catalog_linked_database"):
```
aahel, line 59: get_catalog_integration_by_database lives in the base dbt-adapters package, but catalog_linked_database is a Snowflake-specific concept that doesn't exist on the base CatalogIntegration class. The hasattr check is a sign that the abstraction boundary is being crossed.

Since `_preserve_identifier_case` in impl.py is the only caller, I'd suggest keeping this logic entirely in Snowflake code. You could add a simple __iter__ to CatalogIntegrationClient in the base (just exposing the values, no Snowflake knowledge), and move the lookup into _preserve_identifier_case itself using isinstance(integration, IcebergRestCatalogIntegration):

```
### _client.py

def __iter__(self):
    return iter(self.__catalog_integrations.values())

### snowflake/impl.py
def _preserve_identifier_case(self, database: str) -> bool:
    if not database:
        return False
    for integration in self._catalog_client:
        if (
            isinstance(integration, IcebergRestCatalogIntegration)
            and integration.catalog_linked_database
            and integration.catalog_linked_database.upper() == database.upper()
        ):
            return integration.preserve_identifier_case is True
    return False
```
This way the base package stays adapter-agnostic, and the Snowflake-specific field access is explicit via isinstance rather than implicit via hasattr.

dbt-snowflake/tests/functional/iceberg/test_iceberg_partition_by.py
```diff
@@ -179,6 +198,26 @@ def catalogs(self):
...
+                    "write_integrations": [
+                        {
+                            "name": "unity_iceberg_rest_catalog_integration",
+                            "catalog_type": "iceberg_rest",
+                            "table_format": "iceberg",
+                            "adapter_properties": {
+                                "catalog_linked_database": os.getenv(
+                                    "SNOWFLAKE_TEST_CATALOG_LINKED_DATABASE_GLUE"  # ideally can be driven by Catalog linked DB with Unity catalog. Reusing Glue for now, as behaviour is same.
+                                ),
+                                "ctas_not_supported": True,  # allows to use same insert-into process like Glue, without being vendor specific
```
aahel, line 213: functional test for `preserve_identifier_case` is missing

dbt-snowflake/tests/functional/iceberg/test_iceberg_partition_by.py
```diff
@@ -179,6 +198,26 @@ def catalogs(self):
...
+                    "name": "unity_iceberg_rest_catalog",
+                    "active_write_integration": "unity_iceberg_rest_catalog_integration",
+                    "write_integrations": [
+                        {
+                            "name": "unity_iceberg_rest_catalog_integration",
+                            "catalog_type": "iceberg_rest",
+                            "table_format": "iceberg",
+                            "adapter_properties": {
+                                "catalog_linked_database": os.getenv(
+                                    "SNOWFLAKE_TEST_CATALOG_LINKED_DATABASE_GLUE"  # ideally can be driven by Catalog linked DB with Unity catalog. Reusing Glue for now, as behaviour is same.
```
aahel, line 211: i think this should be tested with UNITY. We would need to add UNITY CLD to our CI

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_iceberg_rest.py
```diff
@@ -25,6 +25,7 @@ class IcebergRestCatalogRelation:
     target_file_size: Optional[str] = None
     max_data_extension_time_in_days: Optional[int] = None
     auto_refresh: Optional[bool] = None
+    ctas_not_supported: Optional[bool] = False
```
VersusFacit, line 28: This should be a behavior flag which will then be rolled forward in time because CTAS is now supported. It will be released GA after a period of testing

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -142,22 +151,42 @@ def _catalog_filter_table(
         lowered = table.rename(column_names=[c.lower() for c in table.column_names])
         return super()._catalog_filter_table(lowered, used_schemas)
 
+    def _preserve_identifier_case(self, database: str) -> bool:
```
VersusFacit, line 154: ideally this handled as part of the warehouse knowing what to do based on the type of rest catalog and the subsequent rules therein, as opposed to a toggle on the model. I propose we opt for a more sophisticated solution @aahel: we plan to do something similar in Fusion. based on the rules of target datawarehouse, separate quoting and casing rules from the overall project to avoid conflicts. While this would work in the short term, it's a product in part of not having the right architecture

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -142,22 +151,42 @@ def _catalog_filter_table(
...
+        return (
+            integration is not None
+            and integration.catalog_type == constants.ICEBERG_REST_CATALOG_TYPE
+            and getattr(integration, "preserve_identifier_case", None) is True
+        )
+
     def _make_match_kwargs(self, database, schema, identifier):
         # if any path part is already quoted then consider same casing but without quotes
         quoting = self.config.quoting
+        should_uppercase = not self._preserve_identifier_case(database)
```
VersusFacit, line 175: will this boolean interfere with the many other forms of database override rules we support?  Model, properties files, dbt_profile level, catalog_database -- all have a place in the heirarchy and I think we'd have to be careful here

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create.sql
```diff
@@ -10,9 +11,10 @@
         {%- elif catalog_relation.catalog_type == 'BUILT_IN' -%}
             {{ snowflake__create_table_built_in_sql(relation, compiled_code) }}
         {%- elif catalog_relation.catalog_type == 'ICEBERG_REST' -%}
-            {%- if catalog_relation.catalog_linked_database_type is defined and
-            catalog_relation.catalog_linked_database_type == 'glue' -%}
-                {{ snowflake__create_table_iceberg_rest_with_glue(relation, compiled_code, catalog_relation) }}
+            {%- if catalog_relation.ctas_not_supported or
+            (catalog_relation.catalog_linked_database_type is defined and
+            catalog_relation.catalog_linked_database_type|lower == 'glue') -%}
```
VersusFacit, line 16: probably best to support this branching primarily on glue? I'd strongly propose **not** being dry here to have easier to read jinja. I'd also propose normalization happen upstream

dbt-snowflake/src/dbt/include/snowflake/macros/relations/table/create.sql
```diff
@@ -264,10 +266,10 @@ as (
 {%- endmacro %}
 
 
-{% macro snowflake__create_table_iceberg_rest_with_glue(relation, compiled_code, catalog_relation) -%}
+{% macro snowflake__create_insert_into_table_iceberg_rest(relation, compiled_code, catalog_relation) -%}
 {#-
-    Creates an Iceberg table for Catalog Linked Databases (e.g., AWS Glue) with explicit column definitions.
-    This is used when CTAS is not supported.
+    Creates an Iceberg table for Catalog Linked Databases (e.g., AWS Glue, Unity Catalog)
```
VersusFacit, line 271: prefer CTAS, for this older case, make another macro, but name it legacy. At best, use a behavior flag to flip on the CTAS behavior and default to this essentially duplicated logic

dbt-snowflake/tests/functional/iceberg/test_iceberg_partition_by.py
VersusFacit: we have a couple in the adapters snowflake ready to go o7

VersusFacit (commented):
Honestly this is a wonderful PR -- thank you so much for putting in the hard work here. The main problem: we're in the middle of a big rework on catalogs. In particular, we need to avoid crossing paths in the name of DRY (we observe that as a failure mode in this particular layer) and find a way to handle casing more architecturally. That would change your interface meaningfully. That said, this provides a good start the team can take over the finish line appropriately

dbt-adapters/src/dbt/adapters/catalogs/_client.py
ameyakhedekar: Good point @aahel . Will look into this.

dbt-snowflake/src/dbt/adapters/snowflake/catalogs/_iceberg_rest.py
ameyakhedekar: Thanks @VersusFacit Any examples of changes required here? Its false by default.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
ameyakhedekar: @VersusFacit
Thanks for raising this point. By "warehouse knowing what to do based on the type of rest catalog" , you mean based on catalog type?

Isnt the tradeoff that new catalog adoption and upgrades will require code changes? End user can use this flag in the adapter_properties (catalogs.yml). This setting is associated with a catalog-integration isnt it?

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
ameyakhedekar: Hi @VersusFacit 
This flag is with associated with a catalog integration(adapter_properties), akin to catalog_linked_database_type. Is hierarchy applied to those as well?

ameyakhedekar:
> Honestly this is a wonderful PR -- thank you so much for putting in the hard work here. The main problem: we're in the middle of a big rework on catalogs. In particular, we need to avoid crossing paths in the name of DRY (we observe that as a failure mode in this particular layer) and find a way to handle casing more architecturally. That would change your interface meaningfully. That said, this provides a good start the team can take over the finish line appropriately

Thanks @VersusFacit  Appreciate it. Please do the needful. Happy to contribute any way I can.

PR #1740, opened 2026-03-11
fix(athena): exclude ICEBERG_FILESYSTEM_ERROR from outer retry

### LINKED ISSUE #1739: [Bug] dbt-athena: ICEBERG_FILESYSTEM_ERROR is retried by outer retry, causing repeated failures with the same S3 location
### Is this a new bug?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [x] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

When a CTAS (CREATE TABLE AS SELECT) query fails with an error like `TOO_MANY_OPEN_PARTITIONS`, Athena leaves partial data at the S3 location (Iceberg CTAS is not atomic). The outer retry in `AthenaCursor.execute()` then re-runs the exact same SQL — including the same hardcoded S3 location UUID — but `delete_from_s3` is only called during Jinja evaluation (first attempt only), so the partial data remains.

As a result, the retry hits `ICEBERG_FILESYSTEM_ERROR: Cannot create a table on a non-empty location`, and the outer retry keeps retrying this error as well, repeating the same failure up to `num_retries` times (default 5).

The existing code comment already acknowledges this situation for `TOO_MANY_OPEN_PARTITIONS`:

```python
# No need to retry if TOO_MANY_OPEN_PARTITIONS occurs.
# Otherwise, Athena throws ICEBERG_FILESYSTEM_ERROR after retry,
# because not all files are removed immediately after first try to create table
```

However, `ICEBERG_FILESYSTEM_ERROR` itself is not excluded from the retry condition.

### Expected Behavior

When `ICEBERG_FILESYSTEM_ERROR` occurs, the outer retry should not attempt the same query again, since it will always fail with the same error at the same S3 location. The error should be re-raised immediately, preserving the original error context for the user.

### Steps To Reproduce

Note: `TOO_MANY_OPEN_PARTITIONS` itself is already excluded from the outer retry when called via `run_query_with_partitions_limit_catching` (which passes `catch_partitions_limit=True`). The issue occurs when the **initial** CTAS failure is a different transient error (e.g., an Athena internal error) that the outer retry attempts to recover from:

1. Configure a dbt-athena project with an Iceberg model using CTAS
2. Trigger a situation where the initial CTAS fails mid-write with a transient error (not `TOO_MANY_OPEN_PARTITIONS`)
3. The outer retry re-runs the same SQL — Iceberg CTAS is non-atomic, so partial data remains at the same hardcoded S3 UUID location
4. The retry hits `ICEBERG_FILESYSTEM_ERROR: Cannot create a table on a non-empty location`
5. Since `ICEBERG_FILESYSTEM_ERROR` is not excluded from the outer retry, it keeps retrying up to `num_retries` times — all of which fail with the same error

### Relevant log output

```shell
01:08:17  Athena query failed: ICEBERG_FILESYSTEM_ERROR: Cannot create a table on a non-empty location: s3://.../my_table/98f97181-b94a-405e-8213-1ea02b5760e7
01:08:28  Athena query failed: ICEBERG_FILESYSTEM_ERROR: Cannot create a table on a non-empty location: s3://.../my_table/98f97181-b94a-405e-8213-1ea02b5760e7
01:08:39  Athena query failed: ICEBERG_FILESYSTEM_ERROR: Cannot create a table on a non-empty location: s3://.../my_table/98f97181-b94a-405e-8213-1ea02b5760e7
... (same UUID, 5 times total)


Note: the S3 UUID is identical across all attempts, confirming that the same SQL is being retried without cleaning up the partial data.
```

### Environment

```markdown
- OS: macOS
- Python: 3.12.8
- dbt-adapters: 1.22.8
- dbt-athena: 1.9.4
```

### Additional Context

The fix is straightforward: add `"ICEBERG_FILESYSTEM_ERROR" in str(e)` to the no-retry condition in the outer `@retry` decorator in `connections.py`, alongside the existing `TOO_MANY_OPEN_PARTITIONS` exclusion.

Note that `ICEBERG_COMMIT_ERROR` is intentionally retried by the *inner* retry (for parallel insert conflicts) and is unaffected by this change.

dtaniwaki:
resolves #1739

Thank you for maintaining this project! I'd appreciate your review on this small bug fix for the Athena adapter.

### Problem

The outer retry decorator in `AthenaCursor.execute()` does not exclude `ICEBERG_FILESYSTEM_ERROR` from retries. When a transient error causes a retry of an Iceberg CTAS query, partial data is left at the same hardcoded S3 UUID location (Iceberg CTAS is non-atomic), and `delete_from_s3` is only invoked during Jinja evaluation (first attempt only). All subsequent retries therefore hit `ICEBERG_FILESYSTEM_ERROR: Cannot create a table on a non-empty location` — and since this error was not excluded from the outer retry, it gets retried up to `num_retries` times against the same S3 path, always failing.

The existing comment already explains why `TOO_MANY_OPEN_PARTITIONS` is excluded for the same reason, but `ICEBERG_FILESYSTEM_ERROR` was never added to the exclusion list.

Note: this is distinct from the `TOO_MANY_OPEN_PARTITIONS` → batch fallback path handled by `run_query_with_partitions_limit_catching`. That path already works correctly. This issue surfaces when a different transient error triggers the outer retry on an Iceberg CTAS.

### Solution

Add `ICEBERG_FILESYSTEM_ERROR` to the no-retry condition alongside `TOO_MANY_OPEN_PARTITIONS`. Since retrying the same SQL with the same S3 UUID location will always fail, failing fast avoids unnecessary retries and surfaces the original error more clearly.

### Checklist


iconara:
I believe this will be indirectly fixed by #1637 (but the change would still be good to apply to the legacy mode introduced by that PR). In the new retry logic, only ICEBERG_COMMIT_ERROR is retried, all other Iceberg errors are assumed to be non-retryable. Iceberg errors are also not retried by the throttling retry logic anymore.

dtaniwaki:
@iconara Thank you for the PR. I'll try it in my project 👍

dtaniwaki:
Superseded by #1637 (PyAthena→boto3 migration), now merged into `main`. The new boto3-based cursor only retries `ICEBERG_COMMIT_ERROR` (throttling API errors are handled separately), so `ICEBERG_FILESYSTEM_ERROR` / deterministic errors like query timeouts are no longer retried by design. Closing as no longer needed.

PR #1747, opened 2026-03-11
[Snowflake][BigQuery] Support packages for python UDFs

### LINKED ISSUE #1737: [UDFs] support pypi packages for python UDFs
### Housekeeping

- [x] I am a maintainer of dbt-core

### Short description

This dbt-adapters half of https://github.com/dbt-labs/dbt-core/issues/12041

Some python UDFs may want to use 3rd party PyPi packages - snowflake [docs](https://docs.snowflake.com/en/developer-guide/udf/python/udf-python-packages)

Packages for UDFs are specifiable in dbt-core already for functions. Now we just need to handle the templating that into the create statement for the function. That is if a python function specifies the packages `numpy`, `pandas`, and `xgboost==1.5.0`. Then we'd expect the templated statement to look something like 
```sql
CREATE OR REPLACE FUNCTION py_udf()
  RETURNS VARIANT
  LANGUAGE PYTHON
  RUNTIME_VERSION = 3.12
  PACKAGES = ('numpy','pandas','xgboost==1.5.0')
  HANDLER = 'udf'
AS $$
import numpy as np
import pandas as pd
import xgboost as xgb
def udf():
  return [np.__version__, pd.__version__, xgb.__version__]
$$;
```

### Acceptance criteria

- When`packages` are specified for python UDFs, they get templated appropriately

### Context

- See [internal slack thread](https://dbt-labs.slack.com/archives/CSSHB76QL/p1758142720330089?thread_ts=1758030491.041499&cid=CSSHB76QL)
- Bigquery [docs](https://docs.cloud.google.com/bigquery/docs/user-defined-functions-python#install_a_package_from_the_python_package_index)

QMalcolm:
resolves #1737

### Problem

Users could specify packages for python UDFs in core, but we weren't adding them to the template for the function to be created

### Solution

Add the packages to the template

### Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,171 @@
+# AGENTS.md — dbt Adapters Monorepo
+
+This file instructs AI coding agents on how to navigate, build, test, and contribute to this repository.
+
+## Repository Layout
+
+```
+dbt-adapters/
+├── dbt-adapters/          # Base framework (BaseAdapter, SQLAdapter, contracts, catalogs)
+├── dbt-tests-adapter/     # Shared test suite for all adapters
+├── dbt-postgres/          # PostgreSQL adapter
+├── dbt-redshift/          # Redshift adapter (extends dbt-postgres)
+├── dbt-snowflake/         # Snowflake adapter
+├── dbt-bigquery/          # BigQuery adapter
+├── dbt-spark/             # Spark / Databricks adapter
+└── dbt-athena/            # AWS Athena adapter
+```
+
+Each adapter is an independent Python package with its own `pyproject.toml`, `hatch.toml`, and test suite.
+
+## Environment Setup
+
+All commands are run from within the specific adapter's directory, not the repo root.
+
+```bash
+cd dbt-{adapter}        # e.g., cd dbt-redshift
+pip install hatch changie  # if not already installed; changie is needed for changelog entries
+hatch run setup            # installs adapter + deps in editable mode (includes pre-commit)
+```
+
+> **Note:** `pre-commit` is managed as a hatch environment dependency — you do not need to install it globally. `hatch run setup` installs it inside the hatch virtualenv and registers the git hooks. `changie` is a standalone tool used outside of hatch for changelog management and must be installed separately as shown above.
+
+To configure the virtual environment for IDE use:
+```bash
+hatch config set dirs.env.virtual .hatch
+```
+
+## Building / Code Quality
+
+```bash
+hatch run code-quality  # runs Black (format), Flake8 (lint), MyPy (types)
+```
+
+Always run this before submitting changes. Fix all errors before committing.
+
+## Testing
+
+### Unit Tests (no database required)
+
+```bash
+hatch run unit-tests
+
+# Run a specific test
+hatch run unit-tests -- tests/unit/test_file.py::ClassName::test_method -v
+```
+
+Unit tests live in `tests/unit/`. They test Python logic without a live database.
+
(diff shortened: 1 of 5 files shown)


dbt-bigquery/src/dbt/include/bigquery/macros/materializations/functions/scalar.sql
```diff
@@ -22,7 +22,7 @@
     CREATE OR REPLACE FUNCTION {{ target_relation.render() }} ({{ formatted_scalar_function_args_sql()}})
     RETURNS {{ model.returns.data_type }}
     LANGUAGE python
-    OPTIONS(runtime_version = "{{ 'python-' ~ model.config.get('runtime_version') }}", entry_point = "{{ model.config.get('entry_point') }}")
+    OPTIONS(runtime_version = "{{ 'python-' ~ model.config.get('runtime_version') }}", entry_point = "{{ model.config.get('entry_point') }}"{% if model.config.get('packages') %}, packages = [{% for pkg in model.config.get('packages') %}'{{ pkg }}'{% if not loop.last %}, {% endif %}{% endfor %}]{% endif %})
```
colin-k-rogers: should we use  the generic macro here `{{ get_function_python_packages_sql() }}`?

dbt-adapters/src/dbt/include/global_project/macros/materializations/functions/aggregate.sql
```diff
@@ -63,3 +64,12 @@
...
     HANDLER = '{{ model.config.get('entry_point') }}'
 {% endmacro %}
+
+--- function python packages (for PyPI dependencies) ---
+
+{% macro get_function_python_packages_sql() %}
+    {{ return(adapter.dispatch('get_function_python_packages_sql', 'dbt')()) }}
+{% endmacro %}
+
+{% macro default__get_function_python_packages_sql() %}
```
colin-k-rogers: should we raise an exception that this needs to be implemented?

dbt-bigquery/src/dbt/include/bigquery/macros/materializations/functions/scalar.sql
QMalcolm: Agreed. And we should maybe abstract the templating of `OPTIONS` for BigQuery into a sub macro as well

dbt-adapters/src/dbt/include/global_project/macros/materializations/functions/aggregate.sql
QMalcolm: Yeaaaaaa lets do that 👍🏻

QMalcolm: Instead of raising an exception, I realized we have a `default__get_function_python_options`, and moved it in there. Doing this is more in line with how we've built the infrastructure for functions overall

dbt-tests-adapter/src/dbt/tests/adapter/functions/files.py
```diff
@@ -148,3 +148,28 @@ def finish(self):
...
+    return float(np.sqrt(x))
+""".strip()
+
+MY_UDF_PYTHON_WITH_PACKAGES_YML = """
+functions:
+  - name: sqrt_input
+    description: Return the square root of the input using numpy
+    config:
+      entry_point: sqrt_input
+      runtime_version: "3.12"
```
xuliangs: @colin-rogers-dbt looks like changing this to 3.11 will fix the failed tests?

PR #1753, opened 2026-03-13
fix(bigquery): format expiration_timestamp as TIMESTAMP literal in materialized view OPTIONS

### LINKED ISSUE #1209: dbt-bigquery materialized view doesn't wrap `expiration_timestamp` in string
Hey folks 👋 I'm building a materialized view using the `dbt-bigquery` adapter, when I set `hours_to_expiration=24` in a `config()` block, the resulting `expiration_timestamp` is passed into the BigQuery configuration as a raw datetime object instead of a string as expected by the [materialized view options](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#materialized_view_option_list). 

Simple example:
```sql
{{
  config(
    materialized = 'materialized_view',
    hours_to_expiration = 24,
  )
}}

SELECT 1 AS howdy
```

Compiles to:
```sql

        create materialized view if not exists `my_project`.`my_schema`.`mv`
    
    
    OPTIONS(
      enable_refresh=True,
    
      refresh_interval_minutes=30,
    
      expiration_timestamp=2025-07-18 18:56:22.153305
    )
    as 

SELECT 1 AS howdy
```

With the syntax error:
```
Syntax error: Expected ")" or "," but got integer literal "18" at [11:39]
```

I suspect that this function may need to cast the `expiration_timestamp` to a string rather than an interval.

https://github.com/dbt-labs/dbt-adapters/blob/fed0e2e7a2e252175dcc9caccbdd91d354ac6a9d/dbt-bigquery/src/dbt/adapters/bigquery/relation_configs/_options.py#L59

mehdigati:
## Problem

When `hours_to_expiration` is set on a BigQuery materialized view, the `expiration_timestamp` is injected into the `OPTIONS()` clause as a raw Python `datetime` object:

```sql
OPTIONS(
  expiration_timestamp=2025-07-18 18:56:22.153305 
)
```

BigQuery rejects this with:
```
Syntax error: Expected ")" but got integer literal
```

## Fix

Added a dedicated `timestamp()` formatter that wraps the value in a proper BigQuery `TIMESTAMP` literal:

```sql
OPTIONS(
  expiration_timestamp=TIMESTAMP '2025-07-18 18:56:22.153305' 
)
```

## Testing

- Reproduced the error on a live BigQuery project before the fix
- Confirmed the materialized view creates successfully after the fix
- Added unit test `test_expiration_timestamp_formatted_as_bq_literal` verifying the output format

Fixes #1209


mehdigati:
@cla-bot check

PR #1756, opened 2026-03-14
fix(bigquery): handle 409 Conflict as success during job creation retries

### LINKED ISSUE #1746: [Bug] Handle 409 Conflict as success during BigQuery job creation retries

## Description

When dbt retries a BigQuery job creation (e.g. due to a timeout or transient network spike), BigQuery may return a `409 Conflict` response, indicating the job was already registered. dbt currently treats this as a fatal error and fails, instead of recognizing that the job was successfully submitted and continuing normally.

The root cause is not the timeout itself — subsequent runs complete without issues, suggesting this is an edge case where the job confirmation did not return to dbt in time. The real problem is that dbt's retry logic does not handle `409` gracefully.

## Actual Behavior

dbt raises an error on `409 Conflict`, causing the run (and any orchestration flow, e.g. Prefect) to fail.

### Expected Behavior

When dbt retries a job and receives `409 Conflict` from BigQuery, it should recognize that the job already exists and treat it as a successful submission — then continue polling for the result as normal.

### Steps To Reproduce

1. Run a dbt model against BigQuery
2. Simulate a transient delay (network spike, VM slowdown) that causes the job confirmation to not return within `job_creation_timeout_seconds`
3. dbt retries with the same job ID
4. BigQuery returns `409 Conflict`
5. dbt fails instead of continuing with the existing job

### Relevant log output

```shell
ValueError: Failures detected during invocation of dbt command 'test--selectpath:models/xxx/xxx':
Test relationships_context_arrival_map_site_id__site_id__ref_context_all_sites_ errored with message: "Database Error in test relationships_xxxxx_ (models/xxx/xxx/xxx.yaml)
  409 POST https://bigquery.googleapis.com/xxxx: Already Exists: Job xxx-xxx:US.xxxx-xxxx-xxxx-xxxx-xxxx
```

### Environment

```markdown
- OS: Debian 12
- Python: 3.12
- dbt: 1.10.3
```

bigquery

### Additional Context

_No response_

sunilp:
## Summary

When dbt retries a BigQuery job creation (e.g., due to a timeout or transient network spike), BigQuery may return a `409 Conflict` response indicating the job was already registered. dbt currently treats this as a fatal error.

This PR catches `google.api_core.exceptions.Conflict` during `client.query()` and recovers the existing job via `client.get_job(job_id)`, allowing the pipeline to continue normally.

**Change:** Wraps the `client.query()` call in a try/except for `Conflict`, recovering via `get_job()` — exactly as suggested by @dbeatty10 in #1746.

## Impact

- Prevents spurious pipeline failures during transient network delays
- No behavior change for successful job creation (the happy path is unchanged)
- Only affects the retry path when BigQuery confirms the job already exists

Fixes #1746

## Checklist

- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] This PR includes a bug fix
- [x] Changes are scoped to `dbt-bigquery` only


dbt-bigquery/src/dbt/adapters/bigquery/connections.py
```diff
@@ -611,13 +612,18 @@ def _query_and_results(
...
+                query=sql,
+                job_config=query_job_config,
+                job_id=job_id,
+                job_retry=None,
+                timeout=self._retry.create_job_creation_timeout(),
+            )
+        except Conflict:
+            logger.info("Caught 409 Conflict for job_id=%s, recovering via get_job()", job_id)
+            query_job = client.get_job(job_id)
+            logger.info("Job recovered, state: %s", query_job.state)
```
colin-k-rogers: I think it's safe to make this a debug as generally a user doesn't need to know about this:
```suggestion
            logger.debug("Job recovered, state: %s", query_job.state)
```

dbt-bigquery/src/dbt/adapters/bigquery/connections.py
```diff
@@ -611,13 +612,18 @@ def _query_and_results(
...
+        try:
+            query_job = client.query(
+                query=sql,
+                job_config=query_job_config,
+                job_id=job_id,
+                job_retry=None,
+                timeout=self._retry.create_job_creation_timeout(),
+            )
+        except Conflict:
+            logger.info("Caught 409 Conflict for job_id=%s, recovering via get_job()", job_id)
```
colin-k-rogers:
```suggestion
            logger.debug("Caught 409 Conflict for job_id=%s, recovering via get_job()", job_id)
```

dbt-bigquery/src/dbt/adapters/bigquery/connections.py
```diff
@@ -611,13 +612,18 @@ def _query_and_results(
...
+            query_job = client.get_job(job_id)
```
colin-k-rogers: does this include a sane default retry policy or should we supply one?

sunilp:
Pushed changes addressing all review feedback:

1. **`logger.info` → `logger.debug`** — done per @colin-rogers-dbt's suggestion
2. **try/except around `get_job()`** — wraps in try/except, raises `DbtRuntimeError` with original `Conflict` as `__cause__` so the root cause is preserved
3. **timeout on `get_job()`** — passes `self._retry.create_job_creation_timeout()` to stay consistent with `client.query()`. Left the default retry policy (which retries on transient errors) since `get_job` defaults to `DEFAULT_RETRY` — that's a sane default and disabling it with `retry=None` would be worse here
4. **Unit tests** — added two tests: successful 409 recovery (verifies `get_job` is called and recovered job is polled) and recovery failure (verifies `DbtRuntimeError` is raised with the original `Conflict` as cause)

Also added the changelog entry in the previous commit.

sunilp:
Rebased onto main to resolve the behind-branch status. CI needs workflow approval to re-run. @colin-rogers-dbt could you approve when you get a chance?

sunilp:
Hi @colin-rogers-dbt, just checking in. CI is green after the rebase. Let me know if there's anything else needed on this one.

colin-k-rogers:
Sorry @sunilp for the delay, running the CI now, only outstanding issues is the log message formatting which copilot identified

dnamufetha-oaknorth:
Hi @colin-rogers-dbt & @sunilp 

Thank you for working on this fix - this will really help our team aswell 🙌 

Quick question: is there a rough ETA for when this fix will release with `v1.12.0`?

I noticed that there are 3 other issues mentioned in [the milestone](https://github.com/dbt-labs/dbt-adapters/milestone/4), but I'm wondering if they are blockers for releasing `v1.12.0`.

Thanks.

sunilp:
Thanks @dnamufetha-oaknorth — glad it'll help your team too. Timing on `1.12.0` is the maintainers' call, but the fix itself is mergeable and CI was green on the last run.

@colin-rogers-dbt — I've just pushed [3d499821](https://github.com/dbt-labs/dbt-adapters/pull/1756/commits/3d499821) addressing the `AdapterLogger` format-placeholder nit you flagged on 2026-05-06 (the `%s` → `{}` substitution Copilot caught). That was the only outstanding item I had on my side. Ready for another look whenever you have a moment.

PR #1760, opened 2026-03-16
Fix nested STRUCT fields misassigning values when YAML order differs

colin-rogers-dbt:
## Changes

- Add `_parse_struct_fields()` function to parse BigQuery STRUCT type strings into field dictionaries
- Add `get_struct_select_expression()` function to generate explicit STRUCT SELECT expressions that preserve field order from DDL definitions
- Add `_build_struct_field_expressions()` helper to recursively construct nested STRUCT field expressions
- Export new function via adapter interface as `get_struct_select_expr()` macro
- Update `get_columns_spec_ddl.sql` macro to use explicit STRUCT field selection instead of plain column names
- Add comprehensive unit tests for struct field parsing and SELECT expression generation

## Problem

When YAML column definitions have fields in a different order than the actual table STRUCT definition, values would be misassigned due to positional alignment during inserts/updates.

## Solution

Generate explicit STRUCT constructors that name each field, ensuring positional alignment with the DDL definition regardless of YAML field order.

### DIFF
--- a/dbt-bigquery/src/dbt/adapters/bigquery/column.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/column.py
@@ -15,6 +15,7 @@ def _parse_struct_fields(data_type: str) -> Optional[List[Dict[str, str]]]:
 
     Returns None if the data_type is not a STRUCT or is malformed.
     Handles trailing constraints like `struct<x INT64> not null` correctly.
+    Handles field types with parenthesised parameters like NUMERIC(10, 2).
 
     Examples:
     >>> _parse_struct_fields("struct<x INT64, y STRING>")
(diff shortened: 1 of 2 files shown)


dbt-bigquery/src/dbt/include/bigquery/macros/utils/get_columns_spec_ddl.sql
```diff
@@ -32,6 +32,6 @@
     {%- for i in user_provided_columns %}
       {%- set col = user_provided_columns[i] -%}
       {%- set col_name = adapter.quote(col['name']) if col.get('quote') else col['name'] -%}
-      {{ col_name }}{{ ", " if not loop.last }}
+      {{ api.Column.get_struct_select_expression(col_name, col['data_type']) }}{{ ", " if not loop.last }}
     {%- endfor -%}
 {% endmacro %}
```
MichelleArk, line 37: Is `col['data_type']` being undefined at this point a valid case to consider? Presumably we'd just return the `col_name` directly if that was the case which should be pretty straightforward to plumb through.

colin-k-rogers: updated

MichelleArk (approved):

PR #1768, opened 2026-03-17
Fix dbt clone for Snowflake Iceberg tables

### LINKED ISSUE #1767: dbt clone fails on Snowflake Iceberg tables
## Problem

`dbt clone` fails when the source table is a Snowflake Iceberg table. Snowflake requires `CREATE ICEBERG TABLE ... CLONE ...` syntax for Iceberg tables, but the Snowflake adapter's `snowflake__create_or_replace_clone` macro always generates `CREATE [TRANSIENT] TABLE ... CLONE ...`.

## Expected behavior

`dbt clone` should automatically detect whether the source relation is an Iceberg table and use the correct DDL syntax, without requiring any additional user configuration.

## Proposed fix

The Snowflake adapter already tracks Iceberg status via `SnowflakeRelation.is_iceberg_format` (populated from Snowflake's `SHOW OBJECTS` metadata). The fix is to look up the source relation in the cache inside the clone macro and conditionally emit the `ICEBERG` keyword:

- Source is Iceberg → `CREATE OR REPLACE ICEBERG TABLE ... CLONE ...`
- Source is not Iceberg → `CREATE OR REPLACE [TRANSIENT] TABLE ... CLONE ...` (existing behavior)

The `transient` keyword is also correctly omitted for Iceberg clones since Iceberg tables cannot be transient.

jecolvin:
## Summary

- Detect Iceberg source tables in the clone macro via `load_cached_relation` and emit `CREATE OR REPLACE ICEBERG TABLE ... CLONE ...` instead of `CREATE OR REPLACE [TRANSIENT] TABLE ... CLONE ...`
- Omit the `transient` keyword for Iceberg clones (Iceberg tables cannot be transient)
- Add functional test for cloning an Iceberg table

Resolves #1767

## Test plan

- [ ] Existing clone tests pass (`TestSnowflakeClonePossible`, `TestSnowflakeCloneTrainsentTable`)
- [ ] New `TestSnowflakeCloneIcebergTable` test passes against a Snowflake account with an Iceberg-enabled external volume
- [ ] Verify generated SQL with `--log-level debug` shows `CREATE OR REPLACE ICEBERG TABLE` for Iceberg sources and `CREATE OR REPLACE TRANSIENT TABLE` for regular sources

### DIFF
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/materializations/clone.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/materializations/clone.sql
@@ -3,8 +3,14 @@
 {% endmacro %}
 
 {% macro snowflake__create_or_replace_clone(this_relation, defer_relation) %}
+    {%- set source_relation = load_cached_relation(defer_relation) -%}
+    {%- set is_iceberg = source_relation is not none and source_relation.is_iceberg_format -%}
     create or replace
-      {{ "transient" if config.get("transient", true) }}
+      {% if is_iceberg -%}
+        iceberg
+      {%- else -%}
+        {{ "transient" if config.get("transient", true) }}
+      {%- endif %}
       table {{ this_relation }}
       clone {{ defer_relation }}
       {{ "copy grants" if config.get("copy_grants", false) }}
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/materializations/clone.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/materializations/clone.sql
@@ -24,11 +24,13 @@
 
     {#- Cache miss: query Snowflake directly -#}
     {%- set show_sql -%}
-        show tables like '{{ relation.identifier }}' in {{ relation.database }}.{{ relation.schema }}
+        show tables like '{{ relation.identifier }}' in schema {{ relation.database }}.{{ relation.schema }}
     {%- endset -%}
     {%- set results = run_query(show_sql) -%}
-    {%- if results and results | length > 0 -%}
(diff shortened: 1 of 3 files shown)


jecolvin:
@colin-rogers-dbt Pushed a fix for the failing integration test!

jecolvin:
@colin-rogers-dbt I've addressed the latest round of feedback!

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/clone.sql
```diff
@@ -3,9 +3,33 @@
...
       {{ "copy grants" if config.get("copy_grants", false) }}
 {% endmacro %}
+
+{% macro snowflake__is_iceberg_table(relation) %}
+    {%- set source_relation = load_cached_relation(relation) -%}
+    {%- if source_relation is not none -%}
+        {{ return(source_relation.is_iceberg_format) }}
+    {%- endif -%}
+
+    {#- Cache miss: query Snowflake directly -#}
```
colin-k-rogers: I don't think we'll need this as the cache should already have the relation but if we do need this then we should be able to use `adapter.get_relation` instead of a custom show command

jecolvin:
@colin-rogers-dbt Pushed more fixes based on the comments!

jecolvin:
@colin-rogers-dbt Just refreshed the branch so it's not stale. Could I get another run please?

jecolvin:
@colin-rogers-dbt Found the problem with the failing test. Verified in my Snowflake account the full set of clone tests are green now!

PR #1781, opened 2026-03-19
dbt-bigquery adapter: snapshot fix for count mismatch and invalid casting

### LINKED ISSUE #1702: [Bug] Snapshot with `hard_deletes:new_record` erroring with new column (incompatible types)

### Current Behavior

I can see some other bugs around snapshots and `hard_deletes:new_record` but I can't find this specific issue:

I'm adding a new column to an ephemeral model that is feeding a yml defined snpashot. (where the snapshot has already been run).
On running the snapshot, it fails with a type mismatch error in the `UNION ALL`

```
Database Error in snapshot snapshot_monzo_change_log_change_id
Column 4 in UNION ALL has incompatible types: STRING, STRING, STRING, INT64 at [238:5]
```

It looks like this occurs du to the `deletion_records` CTE:

```
deletion_records as (
    select 'insert' as dbt_change_type,
      snapshotted_data.airtable_id, 
      snapshotted_data.change_id, 
      NULL as proposal_name,  -- New column doesn't exist in existing table
      snapshotted_data.dbt_unique_key as dbt_unique_key,
      ...
)
```

### Expected Behavior

Expected that the new column will be added to the snapshot.

I suspect that the type of the column needs to be inferred from the newly added column and the CTE look something like 
```
deletion_records as (
    select 'insert' as dbt_change_type,
      snapshotted_data.airtable_id, 
      snapshotted_data.change_id, 
      CAST(NULL AS STRING) as proposal_name,  -- New column doesn't exist in existing table
      snapshotted_data.dbt_unique_key as dbt_unique_key,
      ...
)
```

Possibly this is a bigquery specific issue, but I'm not totally sure.

### Steps To Reproduce

Create ephemeral model:
```
{{
  config(materialized="ephemeral")
}}

SELECT
  id AS airtable_id,
  change_id
FROM {{ source('some_source', 'some_table') }}
```
Create yml snapshot based on model

```
version: 2
snapshots:
  - name: snapshot_test
    relation: ref('eph_model')
    config:
      strategy: check
      unique_key: airtable_id
      check_cols: all
      hard_deletes: new_record
```

Run the snapshot
`dbt snapshot --select snapshot_test`

Add new column to the ephemeral model:
```
SELECT
  id AS airtable_id,
  change_id,
  proposal_name  -- New column added
FROM {{ source('some_source', 'some_table') }}
```

Run the snapshot again, and the above error is seen.

### Relevant log output

```shell
Database Error in snapshot snapshot_my_table (snapshots/snapshot_my_table.yml)
  Column 4 in UNION ALL has incompatible types: STRING, STRING, STRING, INT64 at [238:5]
1 of 1 ERROR snapshotting dev_db.dbt_dev.snapshot_my_table [ERROR in 5.77s]



create or replace table `prod`.`snapshot_my_table__dbt_tmp`
OPTIONS(
    expiration_timestamp=TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL 12 hour)
)
as (
    with snapshot_query as (
        with __dbt__cte__eph_my_table as (
            SELECT
              id AS airtable_id,
              change_id,
              proposal_name  -- NEW COLUMN
            FROM `prod`.`my_source_table`
        )
        select * from __dbt__cte__eph_my_table
    ),

    snapshotted_data as (
        select *,
        airtable_id as dbt_unique_key
        from `prod`.`snapshot_my_table`
        where valid_to_timestamp is null
        -- NOTE: This table does NOT have the 'proposal_name' column yet
    ),

    insertions_source_data as (
        select *,
        airtable_id as dbt_unique_key,
        current_timestamp() as _meta_insertion_timestamp,
        current_timestamp() as valid_from_timestamp,
        coalesce(nullif(current_timestamp(), current_timestamp()), null) as valid_to_timestamp,
        to_hex(md5(concat(coalesce(cast(airtable_id as string), ''), '|',coalesce(cast(current_timestamp() as string), '')))) as episode_id
        from snapshot_query
    ),

    updates_source_data as (
        select *,
        airtable_id as dbt_unique_key,
        current_timestamp() as _meta_insertion_timestamp,
        current_timestamp() as valid_from_timestamp,
        current_timestamp() as valid_to_timestamp
        from snapshot_query
    ),

    deletes_source_data as (
        select *,
        airtable_id as dbt_unique_key
        from snapshot_query
    ),

    insertions as (
        select
            'insert' as dbt_change_type,
            source_data.*,
            'False' as is_deleted
        from insertions_source_data as source_data
        left outer join snapshotted_data
            on snapshotted_data.dbt_unique_key = source_data.dbt_unique_key
        where snapshotted_data.dbt_unique_key is null
           or (snapshotted_data.dbt_unique_key is not null and (TRUE or snapshotted_data.is_deleted = 'True'))
    ),

    updates as (
        select
            'update' as dbt_change_type,
            source_data.*,
            snapshotted_data.episode_id,
            snapshotted_data.is_deleted
        from updates_source_data as source_data
        join snapshotted_data
            on snapshotted_data.dbt_unique_key = source_data.dbt_unique_key
        where (TRUE or snapshotted_data.is_deleted = 'True')
    ),

    deletes as (
        select
            'delete' as dbt_change_type,
            source_data.*,
            current_timestamp() as valid_from_timestamp,
            current_timestamp() as _meta_insertion_timestamp,
            current_timestamp() as valid_to_timestamp,
            snapshotted_data.episode_id,
            snapshotted_data.is_deleted
        from snapshotted_data
        left join deletes_source_data as source_data
            on snapshotted_data.dbt_unique_key = source_data.dbt_unique_key
        where source_data.dbt_unique_key is null
          and not (snapshotted_data.is_deleted = 'True' and snapshotted_data.valid_to_timestamp is null)
    ),

    deletion_records as (
        select
            'insert' as dbt_change_type,
            snapshotted_data.`airtable_id`,
            snapshotted_data.`change_id`,
            NULL as `proposal_name`,  -- PROBLEM: Tries to add this column but snapshotted_data doesn't have it
            snapshotted_data.dbt_unique_key as dbt_unique_key,
            current_timestamp() as valid_from_timestamp,
            current_timestamp() as _meta_insertion_timestamp,
            snapshotted_data.valid_to_timestamp as valid_to_timestamp,
            to_hex(md5(concat(coalesce(cast(episode_id as string), ''), '|',coalesce(cast(current_timestamp() as string), '')))) as episode_id,
            'True' as is_deleted
        from snapshotted_data
        left join deletes_source_data as source_data
            on snapshotted_data.dbt_unique_key = source_data.dbt_unique_key
        where source_data.dbt_unique_key is null
          and not (snapshotted_data.is_deleted = 'True' and snapshotted_data.valid_to_timestamp is null)
    )

    -- LINE 238: UNION ALL fails here due to column type mismatch
    select * from insertions
    union all
    select * from updates
    union all
    select * from deletes
    union all
    select * from deletion_records
);
```

### Environment

```markdown
OS: Debian GNU/Linux 12 
Python: 3.12.12 
dbt-core: 1.10.13 
Database adapter: dbt-bigquery 1.10.2
```

bigquery

### Additional Context

_No response_

antoniabadarau:
Resolves #1702 

Please note that this PR, isolates the fix to the Bigquery adapter instead of modifying the default behaviour. If maintainers consider that this to be feasible as the default behaviour, I am happy to iterate and make these changes.

### Problem
Resolves the following issues with snapshots in BigQuery:
1. Queries in UNION ALL have mismatched column count
2. UNION ALL has incompatible types: ARRAY<STRING>, ARRAY<STRING>, ARRAY<STRING>, INT64

---

Both of these are caused by the default behaviour of snapshot_staging_table macro within the deletion_records CTE:

#### 1. Queries in UNION ALL have mismatched column count
- [get_column_schema_from_query](https://github.com/dbt-labs/dbt-adapters/blob/b5dc584eba15e8aed402a125f5bb3238fc2b190f/dbt-bigquery/src/dbt/adapters/bigquery/impl.py#L510) flattens RECORD columns into their subfields causing the unexpected increase in the column number
  - `deletion_records` CTE then loops over the flatten columns and adds them to the select list
  ```sql
    {% set source_sql_cols = get_column_schema_from_query(source_sql) %}
    ...
    select
    'insert' as dbt_change_type,
    {%- for col in source_sql_cols -%}
	    {%- if col.name in snapshotted_cols -%}
		    snapshotted_data.{{ adapter.quote(col.column) }},
	    {%- else -%}
		    NULL as {{ adapter.quote(col.column) }},
	    {%- endif -%}
    {% endfor -%}
  ```
- looking at the other all the other CTEs (insertions, updates, deletes), these simply use `source_data.*` which keep the original structure of the columns without the flattening
  ```sql
   insertions as (
    select
        'insert' as dbt_change_type,
        source_data.*
    ...
    ),
    updates as (
	      select
	          'update' as dbt_change_type,
	          source_data.*,
	      ...
    ),
    deletes as (
        select
            'delete' as dbt_change_type,
            source_data.*,
		    ...
    ),
    ...
   ```
- this is then causing the following UNION operation to fail given than the `deletion_records` has more columns than the others
   ```sql
   select * from insertions
    union all
    select * from updates
    union all
    select * from deletes
    union all
    select * from deletion_records
   ```

#### 2. UNION ALL has incompatible types: ARRAY<STRING>, ARRAY<STRING>, ARRAY<STRING>, INT64
- when a column doesn't yet exist in the snapshot table (new column), the macro emits a bare NULL - in BigQuery, an uncast NULL is inferred as INT64 which leads to the errors we’re seeing
```sql
{%- if col.name in snapshotted_cols -%}
      snapshotted_data.{{ adapter.quote(col.column) }},
  {%- else -%}
      NULL as {{ adapter.quote(col.column) }}, # invalid NULL
{%- endif -%}
```
- when the UNION ALL combines this with the same column from other CTEs (insertions, updates, deletes) where the column has its actual type, BQ fails because INT64 is incompatible with ARRAY
```
Queries in UNION ALL have mismatched column count; query 1 has 14 columns, query 4 has 15 columns at [307:5]
```

#### 1. Queries in UNION ALL have mismatched column count
- in the `deletion_records` CTE, make use of [get_columns_in_query](https://github.com/dbt-labs/dbt-adapters/blob/b5dc584eba15e8aed402a125f5bb3238fc2b190f/dbt-adapters/src/dbt/include/global_project/macros/adapters/columns.sql#L85) rather than [get_column_schema_from_query](https://github.com/dbt-labs/dbt-adapters/blob/b5dc584eba15e8aed402a125f5bb3238fc2b190f/dbt-bigquery/src/dbt/adapters/bigquery/impl.py#L510)
- this will return the top-level column names only (witch matches the insertions, updates etc CTEs) rather then the flattened structures
```sql
{% set source_col_names = get_columns_in_query(source_sql) %}
	{%- for col_name in source_col_names -%}
	  snapshotted_data.{{ adapter.quote(col_name) }}
	  ...
{% endfor -%}
```

#### 2. UNION ALL has incompatible types: ARRAY<STRING>, ARRAY<STRING>, ARRAY<STRING>, INT64
- instead of emitting bare NULL reference `source_data.col` from the existing LEFT JOIN with `deletes_source_data`
- given that the join filters on dbt_unique_key being NULL, all `deletes_source_data` columns as NULL but BQ will preserve the schema and infer the correct data type
```sql
{%- else -%}
  source_data.{{ adapter.quote(col_name) }},
{%- endif -%}
```

### Checklist


dbeatty10:
Thanks for opening this @antoniabadarau! 🤩

> Please note that this PR, isolates the fix to the Bigquery adapter instead of modifying the default behaviour. If maintainers consider that this to be feasible as the default behaviour, I am happy to iterate and make these changes.

I tried applying your solution to be the default solution for all adapters in https://github.com/dbt-labs/dbt-adapters/pull/1879, and it passed the tests for Snowflake, BigQuery, Redshift, and Postgres.

So this looks feasible as the default -- could you move your solution into `dbt-adapters/src/dbt/include/global_project/macros/materializations/snapshots/helpers.sql`?

Then we can queue this up for our engineering team for review.

antoniabadarau:
Absolutely @dbeatty10, I will look at this next week. 

Thank you so much for looking into this and testing the solution across all adapters ☺️!

antoniabadarau:
This is now done @dbeatty10 - I moved everything to global and kept a couple of tests explicitly on BigQuery as this was impacted by the bug previously.

Making this global also helps with the review process hopefully as the changes become a lot clearer!

Please have a read through and let me know your thoughts 🙌

antoniabadarau:
Just as a heads up @dbeatty10 - I have a similar PR open to address related snapshotting issues in the BigQuery adapter https://github.com/dbt-labs/dbt-adapters/pull/1785.
This resolves how we detected changes to snapshotted data given that REPEATED types do not support equality comparison.

If you have the time to have a look at this as well, I would greatly appreciate it. Thank you!!! ☺️

antoniabadarau:
@dbeatty10 can I please check if there are any updates on this from your side? 🙏

antoniabadarau:
Good morning @dbeatty10! Is this something that can be looked at? Alternatively, is there somewhere where I can raise / bump the priority of this? Thank you!

dbt-bigquery/tests/functional/adapter/test_snapshot_struct_array.py
```diff
@@ -0,0 +1,188 @@
+"""
```
tauhid621, line 1: Since we are making global change, let's add the `else` block (source_data.{{ adapter.quote(col_name) }},) test to the base adapter tests as well https://github.com/dbt-labs/dbt-adapters/tree/main/dbt-tests-adapter/src/dbt/tests/adapter/simple_snapshot

tauhid621 (commented):
Thanks for your contribution. Changes look good. Perhaps we can add the new else scenario test to the base adapter for global coverage.

dbt-bigquery/tests/functional/adapter/test_snapshot_struct_array.py
dbeatty10: ### Good news
I believe that [`BaseSnapshotNewColumnTimestampStrategy` within `dbt-tests-adapter/src/dbt/tests/adapter/simple_snapshot/test_ephemeral_snapshot_hard_deletes.py`](https://github.com/dbt-labs/dbt-adapters/blob/e008d8f6f7a7d76a6a63084512f4862f9bd41090/dbt-tests-adapter/src/dbt/tests/adapter/simple_snapshot/test_ephemeral_snapshot_hard_deletes.py#L150-L194) already handles the `else` case in the base adapter (added in https://github.com/dbt-labs/dbt-adapters/pull/1810).

### More info
The solution in this PR is in the base adapter, and tests added in https://github.com/dbt-labs/dbt-adapters/pull/1810 are still in-place to make sure the behavior is working across adapters.

This particular test can't be added to the base adapter tests as-is because it uses bigquery-specific syntax on line 27 defining a `struct`. This seems appropriate because the only remaining known bug is bigquery-specific (specifically their `struct` and `arrays`).

tauhid621: This looks good then ❤️

tauhid621 (approved):

dbeatty10:
Thanks for sticking with this @antoniabadarau! Appreciate you finding a fix for this and doing the work to promote it all the way to the global implementation 🏆

antoniabadarau:
Thank you both! @dbeatty10 @tauhid621

PR #1788, opened 2026-03-20
Add refresh_warehouse config for Snowflake dynamic tables

### LINKED ISSUE #1193: [Feature] Support second warehouse config for Snowflake dynamic tables

### Describe the feature

This was raised by an enterprise dbt client to request support for configuring an additional Snowflake warehouse for dynamic tables. The request is to basically have the ability to set a warehouse for dynamic table DDL that is different to the warehouse that is used to execute the create statement. At the moment - how it appears we have coded it is that:
1. The SQL statement that creates the dynamic table.
2. The dynamic table doing dynamic table things.
Will both be using the same snowflake_warehouse config. Sharing the user's feedback and insights below:

> The specification for the dynamic table refresh warehouse is here:
> [Snowflake configurations | dbt Developer Hub](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#dynamic-tables)
> However, you can see it shares the same config as the overarching warehouse specification here:
> [Snowflake configurations | dbt Developer Hub](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#configuring-virtual-warehouses)
> So, think of it like this I want dbt to run code like this:

```
USE WAREHOUSE my_xlarge_dbt_warehouse;

CREATE OR REPLACE DYNAMIC TABLE my_dynamic_table

TARGET_LAG = ‘1 HOUR’

WAREHOUSE = my_xsmall_dynamic_table_refresh_warehouse

AS
SELECT * FROM staging_table;
```

 

> So you see how there are two warehouses involved here: one that dbt uses to execute the CREATE DYNAMIC TABLE statement, and another one that the dynamic table will use each out to perform its self-refresh.   In the current design of dbt, you cannot specify these two separately.

 

> So what is happening to us is that dbt actually runs like this:

 
```
USE WAREHOUSE my_xsmall_dynamic_table_refresh_warehouse;

CREATE OR REPLACE DYNAMIC TABLE my_dynamic_table

  TARGET_LAG = ‘1 HOUR’

  WAREHOUSE = my_xsmall_dynamic_table_refresh_warehouse

  AS

    SELECT * FROM staging_table

;
```

 

> You see by setting `snowflake_warehouse= ‘my_xsmall_dynamic_table_refresh_warehouse’` in the dynamic table model it ALSO CHANGES the warehouse that dbt uses to run the CREATE command.  Now in our case it’s not true that the my_xsmall_dynamic_table_refresh_warehouse even has permission to create dynamic tables in this database.

_No response_

### Who will this benefit?

All users who would like to set a separate Snowflake warehouse for the DDL step of dynamic tables.

_No response_

### Anything else?

_No response_

colin-rogers-dbt:
## Summary

Closes #1193

Adds a new optional `refresh_warehouse` config parameter that allows users to specify a separate warehouse for the dynamic table's **self-refresh** operations, independent of `snowflake_warehouse` which continues to control DDL execution (via the existing `pre_model_hook` mechanism).

**When only `snowflake_warehouse` is set** (existing behavior, no change):
```sql
CREATE DYNAMIC TABLE my_dynamic_table
    TARGET_LAG = '1 HOUR'
    WAREHOUSE = my_warehouse   -- snowflake_warehouse serves both roles
    AS SELECT * FROM staging_table
```

**When `refresh_warehouse` is also set** (new behavior):
```sql
-- pre_model_hook switches to snowflake_warehouse before executing DDL
CREATE DYNAMIC TABLE my_dynamic_table
    TARGET_LAG = '1 HOUR'
    WAREHOUSE = my_small_refresh_wh   -- refresh_warehouse
    AS SELECT * FROM staging_table
```

**Example model config:**
```yaml
models:
  - name: my_dynamic_table
    config:
      materialized: dynamic_table
      target_lag: "1 hour"
      snowflake_warehouse: my_large_execution_wh  # executes the DDL
      refresh_warehouse: my_small_refresh_wh      # WAREHOUSE = in DDL (new)
```

## Changes

- **`impl.py`**: Added `refresh_warehouse` field to `SnowflakeConfig` dataclass
- **`dynamic_table.py`**: Added `refresh_warehouse` optional field to `SnowflakeDynamicTableConfig`; added `ddl_warehouse` property (`refresh_warehouse or snowflake_warehouse`); wired through `from_dict()` and `parse_relation_config()`
- **`relation.py`**: ALTER changeset comparison now uses `ddl_warehouse` (so changing `refresh_warehouse` triggers `ALTER DYNAMIC TABLE SET WAREHOUSE`)
- **`create.sql`**: Both info schema and iceberg CREATE macros now use `dynamic_table.ddl_warehouse` for the `WAREHOUSE =` clause
- **`test_dynamic_table_config.py`**: Added `TestRefreshWarehouseOptional` and `TestRefreshWarehouseChangeset` test classes covering the new field and `ddl_warehouse` property

## Test plan

- [x] 181 unit tests pass (`hatch run unit-tests`)
- [x] Code quality clean (`hatch run code-quality` — black, flake8, mypy all pass)
- [ ] Integration test: model with `refresh_warehouse` set uses it in the `WAREHOUSE =` clause while `snowflake_warehouse` continues to control DDL execution
- [ ] Regression: model without `refresh_warehouse` produces identical SQL to before

### DIFF
--- a/dbt-redshift/.changes/unreleased/Under the Hood-20260327-190557.yaml
+++ b/dbt-redshift/.changes/unreleased/Under the Hood-20260327-190557.yaml
@@ -1,0 +1,6 @@
+kind: Under the Hood
+body: Add datasharing profile variant tests for incremental on_schema_change, incremental unique_key, materialized view basic, and UDFs to catch metadata regressions when SHOW APIs are enabled
+time: 2026-03-27T19:05:57.000000+00:00
+custom:
+    Author: colinrogers99
+    Issue: "1808"
--- a/dbt-redshift/tests/functional/adapter/catalog_tests/test_get_catalog.py
+++ b/dbt-redshift/tests/functional/adapter/catalog_tests/test_get_catalog.py
@@ -147,10 +147,15 @@ class TestGetCatalogDatasharing(TestGetCatalog):
     """Same catalog tests but with datasharing enabled."""
 
     @pytest.fixture(scope="class")
-    def profiles_config_update(self, dbt_profile_target):
-        outputs = {"default": dbt_profile_target}
-        outputs["default"]["datasharing"] = True
-        return outputs
+    def profiles_config_update(self, dbt_profile_target, unique_schema):
+        return {
+            "test": {
+                "outputs": {
+                    "default": {**dbt_profile_target, "schema": unique_schema, "datasharing": True}
+                },
+                "target": "default",
+            }
+        }
 
     def test_get_one_catalog_by_relations(
         self,
--- a/dbt-redshift/tests/functional/adapter/catalog_tests/test_relation_types.py
+++ b/dbt-redshift/tests/functional/adapter/catalog_tests/test_relation_types.py
@@ -48,7 +48,12 @@ class TestCatalogRelationTypesDatasharing(TestCatalogRelationTypes):
     """Same relation type tests but with datasharing config enabled."""
 
     @pytest.fixture(scope="class")
-    def profiles_config_update(self, dbt_profile_target):
-        outputs = {"default": dbt_profile_target}
-        outputs["default"]["datasharing"] = True
-        return outputs
+    def profiles_config_update(self, dbt_profile_target, unique_schema):
+        return {
+            "test": {
+                "outputs": {
+                    "default": {**dbt_profile_target, "schema": unique_schema, "datasharing": True}
+                },
+                "target": "default",
+            }
+        }
--- a/dbt-redshift/tests/functional/adapter/incremental/test_incremental_on_schema_change.py
+++ b/dbt-redshift/tests/functional/adapter/incremental/test_incremental_on_schema_change.py
@@ -10,6 +10,24 @@ class TestIncrementalOnSchemaChange(BaseIncrementalOnSchemaChange):
     pass
 
 
+class TestIncrementalOnSchemaChangeWithDatasharing(TestIncrementalOnSchemaChange):
+    """Same incremental on_schema_change tests with datasharing enabled.
+
+    Exercises get_columns_in_relation which uses SHOW COLUMNS in datasharing mode.
+    """
+
+    @pytest.fixture(scope="class")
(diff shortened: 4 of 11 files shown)


MazrimT:
I was just reading some issues for fun, I think this change gets the requested feature backwards.

I think the person writing the original issue just wants to be able to set the  warehouse used to refresh the table separately from the warehouse used to create it.

The only change that this would need is to have a new input value:
```yaml
models:
  - name: [<model-name>]
    config:
      materialized: dynamic_table
      on_configuration_change: apply | continue | fail
      target_lag: downstream | <time-delta>
     # existing snowflake_warehouse should be used as normal to do any changes to the table
      snowflake_warehouse: <warehouse-name>
      refresh_mode: AUTO | FULL | INCREMENTAL 
      initialize: ON_CREATE | ON_SCHEDULE
      # new optional value for refresh_warehouse. if set this goes to the `CREATE DYNAMIC TABLE` DDL as WAREHOUSE else default to the standard snowflake_warehouse
      refresh_warehouse: <warehouse-name>
```

colin-k-rogers:
> I was just reading some issues for fun, I think this change gets the requested feature backwards.
> 
> I think the person writing the original issue just wants to be able to set the warehouse used to refresh the table separately from the warehouse used to create it.

Good thinking @MazrimT, thanks!

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -77,6 +79,16 @@ class SnowflakeDynamicTableConfig(SnowflakeRelationConfigBase):
     immutable_where: Optional[str] = None
     transient: Optional[bool] = None
 
+    @property
+    def ddl_warehouse(self) -> str:
```
igorbelianski-cyber: it took me a minute to realize   ddl_warehouse here means "warehouse references as a ddl param"  as opposed to " a warehouse used to execute DDL" 

may be a more explicit name something like "dt_warehouse_parameter"

PR #1797, opened 2026-03-23
feat(athena): add target partition filter in batch_iceberg_merge

### LINKED ISSUE #1801: [Feature] Add automatic target partition filtering in batch_iceberg_merge to enable partition pruning

### Describe the feature

In `batch_iceberg_merge`, each batch filters the **source** by partition conditions (`WHERE batch`), but the **target** table has no corresponding filter in the MERGE ON clause. This means Athena cannot leverage Iceberg partition pruning on the target side, resulting in a full target table scan for every batch.

For example, with `partitioned_by=['DAY(date_column)']` and 365 daily partitions, `get_partition_batches` groups them into batches of up to `partitions_limit` (default 100). Each batch still performs a full target scan, even though only a subset of partitions is relevant.

**Proposed config:**

```yaml
{{ config(
    materialized='incremental',
    incremental_strategy='merge',
    table_type='iceberg',
    partitioned_by=["DAY(date_column)"],
    unique_key='id',
    merge_with_target_partition_filter=true,
) }}
```

**Behavior:**

- Default: `false` (backward compatible, no change to existing behavior)
- When `true`: automatically derive target partition filters from `partitioned_by` config and inject them into the MERGE ON clause during batch merge
- When `incremental_predicates` is also set: skip target partition filter injection (user's explicit predicates take precedence)
- Bucket partitions: skip due to substring replacement complexity

**Generated SQL example:**

```sql
-- Current (merge_with_target_partition_filter=false or unset): no target partition filter → full target scan per batch
MERGE INTO target AS target
USING (
    SELECT ... FROM tmp
    WHERE date_trunc('day', date_column) = DATE '2024-01-01'
       or date_trunc('day', date_column) = DATE '2024-01-02'
) AS src
ON (target.id = src.id)

-- Proposed (merge_with_target_partition_filter=true): target partition filter enables Iceberg pruning
MERGE INTO target AS target
USING (
    SELECT ... FROM tmp
    WHERE date_trunc('day', date_column) = DATE '2024-01-01'
       or date_trunc('day', date_column) = DATE '2024-01-02'
) AS src
ON (
    target.id = src.id
    AND (date_trunc('day', target.date_column) = DATE '2024-01-01'
      or date_trunc('day', target.date_column) = DATE '2024-01-02')
)
```

**Implementation approach:**

Rather than post-hoc string replacement on batch conditions, generate target conditions at the source — in `get_partition_batches`, produce `(source_condition, target_condition)` pairs by applying `target.` prefix to column names during condition construction. This avoids fragile regex/string substitution on expressions like `date_trunc('day', date_column)`.

**Scope:**
- Batch merge only — single merge is unaffected
- Opt-in via `merge_with_target_partition_filter=true`
- Disabled when `incremental_predicates` is set (user's explicit config takes precedence)
- Skipped when bucket partitions are present

  Users can manually set incremental_predicates to add target-side filters, but this requires per-model configuration and duplicates information already available in partitioned_by. Also, `incremental_predicates` is static and cannot dynamically adapt to batch-specific partition values.

### Who will this benefit?

Any dbt-athena user running Iceberg incremental models with merge strategy and batch processing. The performance improvement scales with the number of partitions — larger tables benefit most.

yes

### Anything else?

_No response_

bernard-choi:
resolves #1801

### Problem

In `batch_iceberg_merge`, each batch filters the **source** by partition conditions (`WHERE batch`), but the **target** table has no corresponding filter in the MERGE ON clause. This means Athena cannot leverage Iceberg partition pruning on the target side, resulting in a full target table scan for every batch.

For example, with 365 daily partitions and `partitions_limit=100` (4 batches), the target table is fully scanned 4 times.

### Solution

Add an opt-in `merge_with_target_partition_filter` config option (default `false`) that automatically derives target-side partition filters from `partitioned_by` and injects them into the MERGE ON clause during batch merge.

```yaml
{{ config(
    materialized='incremental',
    incremental_strategy='merge',
    table_type='iceberg',
    partitioned_by=["DAY(date_column)"],
    unique_key='id',
    merge_with_target_partition_filter=true,
) }}
```

**Generated SQL comparison:**

```sql
-- Before (full target scan per batch)
MERGE INTO target USING (SELECT ... FROM tmp WHERE batch_filter) AS src
ON (target.id = src.id)

-- After (target partition pruning enabled)
MERGE INTO target USING (SELECT ... FROM tmp WHERE batch_filter) AS src
ON (target.id = src.id
    AND (date_trunc('day', target.date_column) = DATE'2024-01-01'
      OR date_trunc('day', target.date_column) = DATE'2024-01-02'))
```

<details>
<summary>Why the ON clause enables partition pruning</summary>

MERGE has no WHERE clause, so the ON clause is the only place to attach conditions to the target table. When a predicate in the ON clause references **only target columns** (not src), Trino's optimizer can push it down as a scan-level filter before the join executes. For Iceberg tables, this pushed-down predicate is matched against partition metadata, allowing the engine to skip reading files from non-matching partitions entirely.

In short: `target.date_column = DATE'2024-01-01'` in the ON clause behaves the same as a `WHERE target.date_column = DATE'2024-01-01'` would in a regular query — it triggers Iceberg partition pruning.

</details>

<details>
<summary>Implementation details</summary>

- Split `merge_part` into `merge_on_part` + `merge_actions_part` to inject target filter between the ON clause and WHEN clauses. The internal macro `batch_iceberg_merge` signature changed accordingly (`merge_part` → `merge_on_part, merge_actions_part`). This macro is not dispatched and is only called within the same file, so end users are not affected.
- Added `format_one_partition_key_with_prefix()` to `impl.py` for generating `target.`-prefixed partition conditions, including hidden partition transforms (`DAY()`, `MONTH()`, etc.).
- Extended `get_partition_batches` with `build_target_conditions` parameter — when `True`, returns `{'source': [...], 'target': [...]}` dict instead of a plain list. Existing callers are unaffected (default `False`).
- Bucket partitions are automatically excluded from target filtering because bucket assignment uses murmur3 hashing, which cannot be expressed as a SQL predicate on the original column.
- Coexists with `incremental_predicates` — the original issue (#1801) proposed disabling target filter when predicates are set, but these serve orthogonal purposes (static business logic vs dynamic batch optimization), so both apply together.
- **Caveat:** When a row's partition key value changes between source and target (e.g., `date_column` moves from `2024-01-01` to `2024-01-03`), the target filter will not find the existing row → INSERT instead of UPDATE. Users should only enable this for models where partition keys are immutable (e.g., `event_date`, `created_at`).

</details>

### Checklist

### DIFF
--- a/dbt-athena/tests/functional/adapter/test_incremental_iceberg.py
+++ b/dbt-athena/tests/functional/adapter/test_incremental_iceberg.py
@@ -365,7 +365,7 @@ class TestIcebergUpdateCondition:
         check_relations_equal(project.adapter, [relation_name, expected_seed_name])
 
 
-### Target partition filter tests ###
+# Target partition filter tests
 
 models__target_partition_filter_sql = """
 {{ config(
@@ -494,7 +494,9 @@ class TestIcebergTargetPartitionFilter:
 
     @pytest.fixture(scope="class")
     def seeds(self):
-        return {"expected_target_partition_filter.csv": seeds__expected_target_partition_filter_csv}
+        return {
+            "expected_target_partition_filter.csv": seeds__expected_target_partition_filter_csv
+        }
 
     def test__merge_target_partition_filter(self, project):
         """Target partition filter should produce correct merge results with partition pruning."""
--- a/dbt-athena/tests/unit/test_adapter.py
+++ b/dbt-athena/tests/unit/test_adapter.py
@@ -1348,7 +1348,10 @@ class TestAthenaAdapter:
         ],
     )
     def test_format_one_partition_key_with_prefix(self, partition_key, prefix, expected_result):
-        assert self.adapter.format_one_partition_key_with_prefix(partition_key, prefix) == expected_result
+        assert (
+            self.adapter.format_one_partition_key_with_prefix(partition_key, prefix)
+            == expected_result
+        )
 
     def test_murmur3_hash_with_int(self):
         bucket_number = self.adapter.murmur3_hash(123, 100)
--- a/dbt-athena/tests/unit/test_get_partition_batches.py
+++ b/dbt-athena/tests/unit/test_get_partition_batches.py
@@ -152,7 +152,9 @@ def _render_macro(config, rows, column_types, build_target_conditions=False):
 
     # Load and render the main macro
     template = env.get_template("get_partition_batches.sql", globals=context)
-    template.module.get_partition_batches(sql="test", build_target_conditions=build_target_conditions)
+    template.module.get_partition_batches(
+        sql="test", build_target_conditions=build_target_conditions
+    )
 
     return result_holder["value"]
 


dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/merge.sql
```diff
@@ -48,19 +48,56 @@
...
         {%- do log('BATCH PROCESSING: ' ~ loop.index ~ ' OF ' ~ partitions_batches | length) -%}
         {%- set src_batch_part -%}
             merge into {{ target_relation }} as target
             using (select {{ dest_cols_csv }} from {{ tmp_relation }} where {{ batch }}) as src
         {%- endset -%}
-        {%- set merge_batch -%}
-          {{ src_batch_part }}
-          {{ merge_part }}
-        {%- endset -%}
+        {%- if build_target -%}
```
bernard-choi: **[For reviewers] source/target 배치의 1:1 대응**

source와 target 배치 리스트는 동일한 순서와 길이를 가집니다.
`loop.index0`으로 source batch에 대응하는 target filter를 가져옵니다.

이 1:1 대응은 `get_partition_batches`에서 보장됩니다:
- 두 리스트 모두 같은 순회 순서로 생성
- 같은 `athena_partitions_limit`으로 배치 분할
- `test_target_conditions_aligned_batch_count` 테스트에서 `len(source) == len(target)` 검증

dbt-athena/src/dbt/include/athena/macros/materializations/models/helpers/get_partition_batches.sql
```diff
@@ -1,4 +1,4 @@
-{% macro get_partition_batches(sql, as_subquery=True) -%}
+{% macro get_partition_batches(sql, as_subquery=True, build_target_conditions=False) -%}
```
bernard-choi: The return type changes depending on the `build_target_conditions` parameter:

- `False` (default): `list[str]` — WHERE clause fragments (existing behavior, unchanged)
  ```
  ['"date_col"=DATE\'2024-01-01\' or "date_col"=DATE\'2024-01-02\'', ...]
  ```
- `True`: `dict{'source': list[str], 'target': list[str]}`
  ```
  {
    'source': ['"date_col"=DATE\'2024-01-01\' or ...'],
    'target': ['target."date_col"=DATE\'2024-01-01\' or ...']
  }
  ```
  `source[i]` and `target[i]` correspond to the same partition batch.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/merge.sql
```diff
@@ -48,19 +48,56 @@
...
+
+    {# Check if bucket partitions exist - skip target filter if so #}
+    {%- set has_bucket = namespace(value=false) -%}
+    {%- if partitioned_by is not none -%}
+        {%- for key in partitioned_by -%}
+            {%- if modules.re.search('bucket\\(', key) -%}
+                {%- set has_bucket.value = true -%}
+            {%- endif -%}
+        {%- endfor -%}
+    {%- endif -%}
```
bernard-choi: Bucket partitions use hash-based assignment, so target-side column values cannot directly enable Iceberg partition pruning.

For example, with `bucket(user_id, 10)`, the source filter is `user_id IN ('alice', 'bob')` (original values), but achieving the same pruning on the target would require applying the `bucket()` function itself in the ON clause. Whether Athena/Iceberg can leverage this for pruning is uncertain, and the added complexity outweighs the potential benefit.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/merge.sql
```diff
@@ -119,7 +157,10 @@
           {%- endfor %}
```
bernard-choi, line 157: `merge_on_part` intentionally leaves the ON clause's opening `(` unclosed. This allows `batch_iceberg_merge` to optionally inject `AND (target_filter)` inside the ON clause before closing it.

The closing `)` is added in three places:
- Batch + target filter path (L93)
- Batch-only path (L100)
- Non-batch path (L195)

```sql
-- Generated SQL structure:
ON (                          -- merge_on_part starts here
    target.id = src.id
    AND (target_filter)       -- optionally injected by batch_iceberg_merge
)                             -- closed by the caller
WHEN MATCHED THEN ...         -- merge_actions_part
```

dbt-athena/src/dbt/include/athena/macros/materializations/models/helpers/get_partition_batches.sql
bernard-choi, line 1: The return type changes depending on the `build_target_conditions` parameter:

- `False` (default): `list[str]` — WHERE clause fragments (existing behavior, unchanged)
  ```
  ['"date_col"=DATE\'2024-01-01\' or "date_col"=DATE\'2024-01-02\'', ...]
  ```
- `True`: `dict{'source': list[str], 'target': list[str]}`
  ```
  {
    'source': ['"date_col"=DATE\'2024-01-01\' or ...'],
    'target': ['target."date_col"=DATE\'2024-01-01\' or ...']
  }
  ```
  `source[i]` and `target[i]` correspond to the same partition batch.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/merge.sql
bernard-choi, line 84: `source` and `target` batch lists are guaranteed to have the same length and order. `loop.index0` retrieves the target filter corresponding to the current source batch.

This alignment is ensured by `get_partition_batches`, which builds both lists using the same traversal order and the same `athena_partitions_limit` batch size. Verified by the `test_target_conditions_aligned_batch_count` unit test (`len(source) == len(target)`).

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/merge.sql
bernard-choi, line 63: Bucket partitions use hash-based assignment, so target-side column values cannot directly enable Iceberg partition pruning.

For example, with `bucket(user_id, 10)`, the source filter is `user_id IN ('alice', 'bob')` (original values), but achieving the same pruning on the target would require applying the `bucket()` function itself in the ON clause. Whether Athena/Iceberg can leverage this for pruning is uncertain, and the added complexity outweighs the potential benefit.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/merge.sql
```diff
@@ -48,19 +48,56 @@
...
-    {% set partitions_batches = get_partition_batches(tmp_relation) %}
+{% macro batch_iceberg_merge(tmp_relation, target_relation, merge_on_part, merge_actions_part, dest_cols_csv) %}
+    {%- set merge_with_target_partition_filter = config.get('merge_with_target_partition_filter', False) | as_bool -%}
+    {%- set partitioned_by = config.get('partitioned_by') -%}
+
+    {# Check if bucket partitions exist - skip target filter if so #}
+    {%- set has_bucket = namespace(value=false) -%}
+    {%- if partitioned_by is not none -%}
+        {%- for key in partitioned_by -%}
+            {%- if modules.re.search('bucket\\(', key) -%}
```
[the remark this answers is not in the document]
bernard-choi: Good catch! Applied your suggestion in 0fc21cbd — the bucket detection regex is now case-insensitive with optional whitespace tolerance.

Verified with a manual test model (`BUCKET(category, 3)` uppercase) that the fix correctly detects bucket partitions and disables target filter injection.

PR #1824, opened 2026-04-01
fix(redshift): gate extended grants behind opt-in  flag

tauhid621:
resolves #

### Problem
PR #1597 added group and role support to Redshift's grants config by switching the underlying grants query from the legacy `pg_user + has_table_privilege()` path to `svv_relation_privileges` (and `SHOW GRANTS` for datasharing). This had an unintended side effect: `svv_relation_privileges` reports all grantees — users, groups, and roles — whereas the legacy query only reported users.

For any Redshift cluster where roles or groups have been granted privileges outside of dbt (e.g., via manual GRANT statements or Terraform), dbt now "sees" those grants for the first time. Because they don't appear in the user's grants config, dbt treats them as drift and issues REVOKE statements on every run, silently stripping externally-managed permissions.

This is a breaking change in default behavior — users who upgrade get unexpected revocations without opting in to anything.

### Solution

Introduce a new behavior flag `redshift_grants_extended` (default: False) that gates the extended grants functionality:

When disabled (default): Legacy behavior is fully preserved. The adapter uses `pg_user + has_table_privilege()` (or SHOW GRANTS filtered to users-only when datasharing is on). Grantees are plain usernames with no prefixes. Groups and roles are invisible to the grants diff, so externally-managed permissions are never revoked.

When enabled: The full extended behavior from #1597 is active. Grantees use user:, group:, role: prefixes. svv_relation_privileges (or SHOW GRANTS with group/role detection) is used. Groups and roles can be idempotently managed via grants config.

All four combinations of (use_show_apis, grants_extended) are handled consistently across both the SQL macros (get_show_grant_sql, apply_grants, get_grant_sql, get_revoke_sql) and the Python standardize_grants_dict method.

### Checklist

### DIFF
--- a/dbt-redshift/.changes/unreleased/Fixes-20260401-214927.yaml
+++ b/dbt-redshift/.changes/unreleased/Fixes-20260401-214927.yaml
@@ -1,5 +1,5 @@
 kind: Fixes
-body: Revert breaking grants behavior from 1597 behind opt-in flag `redshift_grants_extended` to prevent unintended revocation of externally-managed role grants
+body: Revert breaking grants behavior from 1597 behind opt-in flag `redshift_grants_extended` to prevent unintended revocation of externally-managed role and group grants
 time: 2026-04-01T21:49:27.086576+05:30
 custom:
     Author: tauhid621
--- a/dbt-redshift/tests/functional/adapter/test_grants.py
+++ b/dbt-redshift/tests/functional/adapter/test_grants.py
@@ -9,9 +9,9 @@ from dbt.tests.adapter.grants.test_snapshot_grants import BaseSnapshotGrants
 class RedshiftGrantsMixin:
     """Normalize expected grants to use 'user:' prefix for consistent comparison.
 
-    Redshift's get_show_grant_sql returns grantee names with 'user:' or 'group:'
-    prefixes when redshift_grants_extended is enabled. The base test expected
-    dicts use plain usernames, so we normalize them before comparison.
+    Redshift's get_show_grant_sql returns grantee names with 'user:', 'group:',
+    or 'role:' prefixes when redshift_grants_extended is enabled. The base test
+    expected dicts use plain usernames, so we normalize them before comparison.
     """
 
     @pytest.fixture(scope="class")


.github/workflows/_integration-tests.yml
```diff
@@ -352,6 +352,7 @@ jobs:
             REDSHIFT_TEST_IAM_USER_ACCESS_KEY_ID: ${{ vars.AWS_USER_ACCESS_KEY_ID }}
             REDSHIFT_TEST_IAM_USER_SECRET_ACCESS_KEY: ${{ secrets.AWS_USER_SECRET_ACCESS_KEY }}
             REDSHIFT_TEST_IAM_ROLE_PROFILE: ${{ vars.AWS_ROLE_PROFILE }}
+            DBT_TEST_GROUP_1: ${{ vars.DBT_TEST_GROUP_1 }}
```
tauhid621, line 355: The variable has been added in GitHub Variables.

dbt-redshift/src/dbt/include/redshift/macros/adapters/apply_grants.sql
```diff
@@ -50,6 +54,32 @@
     where namespace_name = '{{ relation.schema }}'
       and relation_name = '{{ relation.identifier }}'
       and not (identity_type = 'user' and identity_name = current_user)
+  {% else %}
+{#-
```
tauhid621, line 58: This part is same as the previous [implementation](https://github.com/jni-bot/dbt-adapters/blob/11fdfd28a3390992aead49d4527437dbb8237f88/dbt-redshift/src/dbt/include/redshift/macros/adapters/apply_grants.sql#L6)

dbt-redshift/src/dbt/include/redshift/macros/adapters/apply_grants.sql
```diff
@@ -50,6 +54,32 @@
...
+        select 'delete' as privilege_type
+        union all
+        select 'references' as privilege_type
+    )
+    select
+        u.usename as grantee,
+        p.privilege_type
+    from pg_user u
+    cross join privileges p
+    where has_table_privilege(u.usename, '{{ relation }}', privilege_type)
```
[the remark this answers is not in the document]
tauhid621: It would be better to keep the format which were used earlier

dbt-redshift/src/dbt/adapters/redshift/impl.py
```diff
@@ -318,18 +336,43 @@ def build_catalog_from_show_tables_and_svv_columns(
...
+                    privilege = row["privilege_type"].lower()
+                    if privilege in grants_dict:
+                        grants_dict[privilege].append(grantee)
+                    else:
+                        grants_dict[privilege] = [grantee]
+                return grants_dict
+
         if self.use_show_apis():
             # SHOW GRANTS path — need to detect groups from the / prefix
             for row in grants_table:
```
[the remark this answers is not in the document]
tauhid621: This makes sense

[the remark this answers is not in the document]
tauhid621: addressed

dbt-redshift/src/dbt/adapters/redshift/impl.py
```diff
@@ -318,17 +336,42 @@ def build_catalog_from_show_tables_and_svv_columns(
...
+        idempotently managed.  ``SHOW GRANTS`` is used for cross-database
+        support when ``datasharing`` is enabled; ``svv_relation_privileges``
+        is used otherwise.
         """
+
         grants_dict: Dict[str, List[str]] = {}
+        current_user = self.config.credentials.user.lower()
+
+        if not self.use_grants_extended():
+            if not self.use_show_apis():
```
tauhid621, line 355: Similar to initial [implementation](https://github.com/jni-bot/dbt-adapters/blob/11fdfd28a3390992aead49d4527437dbb8237f88/dbt-redshift/src/dbt/adapters/redshift/impl.py#L200) with a bit more filtering on show path

dbt-redshift/src/dbt/include/redshift/macros/adapters/apply_grants.sql
```diff
@@ -73,12 +103,19 @@
 
 
 {% macro redshift__get_grant_sql(relation, privilege, grantees) %}
-    {%- set formatted_grantees = redshift__format_grantees(grantees) -%}
-    grant {{ privilege }} on {{ relation.render() }} to {{ formatted_grantees | join(', ') }}
+    {% if redshift__use_grants_extended() %}
+        {%- set formatted_grantees = redshift__format_grantees(grantees) -%}
+        grant {{ privilege }} on {{ relation.render() }} to {{ formatted_grantees | join(', ') }}
+    {% else %}
+        {{ default__get_grant_sql(relation, privilege, grantees) }}
```
colin-k-rogers, line 110: can we rely on dispatch here? i.e. `get_grant_sql`?

tauhid621: It would create a recursion if adapter.dispatch is called here.

dbt-redshift/.changes/unreleased/Fixes-20260401-214927.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Revert breaking grants behavior from 1597 behind opt-in flag `redshift_grants_extended` to prevent unintended revocation of externally-managed role grants
```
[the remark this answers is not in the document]
tauhid621: addressed

dbt-redshift/tests/functional/adapter/test_grants.py
```diff
@@ -10,10 +10,14 @@ class RedshiftGrantsMixin:
     """Normalize expected grants to use 'user:' prefix for consistent comparison.
 
     Redshift's get_show_grant_sql returns grantee names with 'user:' or 'group:'
-    prefixes. The base test expected dicts use plain usernames, so we normalize
-    them before comparison.
+    prefixes when redshift_grants_extended is enabled. The base test expected
+    dicts use plain usernames, so we normalize them before comparison.
     """
```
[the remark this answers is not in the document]
tauhid621: addressed

dbt-redshift/src/dbt/include/redshift/macros/adapters/apply_grants.sql
colin-k-rogers: 🤦 ofc thanks

FridayPush:
This is currently hitting us and we're getting revoke statements like:
`revoke select on "marts"."a_table" from "group:data_engineer_dev"`

But that attempts to revoke from a *user* call group:data_engineer_dev. Rather than 'group data_engineer_dev'. Not sure if this PR also looks at that.

tauhid621:
@FridayPush Are you using custom `get_revoke_sql` macro?
This should ideally resolve the issue as it would gate the new behaviour

PR #1830, opened 2026-04-06
feat(athena): add build_strategy config for incremental and table materializations

dtaniwaki:
resolves #1829

### Problem

Athena incremental models always stage data into a `__dbt_tmp` table via CTAS before applying MERGE or INSERT. The tmp table exists to support automatic batching when `TOO_MANY_OPEN_PARTITIONS` occurs, but many models never hit this limit. For these models, the tmp table is pure overhead — unnecessary S3 writes, doubled scan costs, and added latency.

### Solution

Add `build_strategy` config option to control how a model's source SQL is materialized. The default `tmp_table` preserves today's behavior; `subquery` skips the full CTAS and uses the compiled SQL directly as a subquery.

- **Empty tmp table for schema comparison**: `CTAS ... WITH NO DATA` (zero data scan) preserves the `process_schema_changes` flow, so `on_schema_change` is fully supported
- **Merge strategy**: `MERGE INTO target USING (subquery) AS src ON ...`
- **Append strategy**: `INSERT INTO target SELECT ... FROM (subquery)`
- **Insert overwrite strategy**: subquery feeds the partition-overwrite INSERT directly
- **Table materialization**: `CREATE TABLE AS subquery` skips the intermediate tmp table
- Works with both Iceberg and Hive table types

Example configuration:

```yaml
# dbt_project.yml or model config
models:
  my_project:
    +build_strategy: subquery
```

```sql
-- model file
{{ config(build_strategy='subquery') }}
select ...
```

**Constraints:**

- `subquery` is incompatible with `force_batch` — batching requires physical tmp table partition metadata
- `subquery` is not supported with Python models
- With `subquery`, `TOO_MANY_OPEN_PARTITIONS` raises an error instead of auto-batching (clear message guides user to switch back to `tmp_table`)

#### Test infrastructure (Terraform)

The functional tests in `tests/functional/adapter/test_build_strategy.py` exercise every combination of `build_strategy` × incremental strategy × table type against a real Athena account. The Terraform below provisions the minimal AWS resources needed.

<details>
<summary>Terraform (click to expand)</summary>

```hcl
terraform {
  required_version = ">= 1.5"

  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 5.0"
    }
  }
}

provider "aws" {
  profile = "<your-aws-profile>"
  region  = "<your-region>"
}

locals {
  name = "<your-test-name>"
}

resource "aws_s3_bucket" "test" {
  bucket        = local.name
  force_destroy = true
}

resource "aws_s3_bucket_public_access_block" "test" {
  bucket                  = aws_s3_bucket.test.id
  block_public_acls       = true
  block_public_policy     = true
  ignore_public_acls      = true
  restrict_public_buckets = true
}

resource "aws_glue_catalog_database" "test" {
  name = replace(local.name, "-", "_")
}

resource "aws_athena_workgroup" "test" {
  name        = local.name
  description = "Functional test workgroup for dbt-athena build_strategy"

  configuration {
    enforce_workgroup_configuration = true

    result_configuration {
      output_location = "s3://${aws_s3_bucket.test.bucket}/output/"
    }
  }
}

output "s3_staging_dir" {
  value = "s3://${aws_s3_bucket.test.bucket}/staging/"
}

output "s3_tmp_table_dir" {
  value = "s3://${aws_s3_bucket.test.bucket}/tmp/"
}

output "database" {
  value = aws_glue_catalog_database.test.name
}

output "workgroup" {
  value = aws_athena_workgroup.test.name
}
```

</details>

### Checklist


dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -46,6 +47,18 @@
...
+  {% if build_with_subquery %}
+    {% if model_language == 'python' %}
+      {% do exceptions.raise_compiler_error('build_with_subquery is not supported with Python models.') %}
+    {% endif %}
+    {% if force_batch %}
+      {% do exceptions.raise_compiler_error('build_with_subquery is incompatible with force_batch. Batching requires data in the tmp table.') %}
+    {% endif %}
+    {% if strategy == 'insert_overwrite' %}
+      {% do exceptions.raise_compiler_error('build_with_subquery is not supported with insert_overwrite strategy.') %}
+    {% endif %}
```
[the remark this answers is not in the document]
dtaniwaki: I think there're pros and cons in ignoring the error in the full-refresh phase. The error will be eventually raised in the incremental phase, so we should catch it earlier.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -146,22 +159,36 @@
 
   -- Append Strategy --
   {% elif strategy == 'append' %}
-    {% if old_tmp_relation is not none %}
-      {% do drop_relation(old_tmp_relation) %}
+    {% if build_with_subquery %}
+      {% if old_tmp_relation is not none %}
+        {% do drop_relation(old_tmp_relation) %}
+      {% endif %}
+      {%- set empty_sql = 'SELECT * FROM (' ~ compiled_code ~ ') _dbt_sbq WITH NO DATA' -%}
```
iconara: Is there any chance that `compiled_code` could end with a comment? I assume that this could happen if for example the model's last line has a comment. In that case a newline is needed before the closing paragraph to avoid the rest of the statement ending up inside of the comment.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -186,29 +213,55 @@
...
     {% endif %}
-    {% if old_tmp_relation is not none %}
-      {% do drop_relation(old_tmp_relation) %}
+
+    {% if build_with_subquery %}
+      -- Create empty tmp table for schema comparison (no data scan)
+      {% if old_tmp_relation is not none %}
+        {% do drop_relation(old_tmp_relation) %}
+      {% endif %}
+      {%- set empty_sql = 'SELECT * FROM (' ~ compiled_code ~ ') _dbt_sbq WITH NO DATA' -%}
```
iconara: Same here, a newline before the closing parenthesis is needed if there is a possibility that `compiled_code` could end with a comment.

iconara (commented):
Nice to see this being implemented, I think it will be appreciated.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
dtaniwaki: I fixed it 👍 
https://github.com/dbt-labs/dbt-adapters/pull/1830/changes/f95a3ae74c90ff2ccde0901baedc4dc306dabe02

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
dtaniwaki: I fixed it 👍 
https://github.com/dbt-labs/dbt-adapters/pull/1830/changes/f95a3ae74c90ff2ccde0901baedc4dc306dabe02

dtaniwaki:
@colin-rogers-dbt @iconara Could you review this PR? I think direct subquery insert also solves many situations of HIVE_TOO_MANY_PARTITIONS and ICEBERG_TOO_MANY_OPEN_PARTITIONS.

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -104,6 +104,9 @@ class AthenaConfig(AdapterConfig):
         force_batch: Skip creating the table as ctas and run the operation directly in batch insert mode.
         unique_tmp_table_suffix: Enforce the use of a unique id as tmp table suffix instead of __dbt_tmp.
         temp_schema: Define in which schema to create temporary tables used in incremental runs.
+        build_with_subquery: Use a subquery directly instead of staging data into __dbt_tmp.
+            Creates an empty tmp table (WITH NO DATA) for schema comparison, then applies via subquery.
+            Supported for Iceberg merge and append strategies, and Hive append. Incompatible with force_batch.
     """
```
iconara, line 113: If you aren't familiar with dbt-athena since before this change, you may be confused by the naming of this property. The name is a response to the current default, but the current default is actually not the default behavior that you would expect.

It's natural to add a feature with a config that enables it, but I'm thinking that in this case we are adding something that probably should have been the default mode, and it's only for historical reasons it ended up being added later. At least to me, using a temp table is the alternative way of doing this.

One way to convey this is to use a config that is not true or false, but enables one of two (or more) modes, like `incremental_strategy` with values "subquery" or "tmp_table" (default).

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -46,6 +47,18 @@
     {% endif %}
   {% endif %}
 
+  {% if build_with_subquery %}
+    {% if model_language == 'python' %}
+      {% do exceptions.raise_compiler_error('build_with_subquery is not supported with Python models.') %}
+    {% endif %}
+    {% if force_batch %}
+      {% do exceptions.raise_compiler_error('build_with_subquery is incompatible with force_batch. Batching requires data in the tmp table.') %}
```
iconara: Nitpick: I would say "Batching requires a temp table"

iconara (commented):
I'm not familiar enough with the way the jinja templates for CTAS/INSERT/MERGE worked before to say anything about that part, but based on the tests I think this is sound. I left a philosophical comment about naming, and how looking at this from an outside perspective, this new way of handling incremental loads is probably how most would expect it to work by default, and the config naming should reflect that (the current default behavior needs to stay the current default, but naming can be used to signal what the "basic" or simplest behavior is).

dbt-athena/src/dbt/adapters/athena/impl.py
dtaniwaki: That sounds reasonable. `incremental_strategy` is the same as dbt-core's config like `incremental_strategy: insert_overwrite`. How about `build_strategy`?

dtaniwaki: What do you think about applying this mode to table materialization? This behavior benefits all the materialization types rather than the incremental materialization.

I made another PR over this PR in https://github.com/dtaniwaki/dbt-adapters/pull/12.

iconara: `build_strategy` sounds good to me, that works for table materialization too.

PR #1850, opened 2026-04-10
fix(athena): propagate CatalogId in LfTagsManager and get_table_versions paginator

julietaansola:
## Summary

Two methods in the Athena adapter did not propagate the cross-account `CatalogId` when addressing Glue and Lake Formation APIs, causing failures when using dbt-athena with a federated (cross-account) Glue catalog exposed via an Athena `DataCatalog` resource with a `catalog-id` parameter.

### Bug 1 — `impl.py::_get_glue_table_versions_to_expire`

The `get_table_versions` paginator is called without `CatalogId`, so boto3 defaults to the caller's own account. When the target database lives in a cross-account federated Glue catalog, this fails with:

```
EntityNotFoundException: Database <name> not found
```

The caller `expire_glue_table_versions` already computes `catalog_id` from `_get_data_catalog + get_catalog_id`, but does not propagate it to the read-side paginator.

### Bug 2 — `lakeformation.py::LfTagsManager`

The class does not accept or store a `catalog_id`. The `Resource` dicts built in `process_lf_tags_database`, `process_lf_tags`, `_remove_lf_tags_columns` and `_apply_lf_tags_columns` omit the `CatalogId` field, so Lake Formation resolves the target in the caller's account and returns:

```
EntityNotFoundException: Entity Not Found
```

on `get_resource_lf_tags`, `add_lf_tags_to_resource` and `remove_lf_tags_from_resource`.

## Fix

The fix follows the pattern already established by `LfPermissions` in the same `lakeformation.py` file, which accepts a `catalog_id` in its constructor and passes it to all LF API calls.

- `LfTagsManager.__init__` now accepts `catalog_id: Optional[str] = None` and stores it.
- All four methods building LF `Resource` dicts inject `CatalogId` when set.
- The two call sites in `impl.py` (`add_lf_tags_to_database` and `add_lf_tags`) now compute `catalog_id` via `_get_data_catalog` + `get_catalog_id` and pass it to the manager.
- `_get_glue_table_versions_to_expire` now computes `catalog_id` and passes it to the `get_table_versions` paginator when set.

Backward compatible: `catalog_id` is optional and defaults to `None`; when not set, behavior is identical to before.

## Testing

Validated end-to-end against a real cross-account setup:

- **Account A (data lake)**: Glue catalog + Lake Formation with LF-Tags (`Classification`, etc.) and a cross-account share to Account B.
- **Account B (domain)**: running dbt with an Athena `DataCatalog` resource where `Parameters.catalog-id = <Account A>`.

**Before the fix**: `dbt run --full-refresh` on a model with `lf_tags_config` fails with `EntityNotFoundException` in both `GetTableVersions` and `GetResourceLFTags`.

**After the fix**: the model completes the CREATE TABLE flow successfully against the cross-account catalog, and LF API calls correctly address Account A.

(Remaining errors after the fix are real LF grant issues — e.g. missing `ASSOCIATE` on tags — which are out of scope for this PR.)

## Checklist

- [x] Changelog entry added under `dbt-athena/.changes/unreleased/`
- [x] Code follows existing patterns (`LfPermissions` in the same file)
- [x] Backward compatible (new parameter is optional)
- [ ] Unit tests added (happy to add if maintainers prefer; the existing mock pattern in `tests/unit/test_lakeformation.py` should extend cleanly)


julietaansola:
@cla-bot check

julietaansola:
Hi team! Could someone approve the CI workflow run so the checks can execute? Happy to address any feedback. Thanks!

PR #1851, opened 2026-04-10
feat(snowflake): use CREATE OR ALTER for dynamic table config changes

### LINKED ISSUE #1825: [Snowflake] Use CREATE OR ALTER DYNAMIC TABLE to detect and apply SQL definition changes

## Problem

When the SQL body of a `dynamic_table` model changes, `dbt run` does not detect the change. It only compares configuration parameters (target_lag, warehouse, etc.) and emits a warning like:

```
warning: dbt1065: No configuration changes were identified on: `...`. Continuing.
```

The only workaround is to manually pass `--full-refresh`, which defeats dbt's declarative approach. Every other materialization type (table, view, incremental) applies SQL changes automatically.

This was raised in [dbt-core#8862](https://github.com/dbt-labs/dbt-core/issues/8862) (open since Oct 2023), where the team noted that most warehouses don't expose the deployed SQL, making comparison difficult. That argument does not apply to Snowflake.

## Proposed solution

Snowflake now supports [`CREATE OR ALTER DYNAMIC TABLE`](https://docs.snowflake.com/en/sql-reference/sql/create-dynamic-table#label-create-or-alter-dt-syntax) (Open Preview since Sep 2025, release 9.29). This command is declarative and idempotent:

- **Nothing changed** — no-op, no refresh triggered
- **Config-only change** (target_lag, warehouse, cluster_by, ...) — applied in place, data preserved
- **Query or column change** — triggers reinitialization

The current `SHOW DYNAMIC TABLES` + compare config + `ALTER` lifecycle could be replaced with a single `CREATE OR ALTER DYNAMIC TABLE` statement, letting Snowflake handle change detection natively. This would be **simpler code** and **correct behavior**.

- **Manual `--full-refresh`**: Works, but requires the user to know the SQL changed and remember to pass the flag. Error-prone and not declarative.
- **Comparing deployed SQL via `SHOW DYNAMIC TABLES`**: The `text` column from `SHOW DYNAMIC TABLES` contains the deployed query. dbt could compare it against compiled SQL to detect changes. However, Snowflake may rewrite/reformat the query (whitespace, casing, fully qualified names), making string comparison fragile. `CREATE OR ALTER` avoids this entirely by letting Snowflake handle the comparison internally.

### Who will this benefit?

Anyone using `dynamic_table` materialization on Snowflake. Currently, SQL body changes are silently ignored unless `--full-refresh` is passed, which is a surprising and error-prone behavior.

_No response_

### Anything else?

Caveats of `CREATE OR ALTER DYNAMIC TABLE`:

- Currently in Open Preview (available to all accounts, not yet Generally Available) — could be gated behind a flag or opt-in config if the dbt team prefers to wait for General Availability
- Not atomic (partial changes possible on failure, but recoverable)
- Cannot change `REFRESH_MODE` after creation
- Does not support tag/policy changes (existing ones are preserved)
- Not supported for Iceberg dynamic tables

#### Implementation pointers (LLM-investigated, didn't check *everything*)

The current lifecycle is driven by the materialization macro [`dynamic_table_get_build_sql`](https://github.com/dbt-labs/dbt-adapters/blob/bd80e5a9d4b7b3b0200872892ff41994586c72ef/dbt-snowflake/src/dbt/include/snowflake/macros/materializations/dynamic_table.sql). When the table already exists and `--full-refresh` is not set, it calls `snowflake__get_dynamic_table_configuration_changes` to compare configs. If that returns `None` (no config diff), the macro emits the "No configuration changes" warning and does nothing — this is where the SQL body change is silently ignored.

Key files to modify:

- **Materialization entry point (Jinja)**: [`dynamic_table.sql` L34-67](https://github.com/dbt-labs/dbt-adapters/blob/bd80e5a9d4b7b3b0200872892ff41994586c72ef/dbt-snowflake/src/dbt/include/snowflake/macros/materializations/dynamic_table.sql#L34-L67) — the `dynamic_table_get_build_sql` macro contains the decision tree (create / replace / alter / no-op). The key change: when `configuration_changes is none` ([L49](https://github.com/dbt-labs/dbt-adapters/blob/bd80e5a9d4b7b3b0200872892ff41994586c72ef/dbt-snowflake/src/dbt/include/snowflake/macros/materializations/dynamic_table.sql#L49)), the macro currently emits a no-op — this should instead emit a `CREATE OR ALTER DYNAMIC TABLE` to catch SQL body changes. The `requires_full_refresh` → DROP + CREATE path and `on_configuration_change` = `fail`/`continue` checks still need to be preserved for cases `CREATE OR ALTER` cannot handle.
- **Create SQL (Jinja)**: [`create.sql` L17-63](https://github.com/dbt-labs/dbt-adapters/blob/bd80e5a9d4b7b3b0200872892ff41994586c72ef/dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/create.sql#L17-L63) — `snowflake__create_dynamic_table_info_schema_sql` already generates the full `create dynamic table ... as (...)` DDL. A `CREATE OR ALTER` variant would be nearly identical, just replacing `create` with `create or alter`.
- **Alter SQL (Jinja)**: [`alter.sql` L1-10](https://github.com/dbt-labs/dbt-adapters/blob/bd80e5a9d4b7b3b0200872892ff41994586c72ef/dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/alter.sql#L1-L10) — the current `ALTER DYNAMIC TABLE ... SET/UNSET` logic for target_lag, warehouse, cluster_by, etc. These operations are subsumed by `CREATE OR ALTER`. However, the `requires_full_refresh` fallback path ([L9-10](https://github.com/dbt-labs/dbt-adapters/blob/bd80e5a9d4b7b3b0200872892ff41994586c72ef/dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/alter.sql#L9-L10): DROP + CREATE via `get_replace_sql`) is **still needed** for changes that `CREATE OR ALTER` cannot handle: `REFRESH_MODE` changes, TRANSIENT conversion, and tag/policy modifications.
- **Config comparison (Python)**: [`relation.py` L97](https://github.com/dbt-labs/dbt-adapters/blob/bd80e5a9d4b7b3b0200872892ff41994586c72ef/dbt-snowflake/src/dbt/adapters/snowflake/relation.py#L97) (`dynamic_table_config_changeset`), [`impl.py` L532](https://github.com/dbt-labs/dbt-adapters/blob/bd80e5a9d4b7b3b0200872892ff41994586c72ef/dbt-snowflake/src/dbt/adapters/snowflake/impl.py#L532) (`describe_dynamic_table`), and [`dynamic_table.py` L319-365](https://github.com/dbt-labs/dbt-adapters/blob/bd80e5a9d4b7b3b0200872892ff41994586c72ef/dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py#L319-L365) (`SnowflakeDynamicTableConfigChangeset` with `requires_full_refresh` and `has_changes` properties). This step is still needed to detect `requires_full_refresh` scenarios (REFRESH_MODE, TRANSIENT changes) and to respect `on_configuration_change` = `fail`/`continue`. But the "no changes detected → no-op" branch would change to "no changes detected → `CREATE OR ALTER`" to catch SQL body changes.

igorbelianski-cyber:
## Summary

- **Use `CREATE OR ALTER DYNAMIC TABLE` instead of `CREATE OR REPLACE`** when applying configuration changes (`on_configuration_change=apply`) to existing Snowflake dynamic tables. This preserves table state and avoids unnecessary recreation. It is expected to preserve and not reinit dynamic table in some scenario with SQL changes as well .
- **Iceberg dynamic tables continue to use `ALTER DYNAMIC TABLE`** because Snowflake does not yet support `CREATE OR ALTER DYNAMIC ICEBERG TABLE`.
- **Fix false-positive change detection** by stripping the `LINEAR(...)` wrapper Snowflake adds to `cluster_by` metadata, preventing spurious re-creation on identical re-runs.
- **`refresh_mode` changes no longer require full refresh** — they can be applied via `CREATE OR ALTER`.
- **Transient ↔ non-transient changes still require full refresh** (`CREATE OR REPLACE`), as expected.
- **Remove `row_access_policy` and `table_tag` from CREATE OR ALTER DDL** — Snowflake error 001506: `CREATE OR ALTER does not support setting policies or tags`. Including these clauses caused a hard SQL error. Policies/tags set during initial `CREATE` remain attached since Snowflake preserves them across `CREATE OR ALTER`.

Addresses: https://github.com/dbt-labs/dbt-adapters/issues/1825

### Macros
- **`create_or_alter.sql` (new)**: Generates `CREATE OR ALTER DYNAMIC TABLE` DDL for native (non-Iceberg) dynamic tables with all supported attributes (target_lag, warehouse, refresh_mode, scheduler, initialization_warehouse, immutable_where, cluster_by). Explicitly omits `row_access_policy` and `table_tag` per Snowflake limitation (error 001506).
- **`alter.sql` (modified)**: Refactored to serve Iceberg tables only. Renamed macro to `snowflake__alter_dynamic_table_built_in_sql`. Removed the `requires_full_refresh` branch (handled by caller).
- **`dynamic_table.sql`**: Updated `apply` branch to call `snowflake__get_create_or_alter_dynamic_table_sql` instead of the old alter macro.

### Python
- **`dynamic_table.py`**: Strip `LINEAR(...)` wrapper from Snowflake's `cluster_by` metadata to prevent false-positive change detection. `refresh_mode` changes no longer require full refresh.

### Tests
- Verify `CREATE OR ALTER` DDL for apply path, `CREATE OR REPLACE` for full refresh
- No-op detection for identical configs across `dbt run` and `dbt build` (various `cluster_by` shapes: expression, single column, two columns)
- SQL-only model changes do not trigger `CREATE OR ALTER`
- Iceberg ALTER path for `initialization_warehouse`, `immutable_where`, `cluster_by`
- Transient changes trigger `CREATE OR REPLACE`, not `CREATE OR ALTER`
- Extended scheduler validation unit tests
- **NEW**: `CREATE OR ALTER` succeeds on DTs with `row_access_policy` or `table_tag` (policy/tag omitted from DDL, preserved by Snowflake)
- **NEW**: Policy/tag-only changes documented as no-ops (known limitation: dbt change detection does not track these fields)

## Test plan

- [x] All 237 Snowflake unit tests pass
- [x] All 65 dynamic table integration tests pass (native + Iceberg, including 4 new policy/tag tests)
- [x] Manual verification: `dbt run` and `dbt build` with no config changes produce no DDL
- [x] Manual verification: `dbt build --debug` no longer triggers spurious `CREATE OR ALTER`

Resolves #1825

Made with [Cursor](https://cursor.com)

### DIFF
--- a/dbt-snowflake/.changes/unreleased/Features-20260408-154451.yaml
+++ b/dbt-snowflake/.changes/unreleased/Features-20260408-154451.yaml
@@ -2,5 +2,5 @@ kind: Features
 body: 'Snowflake dynamic tables: use CREATE OR REPLACE only with --full-refresh (and when replacing an existing non-dynamic-table relation of the same name). For existing dynamic tables without --full-refresh, configuration-driven updates use CREATE OR ALTER instead of CREATE OR REPLACE. Dynamic iceberg tables continue to use ALTER DYNAMIC TABLE because Snowflake does not yet support CREATE OR ALTER DYNAMIC ICEBERG TABLE. Note: dbt does not compare compiled SQL to SHOW DYNAMIC TABLES text, so model body-only changes with identical config still require --full-refresh.'
 time: 2026-04-08T15:44:51.442042-07:00
 custom:
-    Author: dbt-labs
+    Author: igorbelianski-cyber
     Issue: "1825"
--- a/dbt-snowflake/.changes/unreleased/Features-20260408-154451.yaml
+++ b/dbt-snowflake/.changes/unreleased/Features-20260408-154451.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: 'Snowflake dynamic tables: use CREATE OR REPLACE only with --full-refresh (and when replacing an existing non-dynamic-table relation of the same name). For existing dynamic tables without --full-refresh, configuration-driven updates use CREATE OR ALTER instead of CREATE OR REPLACE. Dynamic iceberg tables continue to use ALTER DYNAMIC TABLE because Snowflake does not yet support CREATE OR ALTER DYNAMIC ICEBERG TABLE. Note: dbt does not compare compiled SQL to SHOW DYNAMIC TABLES text, so model body-only changes with identical config still require --full-refresh.'
+time: 2026-04-08T15:44:51.442042-07:00
+custom:
+    Author: dbt-labs
+    Issue: "1825"
--- a/dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
@@ -206,11 +206,16 @@ class SnowflakeDynamicTableConfig(SnowflakeRelationConfigBase):
                 # Strip "IMMUTABLE WHERE (" prefix and ")" suffix
                 immutable_where = immutable_where_str[17:-1]  # len("IMMUTABLE WHERE (") = 17
 
-        # Snowflake may return empty string for unset cluster_by
-        # Normalize to Python None for consistency
+        # Snowflake may return empty string for unset cluster_by — normalize to None.
+        # When all cluster_by elements are plain column references, Snowflake wraps
+        # the entire expression with LINEAR(...) as the default clustering method
+        # (e.g. "id" → "LINEAR(id)", "id, value" → "LINEAR(id, value)").
+        # Strip the wrapper so the value matches the user's config.
         cluster_by = dynamic_table.get("cluster_by")
         if cluster_by is not None and str(cluster_by).strip() not in ("", "NONE", "None"):
             cluster_by = str(cluster_by).strip()
+            if cluster_by.upper().startswith("LINEAR(") and cluster_by.endswith(")"):
+                cluster_by = cluster_by[7:-1]
         else:
             cluster_by = None
 
@@ -275,7 +280,7 @@ class SnowflakeDynamicTableRefreshModeConfigChange(RelationConfigChange):
 
     @property
     def requires_full_refresh(self) -> bool:
-        return True
+        return False
 
 
 @dataclass(frozen=True, eq=True, unsafe_hash=True)
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/alter.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/alter.sql
@@ -1,4 +1,4 @@
-{% macro snowflake__alter_dynamic_table_built_in_sql(existing_relation, configuration_changes) -%}
+{% macro snowflake__get_alter_dynamic_table_as_sql(existing_relation, configuration_changes) -%}
 {#-
     Produce DDL that alters a dynamic iceberg table using ALTER DYNAMIC TABLE statements.
 
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/alter.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/alter.sql
@@ -1,15 +1,13 @@
-{% macro snowflake__get_alter_dynamic_table_as_sql(
-    existing_relation,
-    configuration_changes,
-    target_relation,
-    sql
-) -%}
-    {{- log('Applying ALTER to: ' ~ existing_relation) -}}
+{% macro snowflake__alter_dynamic_table_built_in_sql(existing_relation, configuration_changes) -%}
+{#-
+    Produce DDL that alters a dynamic iceberg table using ALTER DYNAMIC TABLE statements.
 
-    {% if configuration_changes.requires_full_refresh %}
-        {{- get_replace_sql(existing_relation, target_relation, sql) -}}
+    Snowflake does not support CREATE OR ALTER DYNAMIC ICEBERG TABLE,
+    so we fall back to individual ALTER statements for iceberg tables.
 
-    {% else %}
+    The requires_full_refresh check is handled by the caller
+    (snowflake__get_create_or_alter_dynamic_table_sql) before dispatching here.
+-#}
 
         {%- set target_lag = configuration_changes.target_lag -%}
         {%- if target_lag -%}{{- log('Applying UPDATE TARGET_LAG to: ' ~ existing_relation) -}}{%- endif -%}
@@ -65,7 +63,5 @@
         {% if has_prior_statements %};{% endif %}
         alter dynamic table {{ existing_relation }} drop clustering key
         {% endif %}
-
-    {%- endif -%}
 
 {%- endmacro %}
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/create_or_alter.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/create_or_alter.sql
@@ -1,0 +1,61 @@
+{% macro snowflake__get_create_or_alter_dynamic_table_sql(
+    existing_relation,
+    configuration_changes,
+    target_relation,
+    sql
+) -%}
+
+    {% if configuration_changes.requires_full_refresh %}
+        {{- log('Applying full refresh (CREATE OR REPLACE) to: ' ~ existing_relation ~ ' due to changes that require table recreation') -}}
+        {{- get_replace_sql(existing_relation, target_relation, sql) -}}
+
+    {% else %}
+
+        {%- set catalog_relation = adapter.build_catalog_relation(config.model) -%}
+
+        {%- if catalog_relation.catalog_type == 'INFO_SCHEMA' -%}
+            {%- set dynamic_table = target_relation.from_config(config.model) -%}
+            {{- log('Applying CREATE OR ALTER to: ' ~ existing_relation) -}}
+            {{ snowflake__create_or_alter_dynamic_table_info_schema_sql(dynamic_table, target_relation, sql) }}
+        {%- elif catalog_relation.catalog_type == 'BUILT_IN' -%}
+            {{- log('Applying ALTER to: ' ~ existing_relation) -}}
+            {{ snowflake__alter_dynamic_table_built_in_sql(existing_relation, configuration_changes) }}
+        {%- else -%}
+            {% do exceptions.raise_compiler_error('Unexpected model config for: ' ~ target_relation) %}
+        {%- endif -%}
+
+    {%- endif -%}
+
+{%- endmacro %}
+
+
+{% macro snowflake__create_or_alter_dynamic_table_info_schema_sql(dynamic_table, relation, sql) -%}
+
+    {%- if dynamic_table.transient is not none -%}
+        {%- set is_transient = dynamic_table.transient -%}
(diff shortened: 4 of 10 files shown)


dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/create_or_alter.sql
```diff
@@ -0,0 +1,61 @@
...
+    {{ optional('initialization_warehouse', dynamic_table.snowflake_initialization_warehouse) }}
+    {{ optional('refresh_mode', dynamic_table.refresh_mode) }}
+    {{ optional('initialize', dynamic_table.initialize) }}
+    {% if dynamic_table.scheduler is not none %}
+    scheduler = '{{ dynamic_table.scheduler }}'
+    {% elif dynamic_table.target_lag is none %}
+    scheduler = 'DISABLE'
+    {% endif %}
+    {{ optional('with row access policy', dynamic_table.row_access_policy, equals_char='') }}
+    {{ optional('with tag', dynamic_table.table_tag, quote_char='(', equals_char='') }}
```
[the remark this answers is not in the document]
igorbelianski-cyber: the old code already had a flavor of this problem  with row access policy and with tag only picked up during Create or replace neer during alter and not analyzing current state . Create or Alter implementation arguable makes this better by always picking up those attributes but still missing code that would look for changes vs existing state ,  
  I added tests for that missing check behavior  and add issue to track fix to properly query current tag access policy when needed (https://github.com/dbt-labs/dbt-adapters/issues/1864)

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/alter.sql
```diff
@@ -1,15 +1,13 @@
-{% macro snowflake__get_alter_dynamic_table_as_sql(
-    existing_relation,
-    configuration_changes,
-    target_relation,
-    sql
-) -%}
-    {{- log('Applying ALTER to: ' ~ existing_relation) -}}
+{% macro snowflake__alter_dynamic_table_built_in_sql(existing_relation, configuration_changes) -%}
```
colin-rogers-dbt: Can we avoid the rename here? This will break anyone using a custom materialization that references `get_alter_dynamic_table_as_sql`

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
```diff
@@ -275,7 +280,7 @@ class SnowflakeDynamicTableRefreshModeConfigChange(RelationConfigChange):
 
     @property
     def requires_full_refresh(self) -> bool:
-        return True
+        return False
```
colin-rogers-dbt, line 283: Would it be safer to make this conditional on catalog? i.e. return True for native DTs but keep False for iceberg?

Specifically thinking of a field like `refresh_mode` which could be silently dropped.

dbt-snowflake/tests/functional/relation_tests/dynamic_table_tests/models.py
```diff
@@ -94,6 +94,67 @@
...
+    materialized='dynamic_table',
+    snowflake_warehouse='DBT_TESTING',
+    target_lag='2 minutes',
+    refresh_mode='INCREMENTAL',
+) }}
+select *, 1 as extra_col from {{ ref('my_seed') }}
+"""
+
+
+DYNAMIC_TABLE_EXTRA_COLUMN_TARGET_LAG_FIVE = """
```
colin-rogers-dbt, line 147: Should this be added to a test?

dbt-snowflake/.changes/unreleased/Features-20260408-154451.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: 'Snowflake dynamic tables: use CREATE OR REPLACE only with --full-refresh (and when replacing an existing non-dynamic-table relation of the same name). For existing dynamic tables without --full-refresh, configuration-driven updates use CREATE OR ALTER instead of CREATE OR REPLACE. Dynamic iceberg tables continue to use ALTER DYNAMIC TABLE because Snowflake does not yet support CREATE OR ALTER DYNAMIC ICEBERG TABLE. Note: dbt does not compare compiled SQL to SHOW DYNAMIC TABLES text, so model body-only changes with identical config still require --full-refresh.'
+time: 2026-04-08T15:44:51.442042-07:00
+custom:
+    Author: dbt-labs
```
colin-rogers-dbt:
```suggestion
    Author: igorbelianski-cyber
```

colin-k-rogers (changes requested):
Primarily concerned about the change to `requires_full_refresh ` default, let's poke at that a bit

dbt-snowflake/tests/functional/relation_tests/dynamic_table_tests/test_configuration_changes.py
```diff
@@ -750,3 +759,415 @@ def test_non_transient_table_not_recreated_when_flag_on_and_config_absent(self,
...
+create or replace tag tag_name COMMENT = 'testing'
+"""
+
+
+class TestRowAccessPolicyWithCreateOrAlter:
+    """Tests that CREATE OR ALTER succeeds when row_access_policy is configured.
+
+    Validates that:
+    - A config change (target_lag) on a DT with row_access_policy triggers CREATE OR ALTER
+    - The CREATE OR ALTER DDL includes the policy and succeeds
```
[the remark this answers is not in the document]
igorbelianski-cyber: done

dbt-snowflake/.changes/unreleased/Features-20260408-154451.yaml
igorbelianski-cyber: done

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/dynamic_table.py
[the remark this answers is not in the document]
igorbelianski-cyber: see above

igorbelianski-cyber: implemented in sql template ( not certain this is the correct approach though) PTAL

dbt-snowflake/src/dbt/include/snowflake/macros/relations/dynamic_table/alter.sql
igorbelianski-cyber: done

dbt-snowflake/tests/functional/relation_tests/dynamic_table_tests/models.py
igorbelianski-cyber: added with a few missing tests cases

shaariqch-carma:
hello! Is there an update to this PR? Would love to have it merged as this is a highly emerging use case.

igorbelianski-cyber:
> hello! Is there an update to this PR? Would love to have it merged as this is a highly emerging use case.

we have a new version of this in the works: https://github.com/dbt-labs/dbt-adapters/commit/443301eb714d45c162199daa80af6fbee4b55ea6 

 please add comments suggestions there

PR #1853, opened 2026-04-12
feat(redshift): add application_name param to connection

### LINKED ISSUE #1852: [Feature] Add application_name in Redshift connector via profile key

### Describe the feature

Redshift, the same as Postgres, has a connection parameter called `application_name`, which is visible via the `STL_CONNECTION_LOG` and `SYS_CONNECTION_LOG` system views. Currently it uses the default set by the connector which defaults to the module that called the connection function, which is currently `dbt.adapters.base.connections`, at least for dbt `1.10.20` which is our current version.

In our company, we have several dbt projects being invoked from multiple places and even though we can control lots of elements from the query comments, profiles, users, etc.; controlling connections would also be very useful and give us an extra layer of control.

This feature has already been implemented for Postgres here: https://github.com/dbt-labs/dbt-core/issues/885

And from these comments I can see it's been implemented similarly for other engines as well: https://github.com/dbt-labs/dbt-core/pull/3182#discussion_r599746515

The simplest alternative I've found is to set a project-wide pre_hook:

```
models:
  project:
    +pre-hook:
      - "SET APPLICATION_NAME = 'some_identifier';"
```

Downside from this workaround is that only model connections set the application_name, whereas internal dbt connections do not.

### Who will this benefit?

dbt-redshift users only, mostly beneffiting managers of dbt implementations and Redshift managers.

Yes, I've already based on the code for the Postgres implementation and will create a PR soon after this issue

### Anything else?

_No response_

jchaconbogarin:
resolves #1852

### Problem

The Redshift connection doesn't currently have the `application_name` parameter, so it can't be configured from the `profiles.yml` file. This makes it more complicated from Redshift management perspective to identify connections coming from dbt, as other Engines do have already like Postgres, BigQuery or Snowflake.

### Solution

The solution here is simple. Add the `application_name` parameter in the RedshiftCredentials class defined on the  `connections.py` file from the Redshift adapter's code and then referring to this parameter in the `base_kwargs` to include it and pass it on the adapter's connection.

### Checklist


dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -275,6 +277,7 @@ def __base_kwargs(credentials) -> Dict[str, Any]:
             "db_groups": credentials.db_groups,
             "timeout": credentials.connect_timeout,
             "is_serverless": is_serverless(credentials),
+            "application_name": credentials.application_name,
```
tauhid621, line 280: Better to have this conditional similar to postgres
https://github.com/dbt-labs/dbt-adapters/blob/0efd8d3d1081e1ab43e38797d5104f7b424a6284/dbt-postgres/src/dbt/adapters/postgres/connections.py#L186

tauhid621 (changes requested):
Thanks for working on this.
We should add unit tests for the change to ensure test coverage.
Changelog entry should also be added.

PR #1869, opened 2026-04-16
Data contract views

### LINKED ISSUE #1844: [Feature] dbt-athena - add support for Data Catalog Views (Multi Dialect Views)

### Describe the feature

AWS Athena supports **Data Catalog Views** (also called Multi Dialect Views / MDVs) — views registered in the AWS Glue Data Catalog with Lake Formation security semantics. They use special DDL syntax:

```sql
CREATE [ OR REPLACE ] PROTECTED MULTI DIALECT VIEW view_name
SECURITY DEFINER
AS athena-sql-statement
```

The dbt-athena adapter currently only generates standard `CREATE OR REPLACE VIEW` DDL.

This feature request proposes adding a config flag `is_data_catalog_view` on the existing view materialization to conditionally generate MDV DDL instead of the standard view DDL.

**Example usage:**

```sql
-- models/my_secure_view.sql
{{ config(
    materialized='view',
    is_data_catalog_view=True,
    lf_tags_config={'tag_key': 'tag_value'}  -- optional, works with existing LF tags
) }}

select orderdate, sum(totalprice) as price
from {{ ref('orders') }}
where order_city = 'SEATTLE'
group by orderdate
```

**Scope of changes:**

1. Register `is_data_catalog_view: bool = False` in `AthenaConfig` (to avoid "unrecognized config" warnings)
2. Modify the `athena__create_view_as` macro to conditionally emit MDV syntax when the flag is set
3. No changes needed to relation types, drop logic, or the view materialization orchestration — MDVs are `VIRTUAL_VIEW` in Glue and existing `DROP VIEW IF EXISTS` works for them

**Separate materialization type** (e.g., `materialized='data_catalog_view'`): This would require duplicating the entire view materialization logic. Since MDVs differ from standard views only in the CREATE DDL statement, a config flag on the existing view materialization is simpler and avoids code duplication.

### Who will this benefit?

Teams using AWS Athena with Lake Formation who need governed, cross-engine views with fine-grained security. Here's how Data Catalog Views differ from standard Athena views:

| | **Standard Athena View** | **Data Catalog View (MDV)** |
|---|---|---|
| **Scope** | Athena-only | Cross-service (Athena, Redshift, EMR, etc.) |
| **Where it lives** | Athena's internal metadata | AWS Glue Data Catalog |
| **Security model** | Standard IAM/Athena permissions | Definer semantics via Lake Formation — invokers don't need access to underlying tables |
| **Lake Formation** | Optional | Required — tables must be LF-managed |
| **Can reference other views?** | Yes | No (also no resource links) |
| **Max tables referenced** | No documented limit | 10 tables max |
| **UDFs / Federated sources** | Yes | No |

This is relevant for any organization that:

- Needs **cross-service access** — the same view queried from both Athena and Redshift Spectrum
- Uses **Lake Formation** to manage fine-grained (row/column-level) data access permissions
- Wants **definer-based security** — expose curated data to downstream consumers without granting them access to the raw underlying tables

Yes, I plan to submit a PR implementing this feature.

### Anything else?

- AWS documentation on Data Catalog Views: https://docs.aws.amazon.com/athena/latest/ug/data-catalog-views.html
- Lake Formation fine-grained access control: https://docs.aws.amazon.com/lake-formation/latest/dg/tag-based-access-control.html
- The existing dbt-athena Lake Formation integration (lf_tags_config, lf_grants) works as-is with MDVs — no changes needed to that integration.

adavoudi:
resolves #1844

### Problem

The dbt-athena adapter only generates standard `CREATE OR REPLACE VIEW` DDL. AWS Athena also supports [Data Catalog Views](https://docs.aws.amazon.com/glue/latest/dg/catalog-views.html) (Multi Dialect Views / MDVs) — views registered in the AWS Glue Data Catalog with Lake Formation security semantics that use a different DDL syntax:

```sql
CREATE OR REPLACE PROTECTED MULTI DIALECT VIEW view_name
SECURITY DEFINER
AS select ...
```

There is currently no way to create these views through dbt-athena, forcing users to manage them outside of dbt.

### Solution

Add a new config flag `is_data_catalog_view` (default `False`) on the existing view materialization that conditionally emits MDV DDL instead of the standard view DDL.

**Usage:**

```sql
{{ config(
    materialized='view',
    is_data_catalog_view=True
) }}

select orderdate, sum(totalprice) as price
from {{ ref('orders') }}
group by orderdate
```

**Changes:**

1. **`dbt-athena/src/dbt/adapters/athena/impl.py`** — Register `is_data_catalog_view: bool = False` in `AthenaConfig` so dbt doesn't raise "unrecognized config" warnings.
2. **`dbt-athena/src/dbt/include/athena/macros/materializations/models/view/create_view_as.sql`** — New `athena__create_view_as` macro that branches on the flag: standard `CREATE OR REPLACE VIEW` when `False`, `CREATE OR REPLACE PROTECTED MULTI DIALECT VIEW ... SECURITY DEFINER` when `True`.
3. **`dbt-athena/tests/unit/test_create_view_as.py`** — Unit tests verifying the macro emits correct DDL for both paths (standard view, data catalog view, with/without contract enforcement).
4. **`dbt-athena/tests/functional/adapter/test_data_catalog_view.py`** — Functional tests verifying end-to-end creation and idempotency against a live Athena environment.

**Design decisions:**

- A config flag on the existing view materialization was chosen over a separate materialization type because MDVs differ from standard views only in the CREATE DDL — all other lifecycle operations (drop, replace, LF tags, persist_docs) are identical.
- No changes to relation types or drop logic — MDVs appear as `VIRTUAL_VIEW` in Glue, so existing `DROP VIEW IF EXISTS` works.
- The existing Lake Formation integration (`lf_tags_config`, `lf_grants`) works as-is with MDVs.

### Checklist

### DIFF
--- a/dbt-athena/src/dbt/include/athena/macros/materializations/models/view/create_view_as.sql
+++ b/dbt-athena/src/dbt/include/athena/macros/materializations/models/view/create_view_as.sql
@@ -2,13 +2,6 @@
   {%- set contract_config = config.get('contract') -%}
   {%- set is_data_catalog_view = config.get('is_data_catalog_view', False) -%}
 
-  {%- if is_data_catalog_view and contract_config.enforced -%}
-    {% do exceptions.raise_compiler_error(
-      "Data Catalog Views (is_data_catalog_view=True) do not support contract enforcement (contract.enforced=True). "
-      ~ "Please disable one of these options for model '" ~ model.name ~ "'."
-    ) %}
-  {%- endif -%}
-
   {%- if contract_config.enforced -%}
     {{ get_assert_columns_equivalent(sql) }}
   {%- endif -%}
--- a/dbt-athena/src/dbt/include/athena/macros/materializations/models/view/create_view_as.sql
+++ b/dbt-athena/src/dbt/include/athena/macros/materializations/models/view/create_view_as.sql
@@ -1,10 +1,28 @@
 {% macro athena__create_view_as(relation, sql) -%}
   {%- set contract_config = config.get('contract') -%}
+  {%- set is_data_catalog_view = config.get('is_data_catalog_view', False) -%}
+
+  {%- if is_data_catalog_view and contract_config.enforced -%}
+    {% do exceptions.raise_compiler_error(
+      "Data Catalog Views (is_data_catalog_view=True) do not support contract enforcement (contract.enforced=True). "
+      ~ "Please disable one of these options for model '" ~ model.name ~ "'."
+    ) %}
+  {%- endif -%}
+
   {%- if contract_config.enforced -%}
     {{ get_assert_columns_equivalent(sql) }}
   {%- endif -%}
+
+  {%- if is_data_catalog_view -%}
+  create or replace protected multi dialect view
+    {{ relation }}
+  security definer
+  as
+    {{ sql }}
+  {%- else -%}
   create or replace view
     {{ relation }}
   as
     {{ sql }}
+  {%- endif -%}
 {% endmacro %}
--- a/dbt-athena/tests/functional/adapter/test_data_catalog_view.py
+++ b/dbt-athena/tests/functional/adapter/test_data_catalog_view.py
@@ -1,0 +1,56 @@
+"""
+Functional tests for Data Catalog Views (Multi Dialect Views).
+
+Requires a live Athena environment with Lake Formation permissions
+configured to allow creating protected multi dialect views.
+"""
+
+import pytest
+
+from dbt.contracts.results import RunStatus
+from dbt.tests.util import run_dbt
+
+
+data_catalog_view_model_sql = """
+{{ config(
+    materialized='view',
+    is_data_catalog_view=True
+) }}
+
+select 1 as id, 'test' as name
+"""
+
+standard_view_model_sql = """
+{{ config(
+    materialized='view'
+) }}
+
+select 1 as id, 'test' as name
+"""
+
+
+class TestDataCatalogView:
+    """Verify that is_data_catalog_view=True produces a protected multi dialect view."""
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "data_catalog_view.sql": data_catalog_view_model_sql,
+            "standard_view.sql": standard_view_model_sql,
+        }
+
+    def test_data_catalog_view_creation(self, project):
+        results = run_dbt(["run"])
+        assert len(results) == 2
+
+        for result in results:
+            assert result.status == RunStatus.Success
+
+    def test_data_catalog_view_idempotent(self, project):
+        first_run = run_dbt(["run"])
+        assert len(first_run) == 2
+
+        second_run = run_dbt(["run"])
+        assert len(second_run) == 2
+        for result in second_run:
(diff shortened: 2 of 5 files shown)


dbt-athena/src/dbt/include/athena/macros/materializations/models/view/create_view_as.sql
```diff
@@ -1,10 +1,21 @@
...
+  {%- set is_data_catalog_view = config.get('is_data_catalog_view', False) -%}
+
   {%- if contract_config.enforced -%}
     {{ get_assert_columns_equivalent(sql) }}
   {%- endif -%}
+
+  {%- if is_data_catalog_view -%}
+  create or replace protected multi dialect view
+    {{ relation }}
+  security definer
```
iconara, line 12: One thing I've found is tricky with definer security with these views is that what you really want is often a different role than the one that runs your pipeline. That role tends to have a lot of permissions since it needs them to do all sorts of things. A feature I'm missing from these views is the ability to say `security 'arn:....:role/some-scoped-down-role'`. To achieve that today, you have to switch roles before running the query.

Since you have a need for this feature, you have obviously used them more than me, and probably thought more about this. I'm curious to hear how you do this in practice.

dbt-athena/tests/functional/adapter/test_data_catalog_view.py
```diff
@@ -0,0 +1,56 @@
+"""
+Functional tests for Data Catalog Views (Multi Dialect Views).
+
+Requires a live Athena environment with Lake Formation permissions
+configured to allow creating protected multi dialect views.
```
iconara, line 5: This may be an issue for CI and for anyone running the tests while developing. Flagging this for @colin-rogers-dbt to think about.

dbt-athena/src/dbt/include/athena/macros/materializations/models/view/create_view_as.sql
adavoudi: At our company, we use team-specific IAM roles for running dbt pipelines in Athena, so the definer is already scoped down to that team's domain.

More importantly, we rely on Lake Formation tag-based access control on the view itself (via lf_tags_config) to govern what downstream consumers can see, independent of the definer's permissions. I think of a Data Catalog View the same way I think of a table: the creator needs permissions to build it, but what consumers see is controlled by LF tags/grants on the resulting object itself.

That said, I agree that being able to specify an explicit definer role ARN would be a useful addition.

dbt-athena/tests/functional/adapter/test_data_catalog_view.py
adavoudi: @iconara I've marked this test to skip:

```
@pytest.mark.skip(reason="requires Lake Formation permissions not available in CI")
```

khoaanguyenn:
👋 Hi there, I'm looking for this too feature too, are we still working on this?

jb3:
Hi there! We're also very interested in this feature so we can adopt MDVs without needing to manage externally. Is this still something that is being worked on? Is there anything to help with to push it across the line? Thanks!

PR #1872, opened 2026-04-16
bigquery: gate MicrobatchConcurrency on use_concurrent_microbatch behavior flag

### LINKED ISSUE #766: [Feature] Implement MIcrobatchConcurrency for dbt-bigquery

### Describe the feature

Concurrency in bigquery microbatch like in snowflake

_No response_

### Who will this benefit?

Microbatch users

_No response_

### Anything else?

_No response_

andtrott:
## Resolves / relates to

Resolves [#766](https://github.com/dbt-labs/dbt-adapters/issues/766).

## Why

The BigQuery adapter currently does not declare `Capability.MicrobatchConcurrency`, so dbt-core runs microbatch batches sequentially and emits the warning:

> `Found N microbatch models with the concurrent_batches config set to true, but the bigquery adapter does not support running batches concurrently. Batches will be run sequentially.`

For models whose `partition_by` produces disjoint partition ranges per batch (the common case for hourly microbatches), sequential execution multiplies wall-clock by the batch count with no correctness benefit. Users in #766 and elsewhere report this makes microbatch on BigQuery prohibitively slow for high-cardinality hourly pipelines.

## What this PR does

Mirrors the [dbt-databricks concurrent-microbatch rollout pattern](https://github.com/databricks/dbt-databricks/) exactly:

1. Adds a `USE_CONCURRENT_MICROBATCH` `BehaviorFlag` (default `False`) with a description documenting the correctness contract.
2. Registers the flag in `BigQueryAdapter._behavior_flags`.
3. Overrides `supports()` to return `True` for `Capability.MicrobatchConcurrency` iff the flag is enabled. All other capability checks delegate to `super().supports()`.

Opt-in only. Default behavior for existing projects is unchanged.

## Why the correctness contract is safe on BigQuery

- BigQuery applies concurrent DML under **snapshot isolation** — `MERGE` statements writing to disjoint partitions can run concurrently and the final table state is deterministic.
- The 1,000 DML statements/day per-table quota was removed in 2020 ([DML without limits, now in BigQuery](https://cloud.google.com/blog/products/data-analytics/dml-without-limits-now-in-bigquery)).
- The microbatch materialization (`bq_generate_microbatch_build_sql`) has no serial-ordering assumptions — it delegates per batch to `bq_insert_overwrite_sql`.

The behavior flag documents that users are responsible for ensuring `partition_by` produces disjoint partitions across batches (true for `TIMESTAMP_TRUNC` expressions and similar; violated only by partitioning on non-batch-scoped expressions).

## Reviewer notes

- Single-file change, ~20 LoC.
- Pattern 1:1 with `dbt-databricks` — existing maintainer familiarity.
- No changes to existing capabilities or materialization macros.
- Default `False` → opt-in only → no behavior change for users who don't set the flag.

### DIFF
--- a/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
@@ -143,6 +143,19 @@ BIGQUERY_USE_STANDARD_SQL_FOR_PARTITIONS = BehaviorFlag(
     ),
 )
 
+USE_CONCURRENT_MICROBATCH = BehaviorFlag(
+    name="use_concurrent_microbatch",
+    default=False,
+    description=(
+        "Enable concurrent execution of microbatch incremental batches on BigQuery. "
+        "When enabled, dbt will run microbatch batches in parallel threads instead "
+        "of sequentially. Safe when batches write to disjoint partitions (as implied "
+        "by a well-chosen `partition_by`) — BigQuery applies concurrent MERGE "
+        "statements atomically under snapshot isolation. Users are responsible for "
+        "ensuring batches do not overlap in partition_key range."
+    ),
+)  # type: ignore[typeddict-item]
+
 _dataset_lock = threading.Lock()
 
 
@@ -209,6 +222,12 @@ class BigQueryAdapter(BaseAdapter):
         }
     )
 
+    def supports(self, capability: Capability) -> bool:
+        # Gate MicrobatchConcurrency on the use_concurrent_microbatch behavior flag.
+        if capability == Capability.MicrobatchConcurrency:
+            return bool(self.behavior.use_concurrent_microbatch)
+        return super().supports(capability)
+
     def __init__(self, config, mp_context: SpawnContext) -> None:
         super().__init__(config, mp_context)
         self.connections: BigQueryConnectionManager = self.connections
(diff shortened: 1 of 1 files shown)


QMalcolm (changes requested):
Hi @andtrott! Thank you for doing this work ❤️ 

I think we can, and should, simplify this. Specifically, we should remove the behavior flag and the related logic. I talked with broader team to doublecheck my gut on this, and we're in agreement on this.

At a high level we get _why_ this might be considered a behavior change to some, but we don't actually think it is. First it was always intended for concurrent batches to be supported by dbt-bigquery, as noted in the documentation that "other adapter support was coming soon". Second there is already a hard opt-out config, `concurrent_batches=Flase`, that can be set. And core by default _won't_ run concurrent batches if there is a `this` referenced used in the microbatch model.

kasper0406:
@andtrott wondering if you have the bandwidth to get this through? My understanding of the above comment is that it would be a very small code change. Would be awesome to be able to parallelize these micro batches

amishas157:
I created https://github.com/dbt-labs/dbt-adapters/pull/2007 using patch in original PR. Then tested with my dbt models(with copy_partitions set to true). 
I do not know how would parallel queries work if copy_partitions is not set to true.

The job run time went from 4 hours to just 36 mins 🤯 . It will be really great if we can merge the original PR by @andtrott  🥳 
Happy to help test further if useful. 🙌

kevinwmerritt:
This is very much needed in order to adopt dbt microbatches when using bigquery.

graciegoheen:
I'm going to close this out, as I think https://github.com/dbt-labs/dbt-adapters/pull/2007 supersedes this - thanks for the start!

PR #1892, opened 2026-04-23
feat(athena): Add S3 Table Bucket support

### LINKED ISSUE #1186: [Feature] Add S3 Tables support to dbt-athena adapter

### Describe the feature

Currently, you can't use S3 Tables with the dbt-athena adapter because AWS requires the special catalog "s3tablescatalog/<bucketname>" as the Catalog. 

Normally, Catalog is something like "AwsDataCatalog" or empty (defaults to AwsDataCatalog) or a Federated Query connector like "mysql-db" or a DynamoDB table. 

But, in the API, SDKs (boto3, etc), and CLI, if you set catalog to `s3tablescatalog/<bucketname>` then it will assume the actual catalog is AwsDataCatalog and then use the new sub-catalog of `s3tablecatalog...`.  This appears to be a way they shoe-horned querying S3 Table buckets in Athena without having to change the API.

However, dbt-athena currently checks to see if the catalog is `AwsDataCatalog` and, if not, does special logic like call GetDataCatalog (see: [impl.py lines 676-691](https://github.com/dbt-labs/dbt-athena/blob/main/dbt-athena/src/dbt/adapters/athena/impl.py#L676-L691)).  If you try to set the DBT database to `s3tablescatalog/...` DBT will throw an error and say that the call to GetDataCatalog API failed. 

It's possible to query Athena against `AwsDataCatalog` and using the database/schema identifier of `"s3tablescatalog/bucketname"."schemaname"."tablename"` in the query. But since DBT quotes all identifiers, it ends up trying to query `""s3tablescatalog/bucketname"."schemaname"."tablename""` and doesn't work.

### Who will this benefit?

Anyone using Athena and the new S3 Tables product in AWS

Yes, but there are many other calls to GetDataCatalog so I'm afraid of breaking other code paths

### Anything else?

_No response_

skillicinski:
resolves #1186
docs N/A

### Problem

dbt-athena cannot target AWS S3 Table Buckets (GA since Dec 2024) because:

1. CTAS always emits `location`, `table_type`, and `is_external` in the WITH clause — all forbidden for S3 Table Buckets
2. S3 cleanup operations (`DeleteObjects`, `ListObjectsV2`) fail against managed storage
3. `ALTER TABLE RENAME` is blocked by AWS, breaking the zero-downtime swap
4. `DROP TABLE` via SQL is blocked by AWS
5. `get_table_type()` checks `RELATION_TYPE_MAP` before `Parameters.table_type`, so S3TB tables with `TableType="customer"` raise ValueError before the Iceberg check runs

A community member confirmed these blockers in #1186.

### Solution

**Detection**: New `@available` method `is_s3_table_bucket(database)` detects S3TB catalogs by checking for `s3tablescatalog/` in the `catalog-id` parameter from `GetDataCatalog`. `_get_data_catalog()` is cached via `lru_cache` (was called 14+ times uncached per run), with quote-stripping normalized before the cache key.

**CTAS DDL** (`create_table_as.sql`): S3TB targets omit `table_type`, `is_external`, and `location` from the WITH clause. `generate_s3_location()` and `delete_from_s3()` are skipped. `table_type` auto-defaults to `iceberg` (errors if explicitly set to `hive`).

**Drop path** (`relation.sql`): S3TB targets skip `clean_up_table()` (no S3 object deletion on managed storage) and call `delete_from_glue_catalog()` directly via Glue API with compound `CatalogId`. Defensive guard on `athena__rename_relation` raises error since AWS blocks RENAME.

**Materializations**: Table and incremental `--full-refresh` use drop-and-recreate instead of rename-swap (brief downtime — known limitation). Normal incremental merge/append work unchanged. `expire_glue_table_versions` is skipped for S3TB. Resolved `table_type` is pushed back into config so downstream macros (e.g. `on_schema_change.sql`) see `iceberg` not `hive`.

**Error guards**: Seeds, views, and Python models raise clear `CompilationError` messages for S3TB targets.

**Bug fixes (pre-existing)**:
- `get_table_type()`: reordered to check `Parameters.table_type` before `TableType` — S3TB tables report `TableType="customer"` which is not in `RELATION_TYPE_MAP`
- `list_schemas()`: now passes `CatalogId` to `glue.get_databases()` (was missing)
- `safe_create_table_as`: `force_batch` branch used wrong variable name for return value

**User configuration**:
```bash
# One-time prerequisite
aws athena create-data-catalog --name my_s3_tables --type GLUE \
  --parameters catalog-id=<account-id>:s3tablescatalog/<bucket-name>
```
```yaml
# profiles.yml
my_project:
  target: dev
  outputs:
    dev:
      type: athena
      database: my_s3_tables    # named Athena data catalog
      schema: my_namespace      # S3 Table Bucket namespace
      s3_staging_dir: s3://my-staging-bucket/athena-results/
      region_name: us-east-1
```

**Known limitations**: No zero-downtime swap (RENAME blocked by AWS), seeds/views/Python models not supported on S3TB, `DROP SCHEMA CASCADE` untested on S3TB.

**Testing**: 226 unit tests pass (216 existing + 10 new). 8 integration tests pass against a real S3 Table Bucket. Code quality (Black/Flake8/MyPy) clean.

### Checklist


dbt-athena/src/dbt/include/athena/macros/materializations/models/table/create_table_as.sql
```diff
@@ -204,6 +223,9 @@
     {%- if language != 'sql' -%}
         {{ return(create_table_as(temporary, relation, compiled_code, language)) }}
     {%- elif force_batch -%}
+      {%- if adapter.is_s3_table_bucket(relation.database) -%}
+        {% do exceptions.raise_compiler_error("force_batch is not supported with S3 Table Buckets.") %}
+      {%- endif -%}
       {%- do create_table_as_with_partitions(temporary, relation, compiled_code, language) -%}
       {%- set query_result = relation ~ ' with many partitions created' -%}
```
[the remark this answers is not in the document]
skillicinski: Seems to be mostly cosmetic as jinja2 should render an `Undefined` object as an empty string by default, if I've understood that correctly.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -1,8 +1,21 @@
 {% materialization incremental, adapter='athena', supported_languages=['sql', 'python'] -%}
   {% set raw_strategy = config.get('incremental_strategy') or 'insert_overwrite' %}
   {% set is_microbatch = raw_strategy == 'microbatch' %}
-  {% set table_type = config.get('table_type', default='hive') | lower %}
+  {% set raw_table_type = config.get('table_type') %}
```
nicor88, line 4: any reason why you are dropping the default `hive`, that can be a breaking change for users that use defaults.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -1,8 +1,21 @@
...
+  {% set is_s3tb = adapter.is_s3_table_bucket(database) %}
```
nicor88: not big fan of shorcuts, can you please use `is_s3_table`, saving few charathers do not change me.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
```diff
@@ -67,7 +80,8 @@
   -- Must use s3_data_naming schema_table_unique in order to support high availability --
   -- on a full fresh for an incremental iceberg table --
   {%- elif (
-    should_full_refresh()
+    not is_s3tb
```
nicor88: can you add a little comment on why s3 tables are excluded here?

dbt-athena/tests/functional/adapter/test_s3_table_bucket.py
```diff
@@ -0,0 +1,199 @@
...
+pytestmark = pytest.mark.skipif(
+    not S3TB_DATABASE,
+    reason="DBT_TEST_ATHENA_S3TB_DATABASE not set — skipping S3 Table Bucket integration tests",
+)
+
+
+TABLE_MODEL = """
+{{ config(
+    materialized='table',
+    format='parquet'
```
nicor88, line 27: isn't there something missing here? aren't s3 tables supporting only iceberg tables? therefore I expect that `table_type` should be set as iceberg.

nicor88:
@skillicinski looks great, I didn't play a lot with S3 tables, only a bit when they were launched in 2024, anyhow as far as I understand S3 tables are pretty much native iceberg tables, shouldn't we add a table_type parameter override that force the table_type to be set as iceberg, when an S3 table is detected? this will remove the burden to the users to set iceberg tables expeplicitely. 

Also, I like the idea of automatic detecting S3 tables, but did you consider to add another "explicit" parameter, maybe called `is_s3_table=true`, which can ovverride the automatic detection? generally speaking both IMHO make sense.

skillicinski:
> @skillicinski looks great, I didn't play a lot with S3 tables, only a bit when they were launched in 2024, anyhow as far as I understand S3 tables are pretty much native iceberg tables, shouldn't we add a table_type parameter override that force the table_type to be set as iceberg, when an S3 table is detected? this will remove the burden to the users to set iceberg tables expeplicitely.
> 
> Also, I like the idea of automatic detecting S3 tables, but did you consider to add another "explicit" parameter, maybe called `is_s3_table=true`, which can ovverride the automatic detection? generally speaking both IMHO make sense.

hey @nicor88, appreciate the feedback and thanks for taking time to review!

you're spot on. we're leaning heavily into the S3 Table buckets at the company so this would be such a helpful addition, AWS promises better latency than Iceberg on gen-purpose S3 - let's see!

you're right, i think both your points make sense. will look over your comments now and push some updates. thanks again!

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
skillicinski: it felt natural to let that default flow into an if-else so this should catch omitted `table_type` and set it to `hive`on R10-R11

my assumption was that this would hold up but i don't have a strong opinion so happy to change it to your recommendation

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
skillicinski: yes, certainly

dbt-athena/tests/functional/adapter/test_s3_table_bucket.py
skillicinski: the idea was to test that the auto-detection works as intended when we omit `table_type`. added a comment in code to clarify, do you think that's enough?

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
skillicinski: of course, done!

dbt-athena/tests/functional/adapter/test_s3_table_bucket.py
nicor88: Make sense, anyhow what is unclear to me if this automatic detection allow to have hybrid setup, it should I suppose, but I want to double check with you.

dbt-athena/src/dbt/include/athena/macros/materializations/models/incremental/incremental.sql
nicor88: I miss the line [R11](https://github.com/dbt-labs/dbt-adapters/pull/1892/changes/BASE..e9911375b0acf33124e170a2dc207d543f27669d#diff-db30bb8221f224c6c8e021c69ab09fadfdd0a3ac363bca6ce26b382a5c8ac20eR11), and I believe that we should be good, as seems backwards compatible.

dbt-athena/tests/functional/adapter/test_s3_table_bucket.py
skillicinski: Nice! Yes, the automatic detection is still on a per model basis by design. The automatic `table_type` resolves on the model database/catalog and overrides project default. The main thing is that the IAM used needs access to each catalog but I feel like users expect that to be the case.

nicor88:
@colin-rogers-dbt do you think that you can set an S3 Table bucket to be used in the CI to be able to test this feature? 
Having a real S3 table bucket allows to have the CI more resilient, and catching relevant issues before going live, unfortunately unit tests cover some aspects, but not a complete end to end environment.

skillicinski:
Hey @colin-rogers-dbt @xuliangs, anything I can do to move this along?

colin-k-rogers:
@skillicinski - apologies for the radio silence here, we want to update `dbt-athena` to leverage the catalogs framework we have for other adapters before merging this as we'll have better primitives/patterns to support this feature. 

Tracking that work here: https://github.com/dbt-labs/dbt-adapters/issues/1989

dataders:
superceded by https://github.com/dbt-labs/dbt-adapters/pull/2047

PR #1908, opened 2026-04-30
Fix _dbt_max_partition false-positive detection in BigQuery incremental

### LINKED ISSUE #1907: [Bug] BigQuery incremental can inject _dbt_max_partition from comment text and fail first run
### Is this a new bug?

### Which package is affected?

- `dbt-bigquery` (in `dbt-labs/dbt-adapters` monorepo)

### Current Behavior

For an incremental model on BigQuery (`insert_overwrite` + partitioned), if the literal token `_dbt_max_partition` appears **only inside SQL comments**, dbt still injects `_dbt_max_partition` script logic.

On a first run (target table does not yet exist), the injected block queries `{{ this }}` before creation and fails:

```sql
declare _dbt_max_partition date default (
  select max(partition_col) from `project.dataset.my_incremental_model`
  where partition_col is not null
);
```

Error:

```text
Not found: Table project:dataset.my_incremental_model was not found in location US
```

### Expected Behavior

`_dbt_max_partition` helper injection should not be triggered by occurrences inside comments (or string literals). It should be based on executable SQL references only.

### Steps To Reproduce

1. Create an incremental model:

```sql
{{
  config(
    materialized='incremental',
    incremental_strategy='insert_overwrite',
    partition_by={'field': 'partition_col', 'data_type': 'date', 'granularity': 'day'},
    on_schema_change='append_new_columns'
  )
}}

/* This text mentions _dbt_max_partition only in a comment */

select date('2026-04-30') as partition_col
```

2. Ensure target table does not exist.
3. Run `dbt run -s <model>`.
4. Observe compiled script contains `declare _dbt_max_partition ... from <target table>` and fails with table-not-found.
5. Remove/rename the comment token (without changing executable SQL), rerun, and model succeeds.

### Relevant notes

- Reproduced on `dbt-core==1.11.2` and `dbt-bigquery==1.11.0`.
- Dataset location was `US`; this does not appear to be a cross-region issue.
- Related but different historical issue: #583.

### Environment

- OS: macOS
- Python: 3.10.x
- dbt-core: 1.11.2
- dbt-bigquery: 1.11.0

ElhadiCherifi95:
## Summary
- fix `declare_dbt_max_partition` to avoid matching `_dbt_max_partition` mentions that only appear in SQL comments/quoted string literals
- only emit the declaration when the target relation exists
- add a functional regression test that runs an `insert_overwrite` incremental model where `_dbt_max_partition` appears only in comments and verifies first + second runs succeed

## Root Cause
`declare_dbt_max_partition` previously used a raw substring check:

```jinja
'_dbt_max_partition' in compiled_code
```

This matched comment text, causing declaration injection even when the model did not actually reference `_dbt_max_partition` in executable SQL.

On first run, that declaration queried the target table before it existed, producing a table-not-found error.

## Changes
- `dbt-bigquery/src/dbt/include/bigquery/macros/materializations/incremental_strategy/common.sql`
  - sanitize `compiled_code` by removing comments and single-quoted string literals before detection
  - use boundary-aware regex for token detection
  - guard declaration behind `load_relation(relation) is not none`
- `dbt-bigquery/tests/functional/adapter/incremental/incremental_strategy_fixtures.py`
  - add `overwrite_comment_only_max_partition_token_sql` fixture
- `dbt-bigquery/tests/functional/adapter/incremental/test_incremental_strategies.py`
  - add `TestBigQueryMaxPartitionCommentToken` regression test

## Issue
- Closes #1907

### DIFF
--- a/dbt-bigquery/src/dbt/include/bigquery/macros/materializations/incremental_strategy/common.sql
+++ b/dbt-bigquery/src/dbt/include/bigquery/macros/materializations/incremental_strategy/common.sql
@@ -1,7 +1,20 @@
 {% macro declare_dbt_max_partition(relation, partition_by, compiled_code, language='sql') %}
 
   {#-- TODO: revisit partitioning with python models --#}
-  {%- if '_dbt_max_partition' in compiled_code and language == 'sql' -%}
+  {%- if language != 'sql' -%}
+    {{ return('') }}
+  {%- endif -%}
+
+  {#-
+    Ignore mentions in comments/quoted literals when deciding whether to declare
+    _dbt_max_partition.
+  -#}
+  {%- set compiled_code_without_comments = modules.re.sub("(?s)/\\*.*?\\*/|--[^\\n\\r]*", " ", compiled_code) -%}
+  {%- set compiled_code_without_comments_or_strings = modules.re.sub("'(?:''|[^'])*'", " ", compiled_code_without_comments) -%}
+  {%- set uses_dbt_max_partition = modules.re.search("(^|[^A-Za-z0-9_])_dbt_max_partition([^A-Za-z0-9_]|$)", compiled_code_without_comments_or_strings) is not none -%}
+  {%- set relation_exists = load_relation(relation) is not none -%}
+
+  {%- if uses_dbt_max_partition and relation_exists -%}
 
     declare _dbt_max_partition {{ partition_by.data_type_for_partition() }} default (
       select max({{ partition_by.field }}) from {{ this }}
--- a/dbt-bigquery/tests/functional/adapter/incremental/incremental_strategy_fixtures.py
+++ b/dbt-bigquery/tests/functional/adapter/incremental/incremental_strategy_fixtures.py
@@ -190,6 +190,28 @@ where date_day >= _dbt_max_partition
 {% endif %}
 """.lstrip()
 
+overwrite_comment_only_max_partition_token_sql = """
+{{
+    config(
+        materialized="incremental",
+        incremental_strategy='insert_overwrite',
+        on_schema_change='append_new_columns',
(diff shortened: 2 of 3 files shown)


vittorfp:
We hit this exact bug in production yesterday (2026-08-13): the first-ever deploy of a new `insert_overwrite` incremental model failed with

```
Not found: Table <project>:<release_dataset>.<model> was not found in location EU at [2:41]
```

because the model's SQL header **comment** mentioned `_dbt_max_partition` in prose (documenting how our incremental-window macro works internally). The raw substring check in `declare_dbt_max_partition` matched the comment text and emitted the `declare ... default (select max(...) from <target>)` block against a table that did not exist yet. Reproduced with and without `--full-refresh`; our stopgap was rewording the comment so the literal token doesn't appear — exactly the landmine this PR removes.

Two review notes from reading the diff:

- GoogleSQL also supports `#` line comments ([lexical docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#comments)), so the sanitizer currently misses `# ... _dbt_max_partition ...`. Adding `|#[^\n\r]*` to the first `re.sub` pattern would close that hole.
- This should also close #580 — the same bug reported back in Feb 2024, so this fix has been ~2.5 years in demand.

Would appreciate a maintainer workflow-approval so CI can actually run on this one — the patch shape looks right to us and we're happy to test a build against our production repro.

vittorfp:
Since this branch now conflicts with `main` and can't be updated by non-authors, I've opened #2123 — a rebase of this fix onto current main, with you kept as commit co-author, plus `#` line-comment support (GoogleSQL supports those too) and a changelog entry. Hopefully one of the two can land; this bug has been open since Feb 2024 (#580) and bit our production deploy this week.

PR #1912, opened 2026-05-04
feat(redshift): add drop_without_cascade profile option

### LINKED ISSUE #1772: feat(redshift): add profile option to drop relations without CASCADE
## Summary

Add a profile credentials option (`drop_without_cascade` or similar) to omit `CASCADE` from `DROP TABLE/VIEW/MATERIALIZED VIEW` statements in dbt-redshift.

## Background

All three dbt-redshift drop macros unconditionally append `CASCADE`:

**`macros/relations/table/drop.sql`** (line 2):
```sql
drop table if exists {{ relation }} cascade
```

**`macros/relations/view/drop.sql`** (line 2):
```sql
drop view if exists {{ relation }} cascade
```

**`macros/relations/materialized_view/drop.sql`** (line 2):
```sql
drop materialized view if exists {{ relation }} cascade
```

`CASCADE` instructs Redshift to automatically drop all dependent objects (views, etc.). On clusters with many objects, Redshift must walk the full dependency graph to resolve each `DROP ... CASCADE`, which is expensive. For users who exclusively use **unbound views** (or have no downstream views at all), there are no dependents to cascade to — the overhead is pure waste.

Additionally, `CASCADE` is the root cause of the concurrent-transaction race condition that requires the global thread lock (see related issue). Removing `CASCADE` also opens the door to removing the lock.

## Proposed Change

Add a boolean profile credential field `drop_without_cascade` (default `False`) that, when `True`, emits `DROP ... RESTRICT` (or simply omits `CASCADE`) instead.

### Files to change

**`dbt-redshift/src/dbt/adapters/redshift/connections.py`**

Add to `RedshiftCredentials` dataclass (~line 190):
```python
# Drop behavior
drop_without_cascade: bool = False
```

Add to `_connection_keys()` (~line 228):
```python
"drop_without_cascade",
```

**`macros/relations/table/drop.sql`**:
```sql
{%- macro redshift__drop_table(relation) -%}
    drop table if exists {{ relation }} {% if not adapter.config.credentials.drop_without_cascade %}cascade{% endif %}
{%- endmacro -%}
```

**`macros/relations/view/drop.sql`**:
```sql
{%- macro redshift__drop_view(relation) -%}
    drop view if exists {{ relation }} {% if not adapter.config.credentials.drop_without_cascade %}cascade{% endif %}
{%- endmacro -%}
```

**`macros/relations/materialized_view/drop.sql`**:
```sql
{% macro redshift__drop_materialized_view(relation) -%}
    drop materialized view if exists {{ relation }} {% if not adapter.config.credentials.drop_without_cascade %}cascade{% endif %}
{%- endmacro %}
```

> **Note:** The exact mechanism for passing credentials into Jinja macros should follow the existing pattern used by `redshift__use_show_apis()` in `macros/metadata/helpers.sql`. Alternatively this could be implemented as a Python-level override of the drop macros using `execute()` directly in `impl.py`.

## Safety

- Default is `False` — no change in behavior for existing users.
- Safe only for projects that don't have dependent views downstream of dropped relations (e.g., projects using only unbound views, or where dbt manages all dependencies).
- Must be documented with a clear warning: if a dependent object exists and `CASCADE` is omitted, Redshift will raise an error rather than silently dropping it.
- Pairs naturally with `drop_without_lock` (see related issue): users who set both get maximum DROP performance with minimal risk when their project has no downstream view dependencies.

## Related

- Related to the `drop_without_lock` feature request — the thread lock exists specifically because of `CASCADE` race conditions.

colin-k-rogers:
## Summary

Adds a `drop_without_cascade` boolean profile credential (default `False`) for dbt-redshift. When enabled, `DROP TABLE/VIEW/MATERIALIZED VIEW` statements are emitted without `CASCADE`.

Closes #1772.

## Why

On Redshift clusters with many objects, resolving the `CASCADE` dependency graph for every `DROP` is expensive. For projects that exclusively use unbound views (or have no downstream view dependents at all), the cost is pure overhead. This option lets those users opt out.

## Changes

- **`connections.py`**: New `drop_without_cascade: bool = False` field on `RedshiftCredentials`, exposed via `_connection_keys()`.
- **`impl.py`**: New `@available drop_without_cascade()` adapter method, mirroring the `use_show_apis()` pattern.
- **`macros/metadata/helpers.sql`**: New `redshift__drop_without_cascade()` Jinja helper.
- **Drop macros** (`table/drop.sql`, `view/drop.sql`, `materialized_view/drop.sql`): Conditionally emit `cascade` based on the helper.
- **Unit tests**: Added `TestDropWithoutCascade` covering the default-`False` and credential-`True` paths.
- **Changelog**: Added Features entry.

## Safety

- Default is `False` — no behavior change for existing users.
- When `True` and a dependent object exists, Redshift raises an error rather than silently dropping it (the desired behavior for opt-in users).
- Pairs naturally with `allow_concurrent_drops` for projects without downstream dependencies.

## Test plan

- [x] Unit tests for default and opt-in credential values
- [ ] Integration tested against a real Redshift cluster
- [ ] Manual verification that DDL emitted matches expectation in both modes


colin-k-rogers:
Added model-level config support per review discussion. Users can now set `drop_without_cascade` per model:

```sql
{{ config(materialized='table', drop_without_cascade=true) }}
```

**Precedence:** model config > profile credential > default (`False`). A model-level setting (either `True` or `False`) always wins over the profile, so users can globally opt in via the profile and selectively opt out for specific models with downstream dependents — or vice versa.

New tests cover both directions of model/profile precedence.

colin-k-rogers:
## CI investigation

Two distinct issues in the failing run:

### 1. Bugs in my new tests (now fixed)

- **Schema-name substring confound:** the auto-generated test schema (e.g. `test1234..._test_drop_without_cascade`) contains the substring `cascade`, so `"cascade" in line.lower()` always matched regardless of whether the keyword was actually emitted. Fixed by stripping quoted identifiers and matching `\bcascade\b` (word boundary).
- **Model-level config didn't propagate:** the drop happens during the materialization swap via `adapter.drop_relation` → `execute_macro('drop_relation')`, which builds a fresh runtime macro context. `config.get('drop_without_cascade')` in that context isn't the calling model's config, so the model-level setting silently no-op'd. The CI failure on `view_opt_in__dbt_backup` (still emitted CASCADE) caught this. **Reverted model-level support — profile-only.** Plumbing model config through the dispatch path is a bigger change than is warranted here.

Locally with `hatch run integration-tests -- tests/functional/test_drop_without_cascade.py`:
```
2 passed, 48 warnings in 28.23s
```
And `hatch run unit-tests` for both unit modules: `10 passed`.

### 2. Pre-existing Redshift flake (unrelated)

The other failures (`TestEphemeralRedshift`, `TestPersistDocs`, `TestGetCatalog`, etc.) all show `Database Error: cache lookup failed for attribute 1 of relation 64439008` during `docs generate` → `Building catalog`. This is a Redshift internal staleness symptom in catalog queries (`pg_attribute`/`pg_class`), nothing to do with the drop macros.

I reproduced it locally on my branch even running serially (`-n0`):
```
FAILED tests/functional/adapter/test_basic.py::TestEphemeralRedshift::test_ephemeral
FAILED tests/functional/adapter/test_persist_docs.py::TestPersistDocs::test_has_comments_pglike
FAILED tests/functional/adapter/catalog_tests/test_get_catalog.py::TestGetCatalog::test_get_one_catalog_by_relations
```
Same `cache lookup failed for attribute 1 of relation` error. None of these tests exercise drop_relation paths, and my changes only touch:

- `drop_without_cascade` field on `RedshiftCredentials`
- `RedshiftAdapter.drop_without_cascade()` method
- `redshift__drop_{table,view,materialized_view}` macros conditionally appending `cascade`

Suggest a re-run to confirm the catalog failures are flake.

dbt-redshift/tests/functional/test_drop_without_cascade.py
```diff
@@ -0,0 +1,94 @@
...
+        assert drop_lines, f"expected at least one DROP VIEW for my_view; logs:\n{logs}"
+        for line in drop_lines:
+            assert _has_cascade_keyword(line), line
+
+
+class TestDropWithoutCascadeProfileLevel:
+    """`drop_without_cascade=True` in the profile → DROP must omit CASCADE."""
+
+    @pytest.fixture(scope="class")
+    def dbt_profile_target(self):
```
tauhid621: Not sure if we need the whole config here.

Perhaps we can keep it simple like other tests?

```
@pytest.fixture(scope="class")
    def profiles_config_update(self, dbt_profile_target):
        outputs = {"default": dbt_profile_target}
        outputs["default"]["drop_without_cascade"] = True
        return outputs
```

dbt-redshift/src/dbt/include/redshift/macros/relations/view/drop.sql
```diff
@@ -1,3 +1,3 @@
 {%- macro redshift__drop_view(relation) -%}
-    drop view if exists {{ relation }} cascade
+    drop view if exists {{ relation }} {% if not redshift__drop_without_cascade() %}cascade{% endif %}
```
tauhid621: NIT: better whitespace 
```suggestion
    drop view if exists {{ relation }}{% if not redshift__drop_without_cascade() %} cascade{% endif %}
```

dbt-redshift/src/dbt/include/redshift/macros/relations/table/drop.sql
```diff
@@ -1,3 +1,3 @@
 {%- macro redshift__drop_table(relation) -%}
-    drop table if exists {{ relation }} cascade
+    drop table if exists {{ relation }} {% if not redshift__drop_without_cascade() %}cascade{% endif %}
```
tauhid621: NIT: better whitespace
```suggestion
    drop table if exists {{ relation }}{% if not redshift__drop_without_cascade() %} cascade{% endif %}
```

dbt-redshift/src/dbt/include/redshift/macros/relations/materialized_view/drop.sql
```diff
@@ -1,3 +1,3 @@
 {% macro redshift__drop_materialized_view(relation) -%}
-    drop materialized view if exists {{ relation }} cascade
+    drop materialized view if exists {{ relation }} {% if not redshift__drop_without_cascade() %}cascade{% endif %}
```
tauhid621: NIT: better whitespace
```suggestion
    drop materialized view if exists {{ relation }}{% if not redshift__drop_without_cascade() %} cascade{% endif %}
```

tauhid621 (approved):
Looks good. Left minor comments

dbt-redshift/tests/functional/test_drop_without_cascade.py
colin-k-rogers: Addressed in 57ee645.

dbt-redshift/src/dbt/include/redshift/macros/relations/view/drop.sql
colin-k-rogers: Addressed in 57ee645.

dbt-redshift/src/dbt/include/redshift/macros/relations/table/drop.sql
colin-k-rogers: Addressed in 57ee645.

dbt-redshift/src/dbt/include/redshift/macros/relations/materialized_view/drop.sql
colin-k-rogers: Addressed in 57ee645.

PR #1915, opened 2026-05-05
fix(redshift): enable retry_all in functional test profile to reduce catalog flakes

colin-k-rogers:
## Summary

- Adds `retry_all: True` to the `dbt_profile_target` fixture in `dbt-redshift/tests/functional/conftest.py`

## Why

The shared Redshift CI cluster is not strongly consistent under concurrent DDL. When `pytest-xdist` runs integration tests in parallel (`-n auto`), multiple workers simultaneously create, populate, and tear down schemas against the same cluster. Catalog queries that join `pg_catalog.pg_class` to `pg_catalog.pg_attribute` can encounter a race where a relation's OID becomes stale between the two reads, producing:

```
Database Error
  cache lookup failed for attribute 1 of relation 64439008
```

This error is a `DbtDatabaseError` but does **not** match any of the hard-coded retry patterns (`could not open relation with OID`, `conflict with concurrent transaction`, `schema does not exist`), so it was not being retried.

Setting `retry_all: True` tells the Redshift connection manager to retry any `DbtDatabaseError`, drawing from the existing `retries: 6` budget already in the profile. The `cache lookup failed` error is transient by definition — a single retry resolves it — so this change should eliminate the flakes without masking real failures.

## What's not changed

- No production adapter code is touched
- Schema isolation is already correct (each test class gets a unique schema via the `unique_schema` fixture); this is purely a retry-on-transient-error fix

## Reviewer notes

`TestRetryAll` in `test_exception_handling.py` already exercises the `retry_all` code path, so the mechanic is tested. The affected flaky tests are catalogued in PR #1912.


dbt-redshift/tests/functional/adapter/catalog_tests/test_get_catalog.py
```diff
@@ -5,6 +5,11 @@
 
 
 class TestGetCatalog:
+    @pytest.fixture(scope="class")
+    def profiles_config_update(self, dbt_profile_target):
+        dbt_profile_target["retry_all"] = True
```
tauhid621: Would this work? 
Over the codebase we do something like:
```
        outputs = {"default": dbt_profile_target}
        outputs["default"]["retry_all"] = True
        return outputs
```

tauhid621 (commented):
Not sure about the way dbt_profile_target is being set. Perhaps we follow what other tests are doing?

colin-k-rogers:
Thanks for the review! Addressed in the latest push:

**`retries` value** — restored to 6; the value of 3 was an unintentional artifact of resolving a rebase conflict, not a deliberate change.

**`profiles_config_update` pattern** — switched all 9 new fixtures from the mutate-and-return-`{}` style to the standard codebase pattern:
```python
outputs = {"default": dbt_profile_target}
outputs["default"]["retry_all"] = True
return outputs
```
Yes @tauhid621, that's exactly the pattern we've now adopted — it matches what's used elsewhere in the test suite.

**Why `retry_all` isn't in the shared `dbt_profile_target`** — the Copilot suggestion to set it globally was actually our first attempt, but it broke `TestUniqueKeyRedshift::test__bad_unique_key` and friends. Those tests intentionally trigger a `DbtDatabaseError` and assert on its message text; with `retry_all=True` the retry logic closes/reopens the connection (rolling back the transaction and dropping the `__dbt_tmp` staging table), so the second attempt fails with a different error and the assertion on `"thisisnotacolumn"` never matches. Keeping `retry_all` scoped to the catalog/persist-docs classes avoids touching any test that deliberately exercises error paths.

tauhid621 (approved):

PR #1936, opened 2026-05-18
Bug: Persist docs 1.12 issues

### LINKED ISSUE #1937: [Bug] Persist Docs warning bugs
### Is this a new bug?

- [x] dbt-adapters
- [ ] dbt-tests-adapter
- [x] dbt-athena
- [ ] dbt-athena-community
- [x] dbt-bigquery
- [x] dbt-postgres
- [x] dbt-redshift
- [x] dbt-snowflake
- [x] dbt-spark

### Current Behavior

The `validate_doc_columns` macro added in [#1684](https://github.com/dbt-labs/dbt-adapters/pull/1684) has two issues raised in post-merge review:

  1. **Quoting config is ignored when comparing column names.** The macro unconditionally lowercases both sides:
 
     ```jinja
     {% set existing_lower = existing_column_names | map("lower") | list %}
     {% if col_name | lower not in existing_lower %}
     ```

When a column is documented with `quote: true` in `schema.yml`, identifiers are case-sensitive — "MyCol" is a distinct column from mycol. The current logic produces:
  - False matches: a documented MyCol (quoted) silently matches a physical mycol, so the comment lands on the wrong column or no warning fires when it should.
  - False misses: a correctly-quoted "FooBar" documented as FooBar (quoted) is reported as "missing" because the unquoted DB return-value differs in case.

  2. Redundant `get_columns_in_relation` call on Snowflake. `default__persist_docs` now pre-fetches the column list:
 
```jinja
  {% set existing_columns = adapter.get_columns_in_relation(relation) | map(attribute="name") | list %}
  {% set filtered_columns = validate_doc_columns(relation, model.columns, existing_columns) %}
```
 `snowflake__alter_column_comment` also calls `get_columns_in_relation` internally — so every Snowflake model with `persist_docs: { columns: true }` now pays for two describe table metadata calls instead of one. The original bug (#1690) only affected Postgres/Redshift (where alter_column_comment would emit empty SQL when no documented column existed); Snowflake never had the empty-query problem because it iterates over existing_columns directly.

### Expected Behavior

1. validate_doc_columns honors the per-column quote flag on column_dict[col_name]:
    - When quote: true → exact, case-sensitive comparison against existing_column_names.
    - When quote is unset/false → case-insensitive comparison (current behavior).
2. The Snowflake path makes exactly one get_columns_in_relation call per model and still emits the missing-column warning. The warning must continue to fire on every adapter — Postgres, Redshift, Snowflake, BigQuery, Spark, Athena.

### Steps To Reproduce

Issue 1 (quoting):
1. Create a model `quoted_case.sql`:
  ```sql
  {{ config(materialized='table') }}
  select 1 as mycol
```
2. Document it with a case-sensitive quoted name in schema.yml:
```yaml
  version: 2
  models:
    - name: quoted_case
      columns:
        - name: MyCol
          description: "case-sensitive name"
          quote: true
```
3. Run with `+persist_docs: { columns: true }`.
4. Expected: Warning that MyCol is not present in the database (mycol and MyCol are different identifiers under quoting). Current: No warning — validate_doc_columns lowercases both sides and falsely matches.

Issue 2 (redundant metadata call on Snowflake):
  1. Materialize any Snowflake table with `persist_docs: { columns: true }`.
  2. Inspect the query history (or trace adapter.get_columns_in_relation).
  3. Expected: One describe table call (from `snowflake__alter_column_comment`).
  Actual: Two describe table calls (`default__persist_docs` + `snowflake__alter_column_comment`).

### Relevant log output

```shell

```

### Environment

```markdown
- OS: macOS 15.x (Darwin 25.2.0)
- Python: 3.12
- dbt-core: main
- dbt-adapters: main ( affects upcoming 1.24 )
- warehouse adapters: main
```

### Additional Context

_No response_

sriramr98:
resolves #1937 

### Problem

[#1684](https://github.com/dbt-labs/dbt-adapters/pull/1684) added `validate_doc_columns` to warn on (and skip) columns documented in `schema.yml` that don't exist in the database. Post-merge review surfaced two issues:

  1. **Quoting is ignored.** The macro unconditionally lowercases both sides before comparing, so `MyCol` documented with `quote: true` silently matches a physical `mycol` (and vice versa) — producing wrong comments or missing warnings.
  2. **Redundant metadata call on Snowflake.** `default__persist_docs` pre-fetches columns via `get_columns_in_relation`, but `snowflake__alter_column_comment` already fetches them internally. Every Snowflake model with `persist_docs: { columns: true }` now pays for two `describe table` calls instead of one. The original empty-query bug never affected Snowflake.

### Solution

  1. `validate_doc_columns` now reads `column_dict[col_name]['quote']` and uses **case-sensitive** comparison when truthy; otherwise it falls back to the existing case-insensitive comparison.
  2. Added a `snowflake__persist_docs` override that skips the pre-fetch and delegates to `snowflake__alter_column_comment`. The `validate_doc_columns` call is moved **inside** `snowflake__alter_column_comment`, reusing the column list it already retrieves. Net: one `describe table` call, missing-column warning preserved.
  3. Other adapters (Postgres, Redshift, BigQuery, Spark, Athena) continue to use `default__persist_docs` / their existing `__persist_docs` overrides and pick up the quote-aware comparison automatically — the warning continues to fire on every adapter.

New shared tests in `dbt-tests-adapter` cover both fixes and are wired into every adapter's functional suite:
  - `BasePersistDocsQuotedColumnCaseSensitive` — quoted column with case mismatch must warn.
  - `BasePersistDocsAllColumnsMissing` — every documented column invalid must not crash and must warn for all.

TODO

### Checklist


dbt-adapters/src/dbt/include/global_project/macros/adapters/persist_docs.sql
```diff
@@ -26,18 +26,23 @@
 {% macro validate_doc_columns(relation, column_dict, existing_column_names) %}
   {% set existing_lower = existing_column_names | map("lower") | list %}
```
xuliangs, line 27: If the `existing_column_names` is returned from adapter.get_columns_in_relation which is directly returned from `describe table` without post-processing, then we may not want to apply lowercase here?

in this case, the existing_column_names are the resolved names in the db, so later when is_quoted is false, we should uppser case the `col_name | upper existing_column_names` before doing the lookup

sriramr98: Hey @xuliangs . Later, when we check for quoted columns, we check against `existing_column_names` which is unmodified. We only check against the `existing_lower` set when quote is false. Also `col_name | upper existing_column_names` wouldn't work since the actual quoted column can have any spelling with camel casing etc.

For non quoted columns, they are case insensitive by definition, hence comparing against names with lower case is the way to go.

xuliangs (approved):

dbt-adapters/src/dbt/include/global_project/macros/adapters/persist_docs.sql
xuliangs: The general principal is that we want to keep the column names as they're since they represent the resolved identifier in the db. Lowercasing the `existing_column_names` will make the later lookup lossy

dbt-athena/hatch.toml
```diff
@@ -56,12 +56,12 @@ check-sdist = [
 
 [envs.ci]
 pre-install-commands = [
+    "pip install dbt-core@git+https://github.com/dbt-labs/dbt-core.git#subdirectory=core",
```
colin-k-rogers, line 59: why are we doing this?

colin-k-rogers (approved):

dbt-adapters/src/dbt/include/global_project/macros/adapters/persist_docs.sql
```diff
@@ -26,18 +26,23 @@
...
   {% set missing = [] %}
+  {% set filtered = {} %}
   {% for col_name in column_dict %}
-    {% if col_name | lower not in existing_lower %}
+    {% set is_quoted = column_dict[col_name]['quote'] %}
+    {% if is_quoted %}
+      {% set present = col_name in existing_column_names %}
```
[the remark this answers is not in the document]
sriramr98: The problem is valid on special cases but the solution suggested is incorrect. This has already been discussed with @xuliangs

dbt-athena/hatch.toml
sriramr98: This is to circumvent a very niche edge case in CI caused by limiting 1.11.x and 1.10.x (latest releases) for dbt-adapters < 1.24. I had to do this to get the `dbt-adapter` changes reflected in the CI.

PR #1939, opened 2026-05-18
fix(snowflake): restore temporary __dbt_tmp for Python incrementals (#1928)

### LINKED ISSUE #1928: [Regression] Python incremental models: tmp_relation_type='table' rejected + __dbt_tmp created as permanent table (since 1.10.x)
### Is this a regression?

- [x] I believe this is a regression in functionality
- [x] I have searched the existing issues, and I could not find an existing issue for this regression

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [x] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

Two problems when using Python incremental models with `tmp_relation_type: table` (at project, folder, or model level):

### 1. Validation rejects the only valid value

```
Compilation Error in model my_model (models/my_model.py)
  Python models currently only support 'table' for tmp_relation_type but table was specified.

  > in macro dbt_snowflake_get_tmp_relation_type (macros/materializations/incremental.sql)
```

The check on line 41 is:

```jinja
{% if language == "python" and tmp_relation_type is not none %}
```

It should be:

```jinja
{% if language == "python" and tmp_relation_type is not none and tmp_relation_type != "table" %}
```

### 2. `__dbt_tmp` is created as a permanent table (not temporary)

Even when the validation is bypassed (by removing the config), the `py_write_table` macro in `create.sql` hardcodes `table_type=''`, which creates a permanent table. The `temporary` flag passed by `create_table_as(True, tmp_relation, ...)` is never propagated to the Python stored proc's `save_as_table()` call.

## Is this a regression?

Yes. This worked correctly in dbt-snowflake 1.9.x and broke in 1.10.x.

**Previous issue:** Originally reported in the archived repo as [dbt-labs/dbt-snowflake#823](https://github.com/dbt-labs/dbt-snowflake/issues/823), closed "Not planned" on Jan 6, 2025. The bug persists in the current monorepo across all versions from 1.10.0 through 1.11.5.
---

## What broke in 1.10.x

The rewrite in 1.10.x:

- Removed the `table_type` parameter from `py_write_table`
- Removed the `temporary` pass-through from `snowflake__create_table_as` for the Python path
- Hardcoded `table_type=''` (or only checks `is_transient`)

This produces `CREATE OR REPLACE TABLE ....__dbt_tmp` (permanent), which:

- Leaves orphaned tables if a run fails mid-way
- Violates the principle of least privilege (temporary tables are session-scoped and auto-cleaned)

---

## Workaround

Override three macros in the project `macros/` folder to restore the 1.9.x behavior:

1. `dbt_snowflake_get_tmp_relation_type` — fix validation to allow `table` for Python
2. `snowflake__create_table_as` — pass `temporary` to `py_write_table`
3. `py_write_table` — accept `temporary` and set `table_type='temporary'`

---

## Suggested Fix

Restore the `temporary`/`table_type` pass-through to `py_write_table` that existed in 1.9.x, and fix the validation check from `is not none` to `!= "table"`.

## Expected Behavior

Python incremental models should:

1. Accept `tmp_relation_type='table'` without error (same as SQL models)
2. Create `__dbt_tmp` as a session-scoped `TEMPORARY` table (same as SQL models)

---

## What worked in 1.9.x

In dbt-snowflake 1.9.2, `py_write_table` correctly received `table_type` from the `temporary` flag:

```jinja
{# create_table_as (1.9.x) — passed table_type to Python path #}
{{ py_write_table(compiled_code=compiled_code, target_relation=relation,
   table_type=relation.get_ddl_prefix_for_create(config.model.config, temporary)) }}

{# py_write_table (1.9.x) — accepted and used it #}
{% macro py_write_table(compiled_code, target_relation, temporary=False, table_type=none) %}
{% if table_type is not none %}
    {#- Just use the table_type as-is -#}
{% elif temporary -%}
    {%- set table_type = "temporary" -%}
{% else %}
    {%- set table_type = "" -%}
{%- endif %}
    df.write.mode("overwrite").save_as_table('...', table_type='{{table_type}}')
```

This produced `CREATE OR REPLACE TEMPORARY TABLE ....__dbt_tmp`.

---

### Steps To Reproduce

```yaml
# dbt_project.yml
models:
  my_project:
    +tmp_relation_type: table
```

```python
# models/my_model.py
def model(dbt, session):
    dbt.config(materialized="incremental", unique_key="id")
    df = session.table("my_source")
    return df
```

```bash
dbt run --select my_model
# → Compilation Error: Python models currently only support 'table' for tmp_relation_type but table was specified.
```

### Relevant log output

```shell
Python models currently only support 'table' for tmp_relation_type but table was specified.
  
  > in macro dbt_snowflake_get_tmp_relation_type (macros/materializations/incremental.sql)
  > called by macro materialization_incremental_snowflake (macros/materializations/incremental.sql)
  > called by model testpython (models/marts/testpython.py)
```

### Environment

```markdown
| Component     | Version                    |
|---------------|----------------------------|
| dbt-core      | 1.10.x / 1.11.x            |
| dbt-snowflake | 1.10.0 through 1.11.5      |
| Snowflake     | Any edition                |
```

### Additional Context

_No response_

tauhid621:
## Summary

Fixes [#1928](https://github.com/dbt-labs/dbt-adapters/issues/1928). Three related fixes restoring 1.9.x table-type semantics for Python models on Snowflake.

PR #932 (ADAP-1170, managed catalog migration) dropped the `temporary` / `table_type` pass-through that `py_write_table` had in 1.9.x. Since 1.10, this has caused two regressions plus an inconsistent validation guard.

### Bug 1 — `tmp_relation_type='table'` rejected for Python models

`incremental.sql:41` checked `tmp_relation_type is not none`, raising a compiler error on any value — including `'table'`, despite the error message stating only `'table'` is supported. Fixed by adding `and tmp_relation_type != "table"` to the guard.

### Bug 2 — `__dbt_tmp` created as a permanent table

`snowflake__create_table_as` received `temporary` but stopped passing it to `py_write_table`, and `py_write_table` was rewritten to look only at `catalog_relation.is_transient`. Result: `table_type` rendered as an empty string, and Snowpark's `save_as_table(table_type='')` produces a **permanent** table — orphaned on run failure. Fixed by propagating `temporary` and prioritizing it over `is_transient`, matching the precedence in `SnowflakeRelation.get_ddl_prefix_for_create` (`temporary` > `transient` > permanent).

### Side effect — Python `materialized='table'` ignored `transient` config

Same `is_transient`-only logic meant Python tables were always permanent regardless of user config. After the fix, Python tables on default Snowflake config become TRANSIENT (matching SQL behavior). Users explicitly setting `{{ config(transient=false) }}` still get permanent tables.

## Behavior matrix

| Scenario | 1.9.x | 1.10.x (current) | After this PR |
|---|---|---|---|
| Python incremental `__dbt_tmp` | TEMPORARY | PERMANENT | TEMPORARY |
| Python `materialized='table'`, default config | TRANSIENT | PERMANENT | TRANSIENT |
| Python `materialized='table'`, `transient=false` | PERMANENT | PERMANENT | PERMANENT |
| `tmp_relation_type='table'` on Python incremental | accepted | compile error | accepted |

## Test plan

- [x] Set `{{ config(materialized='incremental', tmp_relation_type='table') }}` on a Python model, assert no compile error.
- [x] Set `{{ config(materialized='incremental', tmp_relation_type='view') }}` on a Python model, assert the existing error still fires.
- [x] Run a Python `materialized='table'` model with default config and assert `kind = TRANSIENT`.
- [x] Run a Python `materialized='table'` model with `transient=false` and assert `kind = TABLE`.

## Backport

This needs to be cherry-picked to `1.10.patch` once merged (per the on-call discussion). 1.11 is still pre-GA so the main merge ships there directly.


ash2shukla (approved):
LGTM

PR #1940, opened 2026-05-18
excludes non existent schemas from catalog relations

### LINKED ISSUE #1005: [Bug] Adding a INFORMATION_SCHEMA source for BigQuery makes all catalog calls to use that source definition

### Current Behavior

If I create a source file
```yaml
version: 2

sources:
  - name: information_schema
    schema: region-europe-west4.INFORMATION_SCHEMA
    tables:
      - name: column_field_paths
        identifier: COLUMN_FIELD_PATHS
```

And run `dbt docs generate` I get the error

```
Encountered an error while generating catalog: Database Error
  Invalid project ID 'my-project-id.region-europe-west4'. Project IDs must contain 6-63 lowercase letters, digits, or dashes. Some project IDs also include domain name separated by a colon. IDs must start with a letter and may not end with a dash.
```

### Expected Behavior

It shouldn't use that defined source, and there should be a way of referencing information schema tables with `source()` macro that doesn't collide with the catalog generation.

### Steps To Reproduce

Add a source file 
```yaml
version: 2

sources:
  - name: information_schema
    schema: region-europe-west4.INFORMATION_SCHEMA
    tables:
      - name: column_field_paths
        identifier: COLUMN_FIELD_PATHS
```

run `dbt docs generate`

### Relevant log output

```shell

```

### Environment

```markdown
- OS: MacOS, Ubuntu 24
- Python: ">=3.10,<3.11"
- dbt-bigquery = ">=1.8.0,<1.8.1"
- dbt-core = ">=1.8.0,<1.8.1"
```

bigquery

### Additional Context

_No response_

ash2shukla:
resolves #1005 

### Problem
Information schemas in source schemas fail `docs generate` as relations can not be built for these relations.

### Solution
Exclude non existent schemas for catalog relations

### Checklist


dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -1056,6 +1056,28 @@ def _get_catalog_schemas(self, relation_config: Iterable[RelationConfig]) -> Sch
                 )
         return result
 
+    def _get_catalog_relations_by_info_schema(self, relations):
```
aahel, line 1059: Could you add the type annotation to match the base class signature?


def _get_catalog_relations_by_info_schema(
    self, relations
) -> Dict[InformationSchema, List[BaseRelation]]:

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
```diff
@@ -1056,6 +1056,28 @@ def _get_catalog_schemas(self, relation_config: Iterable[RelationConfig]) -> Sch
...
+        schema_exists: Dict[str, Dict[str, bool]] = {}
+        result = {}
+
+        for info_schema, rels in candidates.items():
+            database = info_schema.database
+            schema = info_schema.schema
+            if database not in schema_exists:
+                schema_exists[database] = {}
+            if schema not in schema_exists[database]:
+                schema_exists[database][schema] = self.check_schema_exists(database, schema)
```
aahel, line 1070: _get_catalog_schemas (right above this method) uses self.list_schemas(database) once per database and checks set membership locally  a single API round trip per project. This new method calls check_schema_exists per unique (database, schema) pair, which is N round trips.

I see why check_schema_exists is preferred here (it handles underscore-prefixed datasets that don't appear in list_datasets).
But dosent this does become inconsistent right ?

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
aahel, line 1059: A one-line docstring like this could help:


"""Override the base implementation to drop relation groups whose schema
isn't a real BigQuery dataset. This prevents `dbt docs generate` from
failing when a source references INFORMATION_SCHEMA views.
"""

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
aahel:
```
class BigQueryAdapter(BaseAdapter):
    def __init__(self, ...):
        super().__init__(...)
        self._listed_datasets_cache: Dict[str, Set[str]] = {}

    def _dataset_exists(self, database: str, schema: str) -> bool:
        if database not in self._listed_datasets_cache:
            self._listed_datasets_cache[database] = set(self.list_schemas(database))
        if schema in self._listed_datasets_cache[database]:
            return True
        return self.check_schema_exists(database, schema)
```

self._dataset_exists can be called from both _get_catalog_schemas and _get_catalog_relations_by_info_schema

ash2shukla: It only checks for schemas used in the project, which would be at max in single digits. This should not cause a performance regression.

dbt-bigquery/src/dbt/adapters/bigquery/impl.py
ash2shukla: We need to get this in today so can we merge this as otherwise it will take 4h more to run bq integration tests ? I can do these in a followup PR.

aahel (approved):

PR #1795, opened 2026-03-23
Migrate get_partitions_metadata from Legacy SQL to Standard SQL

aahel:
resolves #1725
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#N/A

### Problem

BigQuery is deprecating Legacy SQL starting June 1, 2026. New GCP projects created after that date will not support legacy SQL at all. `get_partitions_metadata()` currently uses Legacy SQL with `$__PARTITIONS_SUMMARY__`, which will stop working.

### Solution

Add a behavior flag `bigquery_use_standard_sql_for_partitions` to phase the migration from Legacy SQL to Standard SQL (`INFORMATION_SCHEMA.PARTITIONS`).

**Phased rollout:**
- **Phase 1** (this PR): `default=False` — legacy SQL remains the default, users opt in via `flags.bigquery_use_standard_sql_for_partitions: true` in `dbt_project.yml`
- **Phase 2** (next release): flip `default=True` — standard SQL becomes the default
- **Phase 3** (before June 1, 2026): remove flag and legacy code entirely

Only `partition_id` is selected in the standard SQL path since it's the only column used by all internal callers and the only column fully compatible between the two sources.

**Trade-offs:**
- `$__PARTITIONS_SUMMARY__` was free; `INFORMATION_SCHEMA.PARTITIONS` is billed per query (flat rate, not per-byte)
- Column differences: the legacy meta-table returned `project_id`, `dataset_id`, `table_id`, `partition_id`, `creation_time`, `last_modified_time`. `INFORMATION_SCHEMA.PARTITIONS` has different columns — only `partition_id` is shared and fully compatible. Custom macros accessing other columns from `get_partitions_metadata()` results will need updating.

**Potential breaking change — IAM permissions:**
- Legacy `$__PARTITIONS_SUMMARY__` required: `bigquery.jobs.create` + `bigquery.tables.getData`
- `INFORMATION_SCHEMA.PARTITIONS` requires: `bigquery.tables.get` + `bigquery.tables.list`
- Users with custom IAM roles that grant `bigquery.tables.getData` but not `bigquery.tables.get`/`bigquery.tables.list` will need their permissions updated. This does not affect users with standard predefined roles (`roles/bigquery.dataViewer`, `roles/bigquery.dataEditor`, `roles/bigquery.admin`) as those include all required permissions.

**References:**
[...]

PR #1945, opened 2026-05-19
feat(bigquery): flip bigquery_use_standard_sql_for_partitions default to true

aahel:
## Summary

- Flips the `bigquery_use_standard_sql_for_partitions` behavior flag default from `False` to `True`
- This enables Standard SQL (`INFORMATION_SCHEMA.PARTITIONS`) by default for partition metadata queries, replacing Legacy SQL (`$__PARTITIONS_SUMMARY__`)
- BigQuery is deprecating Legacy SQL on **June 1, 2026**, so this change ensures users are migrated before the deadline

## Test plan

- [ ] Existing functional tests in `tests/functional/test_changing_partitions.py` cover both code paths
- [ ] Verify partition metadata queries work correctly with the Standard SQL path enabled by default

### DIFF
--- a/dbt-bigquery/.changes/unreleased/Under the Hood-20260519-085000.yaml
+++ b/dbt-bigquery/.changes/unreleased/Under the Hood-20260519-085000.yaml
@@ -1,6 +1,0 @@
-kind: Under the Hood
-body: Enable `bigquery_use_standard_sql_for_partitions` behavior flag by default, switching partition metadata queries from Legacy SQL to Standard SQL (INFORMATION_SCHEMA.PARTITIONS) ahead of BigQuery's Legacy SQL deprecation on June 1, 2026
-time: 2026-05-19T08:50:00.000000-07:00
-custom:
-  Author: aahel
-  Issue: "1725"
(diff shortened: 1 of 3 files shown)


dbt-bigquery/.changes/unreleased/Under the Hood-20260519-085000.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Under the Hood
```
tauhid621: Should this be breaking change?

tauhid621 (approved):
I think the changelog should be breaking change?

dbt-bigquery/.changes/unreleased/Under the Hood-20260519-085000.yaml
aahel: yeah true. nice catch i forgot.

https://github.com/dbt-labs/dbt-adapters/pull/1795

aahel: changed

xuliangs:
Is the `adapter.get_partitions_metadata` still in use?

aahel:
> Is the `adapter.get_partitions_metadata` still in use?

only bigquery exposes it. so bigquery users can use it in their tests/macros

Issue #1948, opened 2026-05-19
default__concat produces invalid SQL for single-field input when adapters override with CONCAT() function


sdebruyn:
## Summary

The current `default__concat` macro joins fields with ` || `, which works correctly for a single-field list (it simply returns the field). However, many adapters override `default__concat` to emit a SQL `CONCAT(...)` function call instead of the `||` operator. On engines like SQL Server / Microsoft Fabric T-SQL, `CONCAT()` requires at least two arguments — `CONCAT(x)` raises a syntax error.

Because the default macro propagates a single-field list to overrides, every override that uses the function form must add its own defensive branch (or break on single-field input).

## Repro

When `dbt_utils.generate_surrogate_key([single_field])` is called, the package generates `dbt.hash(dbt.concat([field]))`. On Fabric this becomes `HASHBYTES(... CONCAT(field) ...)`, which fails to compile.

## Affected adapters / known workarounds

- dbt-spark: overrides `default__concat` with `concat({{ fields|join(', ') }})`.
- dbt-fabric (community fork): currently ships a duplicated override of `dbt_utils.generate_surrogate_key` to work around this exact problem — see https://github.com/sdebruyn/dbt-fabric/blob/main/src/dbt/include/fabric/macros/dbt_package_support/dbt_utils/sql/generate_surrogate_key.sql

Any adapter that uses the `CONCAT(...)` function form needs the same fix.

## Proposal

Short-circuit `default__concat` on `length == 1` and return the field unchanged. Behavior is identical for multi-field lists; the change is purely defensive.

```jinja
{% macro default__concat(fields) -%}
    {%- if fields | length == 1 -%}
        {{ fields[0] }}
    {%- else -%}
        {{ fields | join(' || ') }}
    {%- endif -%}
{%- endmacro %}
```

A dispatched-down override would have to repeat the same guard in every adapter, so the fix belongs in the default.

I'll open a PR with the change plus a `BaseConcatSingleField` test in `dbt-tests-adapter` so every adapter inherits coverage.

PR #1949, opened 2026-05-19
fix: default__concat returns single element directly for 1-field lists

sdebruyn:
## Problem

`default__concat` joins fields with ` || `, which works correctly for a single-field list (it just returns the field). However, several adapters override `default__concat` with a SQL `CONCAT(...)` **function call** rather than the `||` operator (for example, dbt-spark uses `concat({{ fields|join(', ') }})`).

On engines like SQL Server / Microsoft Fabric T-SQL, `CONCAT()` requires at least two arguments — `CONCAT(x)` is a syntax error. Because the default macro propagates a single-element list down to overrides, every adapter using the function form must add its own defensive branch (or break on single-field input).

Concrete repro: `dbt_utils.generate_surrogate_key([single_field])` compiles to `dbt.hash(dbt.concat([field]))`, which becomes `HASHBYTES(... CONCAT(field) ...)` on Fabric and fails to parse.

This is an actively painful issue in the wild: dbt-fabric (community fork) currently ships a duplicated override of `dbt_utils.generate_surrogate_key` purely to work around this — see https://github.com/sdebruyn/dbt-fabric/blob/main/src/dbt/include/fabric/macros/dbt_package_support/dbt_utils/sql/generate_surrogate_key.sql. Any adapter that uses the function form of `CONCAT()` needs the same workaround, so the fix belongs at the default macro level rather than being repeated per adapter.

Tracked in https://github.com/dbt-labs/dbt-adapters/issues/1948.

## Fix

Short-circuit `default__concat` on `length == 1` and return the field directly:

```jinja
{% macro default__concat(fields) -%}
    {%- if fields | length == 1 -%}
        {{ fields[0] }}
    {%- else -%}
        {{ fields | join(' || ') }}
    {%- endif -%}
{%- endmacro %}
```

## Impact / backwards compatibility

- Multi-field lists: behavior identical (`a || b` etc.).
- Single-field lists with the default `||` macro: behavior identical (output was already the field unchanged, just produced via `join` of one element).
- Single-field lists with overrides that re-dispatch up: those overrides receive a literal field expression instead of a list, so they no longer emit `CONCAT(x)` — fixing the syntax error.
- No new public API.

## Test

Adds `BaseConcatSingleField` to `dbt-tests-adapter` so every adapter inherits regression coverage for the single-field path. Wires `TestConcatSingleField` into `dbt-postgres`'s functional adapter test suite to validate the default behavior in CI.

## Test plan

- [x] `BaseConcatSingleField` added to `dbt-tests-adapter`
- [x] `TestConcatSingleField(BaseConcatSingleField)` added to dbt-postgres
- [ ] Adapter maintainers may want to inherit `BaseConcatSingleField` in their own functional suites to confirm the fix on their dialect (especially dbt-spark and any adapter overriding `default__concat` with a SQL function call)

## Changelog

- `dbt-adapters/.changes/unreleased/Fixes-20260519-124215.yaml`
- `dbt-tests-adapter/.changes/unreleased/Features-20260519-124215.yaml`

### DIFF
--- a/dbt-adapters/.changes/unreleased/Fixes-20260519-124215.yaml
+++ b/dbt-adapters/.changes/unreleased/Fixes-20260519-124215.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: Return the field unchanged from default__concat when called with a single-element
+  list, so adapters that override with a SQL CONCAT() function call do not receive
+  invalid single-argument input
+time: 2026-05-19T12:42:15.000000+00:00
+custom:
+  Author: sdebruyn
+  Issue: "1948"
--- a/dbt-adapters/src/dbt/include/global_project/macros/utils/concat.sql
+++ b/dbt-adapters/src/dbt/include/global_project/macros/utils/concat.sql
@@ -3,5 +3,9 @@
 {%- endmacro %}
 
 {% macro default__concat(fields) -%}
-    {{ fields|join(' || ') }}
+    {%- if fields | length == 1 -%}
+        {{ fields[0] }}
+    {%- else -%}
+        {{ fields | join(' || ') }}
+    {%- endif -%}
 {%- endmacro %}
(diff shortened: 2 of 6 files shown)


sdebruyn:
@cla-bot fixed

sdebruyn:
Human reviewed. This would simplify the dbt-fabric adapter.

PR #1957, opened 2026-05-19
[BigQuery] Fix: short-circuit getQueryResults polling on terminal job failure

### LINKED ISSUE #1956: [BigQuery] getQueryResults polling retries until job_execution_timeout, ignoring job_retries and terminal job state
## Summary

The BigQuery adapter's `getQueryResults` polling can retry for the entire `job_execution_timeout_seconds` window (often hours) when BigQuery returns a retryable-classified error, even when:

1. The underlying query job has already reached a terminal failed state, so retrying the poll cannot succeed.
2. The user has configured `job_retries` (attempt cap) — which is ignored on this code path.
3. The user has configured `job_retry_deadline_seconds` — which is also ignored on this code path.

The result: a query that BQ has already failed reports the failure to dbt only after the full execution timeout (e.g. 8h) elapses, instead of failing fast.

## Reproduction context

Observed a model failing after ~8h 20m. BigQuery returned a `400 internal error during execution` on `getQueryResults` ~57 minutes into the run, but dbt only surfaced the error 8+ hours later. The underlying SQL was not re-executed — dbt was looping on the `getQueryResults` poll for a job that was already dead. 

## Root cause

In [`dbt-bigquery/src/dbt/adapters/bigquery/connections.py`](https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-bigquery/src/dbt/adapters/bigquery/connections.py) `_query_and_results`:

```python
iterator = query_job.result(
    max_results=limit,
    timeout=polling_timeout,
    retry=DEFAULT_JOB_RETRY.with_timeout(timeout),  # timeout == job_execution_timeout_seconds
)
```

The `retry=` deadline is bound to `job_execution_timeout_seconds`. `python-bigquery`'s `DEFAULT_JOB_RETRY` predicate (`_job_should_retry`) classifies errors with reason `internalError` as retryable regardless of HTTP status, so each poll error restarts the retry loop, draining the full execution-timeout budget.

Three orthogonal issues:

1. **Retry budget is coupled to execution timeout.** `RetryFactory` already reads `job_retry_deadline_seconds` from credentials, but only applies it in `create_reopen_with_deadline()`. The `create_retry()` path (and the inline `DEFAULT_JOB_RETRY.with_timeout(timeout)` in `_query_and_results`) ignores it entirely.
2. **`job_retries` attempt cap is not enforced on the main query path.** `_DeferredException` enforces an attempt count, but only via `create_reopen_with_deadline()`. The `query_job.result(retry=...)` call uses raw `DEFAULT_JOB_RETRY` with no attempt ceiling.
3. **Retry predicate retries polls against terminal-failed jobs.** A `getQueryResults` retry cannot succeed once `query_job.state == \"DONE\"` with an `error_result` — the job is finished and failed. The current predicate has no awareness of job state.

## Proposed fix

Bundle the three concerns rather than just adding another timeout config:

1. **Wire existing `job_retry_deadline_seconds` into `_query_and_results`'s retry**, with a sane default (e.g. 600s) decoupled from `job_execution_timeout_seconds`. No new config name needed.
2. **Apply `job_retries` as an attempt cap** on the polling retry (reuse `_DeferredException` or equivalent predicate).
3. **Custom retry predicate** that returns `False` when the underlying job has reached a terminal failed state, even if `_job_should_retry` would otherwise retry.

aahel:
## Summary

Fixes #1956

When BigQuery returns an `internalError` on `getQueryResults` for a job that has already reached a terminal failed state, dbt was retrying the poll for the entire `job_execution_timeout_seconds` window (up to 8+ hours for some customers) instead of surfacing the error immediately. The underlying SQL was never re-executed — dbt was stuck in a loop pinging BQ for a result that was never coming.

**Root causes (all three from the issue):**
- Retry budget for `query_job.result()` was coupled to `job_execution_timeout_seconds` — the existing `job_retry_deadline_seconds` credential was ignored on this path
- `job_retries` attempt cap was not enforced on the main polling path
- Retry predicate had no awareness of whether the underlying job had actually reached a terminal failed state

## Changes

**`retry.py`**
- New `_TerminalJobAwarePredicate` — on each retry, calls `query_job.reload()` to get authoritative job state from `jobs.get`. If `state == "DONE"` and `error_result` is set, returns `False` immediately (stops polling). Also enforces `job_retries` as an attempt cap via `_DeferredException`. Non-retryable errors skip the reload entirely.
- New `RetryFactory.create_query_job_polling_retry(query_job)` — builds the retry with `_TerminalJobAwarePredicate` and a deadline from `job_retry_deadline_seconds` (default 10 min), **decoupled** from `job_execution_timeout_seconds`.

**`connections.py`**
- `_query_and_results`: use `self._retry.create_query_job_polling_retry(query_job)` instead of `DEFAULT_JOB_RETRY.with_timeout(timeout)`.

## Test plan

- `TestTerminalJobAwarePredicate` — 7 unit tests covering: terminal job short-circuit (the customer regression), transient error on running job, attempt cap, `job_retries=0`, non-retryable error skips reload, reload failure graceful fallback, terminal state wins over remaining retry budget
- `TestRetryFactoryPollingRetry` — 3 unit tests: deadline comes from `job_retry_deadline_seconds`, falls back to 600s default (not execution timeout), predicate is `_TerminalJobAwarePredicate`
- `TestBigQueryConnectionManager::test_query_and_results_uses_polling_retry` — integration test verifying `_query_and_results` wires the new retry through to `query_job.result()`

All 29 unit tests pass.


dbt-bigquery/src/dbt/adapters/bigquery/retry.py
```diff
@@ -92,6 +108,48 @@ def __call__(self, error: Exception) -> bool:
...
+        if self._retries == 0:
+            return False
+
+        if not _job_should_retry(error):
+            return False
+
+        # Reload from jobs.get to get authoritative job state.
+        try:
+            self._query_job.reload()
+            if self._query_job.state == "DONE" and self._query_job.error_result:
```
sriramr98: Worth being explicit that this issues an extra `jobs.get` per retry attempt. On a job hitting a string of transient errors before recovering, the API call count roughly doubles compared to the previous behavior.

Probably acceptable — authoritative job state is the whole point of the fix — but please add a one-line comment here saying we accept one `jobs.get` per retry as the cost, so the next person reading this doesn't try to optimize it away by caching or skipping reloads.

dbt-bigquery/src/dbt/adapters/bigquery/retry.py
```diff
@@ -62,6 +63,21 @@ def create_reopen_with_deadline(self, connection: Connection) -> Retry:
...
+        Build a retry for query_job.result() polling that:
+        - Uses job_retry_deadline_seconds (default 10 min) as the retry budget,
+          decoupled from job_execution_timeout_seconds so that a long-running job
+          does not silently burn the full execution timeout on getQueryResults errors.
+        - Enforces job_retries as an attempt cap.
+        - Short-circuits immediately when the underlying BQ job has reached a
+          terminal failed state (state == DONE + error_result), avoiding the
+          scenario where a permanently-dead job is polled for hours.
+        """
+        deadline = self._job_deadline or _DEFAULT_POLLING_RETRY_DEADLINE
```
sriramr98, line 78: Tracing the chain on `main` inside `_query_and_results` ([connections.py#L625-L647](https://github.com/dbt-labs/dbt-adapters/blob/ccfd06c70900545577795f7730596a7d274b1edc/dbt-bigquery/src/dbt/adapters/bigquery/connections.py#L625-L647)):

1. `timeout` is set from `job_execution_timeout_seconds` (or the model-level `job_timeout_ms` if present):
   ```python
   if "job_timeout_ms" in job_params:
       timeout = job_params["job_timeout_ms"] / 1000
   else:
       timeout = self._retry.create_job_execution_timeout()
   ```
2. That same `timeout` is then passed directly as the retry deadline a few lines below:
   ```python
   retry=DEFAULT_JOB_RETRY.with_timeout(timeout),
   ```
3. `create_job_execution_timeout` ([retry.py#L36-L39](https://github.com/dbt-labs/dbt-adapters/blob/ccfd06c70900545577795f7730596a7d274b1edc/dbt-bigquery/src/dbt/adapters/bigquery/retry.py#L36-L39)) returns `self._job_execution_timeout or _DAY` — i.e., the user's `job_execution_timeout_seconds`, falling back to 24h.

So on `main`, the polling retry's deadline = `job_execution_timeout_seconds` (or the model timeout, or 24h fallback). `job_retry_deadline_seconds` is **not** consulted on this path — it's only read by `create_reopen_with_deadline()` for connection-level ops like `cancel_job` / `create_dataset` ([retry.py#L59-L63](https://github.com/dbt-labs/dbt-adapters/blob/ccfd06c70900545577795f7730596a7d274b1edc/dbt-bigquery/src/dbt/adapters/bigquery/retry.py#L59-L63)).

This PR swaps that deadline to `self._job_deadline or _DEFAULT_POLLING_RETRY_DEADLINE` → `job_retry_deadline_seconds or 10 minutes`. For any user who hasn't set `job_retry_deadline_seconds` explicitly, the polling-retry window silently drops from `job_execution_timeout_seconds` (often hours) to 10 minutes.

That's a default-behavior change for every BQ user, not just the customer failure mode. The terminal-job short-circuit in `_TerminalJobAwarePredicate` is what actually fixes the customer regression — and it doesn't depend on this fallback at all, because `DONE + error_result` returns `False` on the very first error regardless of the deadline.

dbt-bigquery/src/dbt/adapters/bigquery/retry.py
```diff
@@ -92,6 +108,48 @@ def __call__(self, error: Exception) -> bool:
...
+            self._query_job.reload()
+            if self._query_job.state == "DONE" and self._query_job.error_result:
+                _logger.debug(
+                    f"Job {self._query_job.job_id} is in a terminal failed state; "
+                    "stopping getQueryResults polling."
+                )
+                return False
+        except Exception as reload_error:
+            _logger.debug(f"Failed to reload job state during retry predicate: {reload_error}")
+
```
sriramr98: Bare `except Exception` paired with a `debug`-level log will swallow real problems here — auth refresh failures, quota exhaustion on `jobs.get`, transport errors that genuinely break the client — and they'll only show up under verbose logging.

Two suggestions:
1. Narrow the catch to the google-api exceptions we actually expect from `reload()` (e.g. `GoogleAPICallError`, transport errors) so anything unexpected propagates.
2. At minimum bump this to `warning` — by the time we're here, the predicate is already in an error path; an additional reload failure is signal, not noise.

sriramr98 (changes requested):

dbt-bigquery/src/dbt/adapters/bigquery/retry.py
aahel: @colin-rogers-dbt do you think changing the default to 10 mins should be ok ?
we had decided to decouple it from `job_execution_timeout_seconds` [here](https://github.com/dbt-labs/dbt-adapters/issues/1956) in the proposed fix section

dbt-bigquery/src/dbt/adapters/bigquery/retry.py
aahel: fixed

dbt-bigquery/src/dbt/adapters/bigquery/retry.py
sriramr98: Ahh I didn't notice it's been discussed. I'm okay with it as long as we know the impact and it's a considered change.

dbt-bigquery/src/dbt/adapters/bigquery/retry.py
```diff
@@ -92,6 +109,62 @@ def __call__(self, error: Exception) -> bool:
...
+        try:
+            self._query_job.reload()
+            if self._query_job.state == "DONE" and self._query_job.error_result:
+                _logger.debug(
+                    f"Job {self._query_job.job_id} is in a terminal failed state; "
+                    "stopping getQueryResults polling."
+                )
+                return False
+        except (GoogleAPICallError, RequestException) as reload_error:
+            _logger.debug(
```
sriramr98: this should also be a warning or error for observability. Any reason why it's a `debug` log?

sriramr98 (approved):

colin-k-rogers (approved):

PR #1984, opened 2026-05-28
fix(athena): cancel in-flight Athena queries when a dbt invocation is cancelled

### LINKED ISSUE #406: [Bug] Cancelling dbt invocation does not cancel the query on Athena

aahel:
resolves #406
docs N/A

### Problem

When a dbt invocation is cancelled (e.g. via the dbt platform UI or Ctrl+C), the running Athena query is not cancelled. The query continues in the background, which causes concurrent request errors and "Destination table already exists" failures when the user retries.

The root cause is that `AthenaConnectionManager.cancel()` was a no-op (`pass`). Since model execution runs in a worker thread, `KeyboardInterrupt` from the main thread never reaches the cursor's `_poll` loop, so the existing `kill_on_interrupt` path inside the cursor is never triggered.

### Solution

- Subclasses `PyAthenaConnection` as `AthenaConnection`, which tracks all cursors it creates via a `weakref.WeakSet` and adds a `cancel()` method that calls `cursor.cancel()` on each tracked cursor.
- Adds `AthenaCursor.cancel()` which calls pyathena's `_cancel(query_id)` directly. Guards against a `None` query ID (returns early with a warning rather than passing `None` to the AWS API).
- Implements `AthenaConnectionManager.cancel()` to call `connection.handle.cancel()`.
- Stores the running query ID on `self.query_id` (instead of a local variable) so it is accessible during cancellation.
- Adds `CancelledQueryException` and raises it when a query reaches `STATE_CANCELLED`, preventing the outer retry loop from retrying after a cancellation.

This is substantially the same approach as the community PR #873, which was closed without merging due to CI failures. Those failures were caused by a bug in that PR's `cancel()` guard (it logged a warning but still called `_cancel(None)`); this implementation returns early instead.

### Checklist

### DIFF
--- a/dbt-athena/.changes/unreleased/Fixes-20260528-162219.yaml
+++ b/dbt-athena/.changes/unreleased/Fixes-20260528-162219.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Cancel in-flight Athena queries when a dbt invocation is cancelled
+time: 2026-05-28T16:22:19.786142+05:30
+custom:
+    Author: aahel
+    Issue: "406"
--- a/dbt-athena/.changes/unreleased/Fixes-20260528-162219.yaml
+++ b/dbt-athena/.changes/unreleased/Fixes-20260528-162219.yaml
@@ -2,5 +2,5 @@ kind: Fixes
 body: Cancel in-flight Athena queries when a dbt invocation is cancelled
 time: 2026-05-28T16:22:19.786142+05:30
 custom:
-    Author: aahel, Jrmyy
+    Author: aahel Jrmyy
     Issue: "406"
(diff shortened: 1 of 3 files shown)


ash2shukla (approved):
Can we add a functional test for this ?

aahel:
> Can we add a functional test for this ?

Didn’t add a functional cancel test in this PR because it’s hard to make reliable in CI: the query has to still be RUNNING when dbt handles the interrupt. Also we would need the specific query execution ID from logs to validate job status.

We have done manual testing for this pr

PR #2004, opened 2026-06-04
feat: support catalog_database on all catalog types

### LINKED ISSUE #2099: Add support for catalog_database in corev1
## Problem

In v2 `catalogs.yml`, there is no universal way to specify which physical database models should land in when using a catalog integration. This creates inconsistencies across adapters:

- **Snowflake CLD types** (glue/iceberg_rest/unity): `catalog_database` was already supported but only via the Snowflake-specific `catalog_linked_database` translation path.
- **Snowflake horizon (BUILT_IN)**: `catalog_database` was not extracted at all — models always landed in `target.database`.
- **Databricks unity**: database routing used `catalog_name` (the dbt label), requiring it to match the actual Unity Catalog name. No way to decouple the dbt label from the physical catalog name.
- **BigQuery and other adapters**: no catalog-level database override existed; models always fell back to `target.database` or a per-model `database` config.

## Solution

Make `catalog_database` the universal mechanism for specifying the physical database in v2 `catalogs.yml`, with the highest priority in `generate_database_name` across all adapters.

### Priority order

| Adapter | Priority |
|---|---|
| Snowflake / BigQuery / default | `catalog_database` > `database` model config > `target.database` |
| Databricks | `catalog_database` > `database` model config > `catalog_name` > `target.database` |

### Example

```yaml
# v2 catalogs.yml
catalogs:
  - name: my_catalog        # arbitrary dbt label
    type: horizon
    table_format: iceberg
    config:
      snowflake:
        catalog_database: PROD_DB   # actual Snowflake database
      databricks:
        catalog_database: main       # actual Unity Catalog name
        file_format: parquet
```

Models using `catalog_name: my_catalog` will land in `PROD_DB` on Snowflake and `main` on Databricks, regardless of `target.database` or model-level `database` config.

aahel:
resolves #2099 

### Problem

`catalog_database` was a Snowflake-only concept. In `catalogs.yml`, Snowflake catalog-linked-database (CLD) types used it to map a catalog to a specific physical database — implemented by remapping `catalog_database` → `catalog_linked_database` and reading it in `snowflake__generate_database_name`. No other adapter or catalog type could use this pattern, and there was no way to make a catalog's database name take precedence over a model's `database` config in a generic way.

### Solution

Promote `catalog_database` to a first-class field on the shared catalog base so **any** catalog type on **any** adapter can specify a database-name override, for both v1/v2 catalogs.

- Add `catalog_database: Optional[str]` to the `CatalogIntegrationConfig` Protocol, the `CatalogWriteIntegrationConfig` dataclass, and `CatalogIntegration.__init__` (stored as `self.catalog_database`).
- Extract `catalog_database` from the v2 platform block in `bridge_v2_catalog`. It is intentionally **kept in `props`** so adapter overrides (e.g. Snowflake's `_translate_v2_properties`) can still remap it to their platform-specific field name — preserving the existing Snowflake `catalog_linked_database` path.
- Add `@available get_catalog_database_override(config)` on `BaseAdapter`: looks up the catalog integration for a node and returns its `catalog_database`. Inherited by every adapter.
- Update `default__generate_database_name` to consult the override first: **`catalog_database` > `database` config > `target.database`**.
- Update `snowflake__generate_database_name` to also call `get_catalog_database_override` at highest priority, so v2 catalogs with `catalog_database` set take effect on Snowflake regardless of CLD type.

Snowflake's existing behavior is fully backward-compatible — both the generic path and the legacy `catalog_linked_database` path remain active.

The companion dbt-core change is in `dbt-labs/dbt-core` (adds the field to the v1 `CatalogWriteIntegrationConfig` resource): https://github.com/dbt-labs/dbt-core/pull/15156

### Checklist


dbt-snowflake/src/dbt/include/snowflake/macros/get_custom_name.sql
```diff
@@ -1,4 +1,10 @@
 {% macro snowflake__generate_database_name(custom_database_name=none, node=none) -%}
+    {%- if node is not none -%}
+        {%- set catalog_db = adapter.get_catalog_database_override(node) -%}
+        {%- if catalog_db -%}
+            {{ return(catalog_db) }}
+        {%- endif -%}
+    {%- endif -%}
     {%- if custom_database_name is none -%}
          {%- if node is not none and node|attr('database') -%}
             {%- set catalog_relation = adapter.build_catalog_relation(node) -%}
```
colin-k-rogers: Instead of introducing a new base adapter method (of which we have a lot already) could we just extend the existing pattern to reference the `catalog_db` property on the catalog_relation?

dbt-snowflake/src/dbt/include/snowflake/macros/get_custom_name.sql
```diff
@@ -1,4 +1,10 @@
 {% macro snowflake__generate_database_name(custom_database_name=none, node=none) -%}
+    {%- if node is not none -%}
+        {%- set catalog_db = adapter.get_catalog_database_override(node) -%}
+        {%- if catalog_db -%}
+            {{ return(catalog_db) }}
+        {%- endif -%}
+    {%- endif -%}
     {%- if custom_database_name is none -%}
```
colin-k-rogers: also it seems unintuitive to not have `custom_database_name` override this. In general dbt has the model level config override the higher level one.

aahel: tagging @jtcohen6 . Since we had decided catalog_db would have highest priority

dbt-snowflake/src/dbt/include/snowflake/macros/get_custom_name.sql
aahel: done

colin-k-rogers (approved):

PR #2015, opened 2026-06-09
perf(bigquery): set relation descriptions via DDL instead of the BigQuery API

### LINKED ISSUE #1568: Bigquery: update bigquery__alter_relation_comment to use alter table ddl instead of API call
Back in 2018ish when this logic was written BigQuery did not have a DDL syntax for updating the table description. They do now so we should take advantage of it as it will be more performant and more stable.
More performant because our current logic requires two calls: one to get the current table state the other to update it while using the ddl would only require a single call. More stable because it would fix the edge case where an update fails because a concurrent call updated the table in between the get/update calls.

See [bigquery__alter_relation_comment](https://github.com/dbt-labs/dbt-adapters/blob/b35814c52ef90054ab6fefffab85ed23d68f2bb5/dbt-bigquery/src/dbt/include/bigquery/macros/adapters.sql#L117) code.

Yaroslaww-1:
Resolves #1568.

### Problem

`bigquery__alter_relation_comment` called `adapter.update_table_description`, which does a get-then-update round trip (two API calls) and can fail when a concurrent update lands between the get and the update.

### Change

- Used single `ALTER {table|view|materialized view} ... SET OPTIONS(description=...)` DDL statement as [per suggestion](https://github.com/dbt-labs/dbt-adapters/issues/1568#issuecomment-3899166702).
- Deprecated `adapter.update_table_description` (via `@available.deprecated`).

### Testing

- Updated init tests assert the generated DDL and the no-op `bigquery_noop_alter_relation_comment` behavior-flag path.
- Rerun existing functional persist_docs + seed tests cover the description being applied end-to-end.

### DIFF
--- a/dbt-bigquery/src/dbt/include/bigquery/macros/adapters.sql
+++ b/dbt-bigquery/src/dbt/include/bigquery/macros/adapters.sql
@@ -136,7 +136,15 @@
       in DDL (e.g. OPTIONS(description=...)).
     -#}
   {%- else -%}
-    {% do adapter.update_table_description(relation.database, relation.schema, relation.identifier, relation_comment) %}
+    {%- if relation.type == 'materialized_view' -%}
+      {%- set relation_type = 'materialized view' -%}
+    {%- else -%}
+      {%- set relation_type = relation.type -%}
+    {%- endif -%}
+    {%- set alter_comment_sql -%}
+      alter {{ relation_type }} {{ relation.render() }} set options(description={{ relation_comment | tojson }});
+    {%- endset -%}
+    {% do run_query(alter_comment_sql) %}
   {%- endif -%}
 {% endmacro %}
 
--- a/dbt-bigquery/src/dbt/include/bigquery/macros/materializations/seed.sql
+++ b/dbt-bigquery/src/dbt/include/bigquery/macros/materializations/seed.sql
@@ -23,12 +23,10 @@
   {{ adapter.load_dataframe(model['database'], model['schema'], model['alias'],
   							agate_table, column_override, model['config']['delimiter']) }}
 
+  {# bigquery_table_options already renders OPTIONS(description=...) when
+     persist_relation_docs is enabled, so the description is set as part of this
+     single ALTER statement. No separate update_table_description call needed. #}
   {% call statement() %}
     alter table {{ this.render() }} set {{ bigquery_table_options(config, model) }}
   {% endcall %}
-
-  {% if config.persist_relation_docs() and 'description' in model %}
-
-  	{{ adapter.update_table_description(model['database'], model['schema'], model['alias'], model['description']) }}
-  {% endif %}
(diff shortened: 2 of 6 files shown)


Yaroslaww-1:
@cla-bot check

dbt-bigquery/src/dbt/include/bigquery/macros/materializations/seed.sql
```diff
@@ -23,12 +23,10 @@
   {{ adapter.load_dataframe(model['database'], model['schema'], model['alias'],
   							agate_table, column_override, model['config']['delimiter']) }}
 
+  {# bigquery_table_options already renders OPTIONS(description=...) when
+     persist_relation_docs is enabled, so the description is set as part of this
+     single ALTER statement. No separate update_table_description call needed. #}
   {% call statement() %}
     alter table {{ this.render() }} set {{ bigquery_table_options(config, model) }}
```
Yaroslaww-1, line 30: I deleted `and 'description' in model` guard below because it's not part of one alter table call and there is no explicit `model['description']` anymore

dbt-bigquery/tests/unit/test_noop_alter_relation_comment_behavior_flag.py
```diff
@@ -11,8 +10,14 @@ def _load_bigquery_adapters_macros(adapter):
...
     )
+    # alter_relation_comment now emits DDL via run_query; capture what it is handed
+    # and expose it on the module as `captured_sql` for assertions.
+    captured_sql = []
+    env.globals["run_query"] = captured_sql.append
     template = env.get_template("adapters.sql")
-    return template.make_module({"adapter": adapter})
+    module = template.make_module({"adapter": adapter})
+    module.captured_sql = captured_sql
+    return module
```
Yaroslaww-1, line 20: This is needed to capture successfull call `{% do run_query(alter_comment_sql) %}`

dbt-bigquery/src/dbt/include/bigquery/macros/adapters.sql
```diff
@@ -136,7 +136,15 @@
...
     -#}
   {%- else -%}
-    {% do adapter.update_table_description(relation.database, relation.schema, relation.identifier, relation_comment) %}
+    {%- if relation.type == 'materialized_view' -%}
+      {%- set relation_type = 'materialized view' -%}
+    {%- else -%}
+      {%- set relation_type = relation.type -%}
+    {%- endif -%}
+    {%- set alter_comment_sql -%}
+      alter {{ relation_type }} {{ relation.render() }} set options(description={{ relation_comment | tojson }});
```
Yaroslaww-1, line 145: I used `tojson` as an alternative to `'"""{}"""'.format(sql_escape(comment))`. I'm not actually sure I need it - maybe `relation_comment` is already escaped.

PR #2016, opened 2026-06-10
feat(athena): add catalogs.yml v2 support

### LINKED ISSUE #2038: dbt-athena: add catalogs.yml v2 support
### Summary

dbt-athena has no catalog-integration support today, so a model's table format (Hive vs Iceberg) can only be chosen via the `table_type` config. Any `catalogs.yml` entry currently fails because no catalog integrations are registered.

Add catalogs.yml v2 support:
- catalog integrations: `glue` → writable Iceberg on the AWS Glue Data Catalog, `info_schema` → default Hive
- the `CatalogsV2` capability and v2 bridge hooks
- catalog-driven table format selection in the table / incremental materializations, with an explicit `table_type` config still taking precedence (backward compatible)

so models can select their format via a catalog referenced with `catalog_name` instead of the `table_type` config.

### Adapter

dbt-athena

aahel:
resolves #2038

### Problem

dbt-athena has no catalog-integration support, so a model's table format (Hive vs Iceberg) can only be chosen via the `table_type` config — and any `catalogs.yml` entry fails because no catalog integrations are registered (`Catalog type is not supported`). This leaves dbt-athena out of the catalogs.yml v2 framework that the other adapters support.

### Solution

Add catalogs.yml v2 support to dbt-athena.

**Catalog integrations** (`adapters/athena/catalogs/`):
- `glue` → writable Iceberg backed by the AWS Glue Data Catalog
- `info_schema` → default Hive

Both are registered in `AthenaAdapter.__init__`, so `glue` and `info_schema` are reserved catalog names.

**Capability + v2 bridge** (`impl.py`):
- declares `Capability.CatalogsV2` (getattr-guarded so the current `dbt-adapters` lower bound still installs)
- `_v2_to_v1_type` (extension hook for catalog-type aliases) and `_v2_table_format` (maps the spec's `default` → Athena `hive`, `iceberg` → `iceberg`, and errors clearly on anything else)

**DDL wiring** via a new `resolve_table_type()` macro, used by the table, incremental, `on_schema_change`, and snapshot materializations. Model config always takes precedence; the catalog is the fallback:
- `table_type`: model `table_type` > catalog `table_format` > `hive`
- `format`: model `format` > catalog `file_format` > `parquet`
- location base: model `s3_data_dir` / `external_location` > catalog `external_volume` > target default

Backward compatible: with no catalog and no `table_type`, behavior is unchanged (the default `info_schema` catalog resolves to `hive`). Registering the integrations enables both the v2 (`config:` per-platform) and the older v1 (`write_integrations`) catalogs.yml shapes.

**Testing.** Unit tests cover the integrations, default-catalog registration (on a real adapter), and the v2 bridge. Functional tests were each verified end-to-end against Athena: table (Iceberg), precedence (explicit `table_type` wins), incremental + `on_schema_change`, `file_format` + `external_volume`, and a check-strategy snapshot.

**Known follow-up.** `cast_timestamp` is not yet catalog-aware: a catalog-only Iceberg model that calls it gets `timestamp` instead of `timestamp(6)`. Explicit `table_type='iceberg'` is unaffected, and it was observed to be non-fatal in testing (Athena coerces the precision). It's deferred because `cast_timestamp` also runs in contexts without a model config and needs a null-safe treatment.

### Checklist

- [x] I have run this code in development and it appears to resolve the stated issue (verified live against Athena)


iconara:
> A model with `catalog_name='glue'` now creates an Iceberg table

The Glue Data Catalog has historically been a Hive metastore. It's now a catalog-of-catalogs, spanning the old domain with Hive, Hudi, Delta, and Iceberg tables, as well as catalogs like S3 Tables, RMS, and federation – but the external name for the Iceberg-compatible multi-catalog is [SageMaker Lakehouse](https://aws.amazon.com/sagemaker/lakehouse/)

In current AWS nomenclature, "sagemaker" would be a more fitting name if the purpose is to convey an Iceberg-oriented catalog.

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -167,6 +187,31 @@ class AthenaAdapter(SQLAdapter):
...
+            **(
+                {_CATALOGS_V2_CAPABILITY: CapabilitySupport(support=Support.Full)}
+                if _CATALOGS_V2_CAPABILITY is not None
+                else {}
+            ),
+        }
+    )
+
+    # catalogs.yml v2 type -> the catalog_type expected by CATALOG_INTEGRATIONS
+    _V2_TO_V1_TYPE: ClassVar[Dict[str, str]] = {"glue": "glue"}
```
iconara: Why is this not using e.g. `DEFAULT_GLUE_CATALOG`?

What does it mean that the v1 type is "glue"? In the v2 implementation, "glue" means Iceberg and read/write, what does it mean in v1? (sorry if these are basic questions, I'm still trying to understand what catalog.yml v2 is).

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -167,6 +187,31 @@ class AthenaAdapter(SQLAdapter):
...
+
+    def __init__(self, config, mp_context: SpawnContext) -> None:
+        super().__init__(config, mp_context)
+        # Register the default catalogs so models can reference them by name and so
+        # models without a catalog fall back to standard Hive behavior.
+        self.add_catalog_integration(DEFAULT_INFO_SCHEMA_CATALOG)
+        self.add_catalog_integration(DEFAULT_GLUE_CATALOG)
+
+    def _v2_to_v1_type(self, catalog_type: str) -> str:
+        return self._V2_TO_V1_TYPE.get(catalog_type, catalog_type)
```
iconara, line 227: `_V2_TO_V1_TYPE` is `{"glue": "glue"}, and this just returns `catalog_type` by default -> is `_V2_TO_V1_TYPE` necessary? There is no case where this method will return anything except the input.

Unless I'm missing something, there would be no way to observe any difference if the implementation instead was this:

```python
def _v2_to_v1_type(self, catalog_type: str) -> str:
    return catalog_type
```

dbt-athena/src/dbt/include/athena/macros/materializations/models/helpers/resolve_table_type.sql
```diff
@@ -0,0 +1,18 @@
+{#
+  Resolve the effective `table_type` ('hive' or 'iceberg') for the current model.
```
iconara, line 2: I don't know if it matters, but Athena also supports Hudi and Delta tables (in read only mode). I assume they get mapped to "hive", but what are the implications of that? Hive tables are mutable (through `INSERT INTO`), Iceberg tables are mutable (through `INSERT INTO`, `MERGE INTO`, `UPDATE`, and `DELETE`), but Hudi and Delta are not.

dbt-athena/tests/unit/test_catalogs.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def test_build_catalog_relation_legacy_catalog_key(self, adapter):
+        relation = adapter.build_catalog_relation(_model({"catalog": "glue"}))
+        assert relation is not None
+        assert relation.table_format == constants.ICEBERG_TABLE_FORMAT
+
+    def test_build_catalog_relation_no_catalog_returns_none(self, adapter):
+        # With no catalog referenced, the base adapter returns None and the macro
+        # layer falls back to the default 'hive' table_type.
+        assert adapter.build_catalog_relation(_model({})) is None
```
iconara, line 109: Does these tests test anything but the test setup? `adapter` is a test fake that has been wired up to behave like this, and it looks to me that if the implementation of `AthenaAdapter` (or its superclasses) changed, these tests would continue to pass.

Why can't `adapter` be a real `AthenaAdapter`?

dbt-athena/tests/unit/test_catalogs.py
```diff
@@ -0,0 +1,111 @@
...
+
+class TestV2Bridge:
+    def setup_method(self):
+        self.adapter = object.__new__(AthenaAdapter)
+
+    def test_v2_to_v1_type_glue(self):
+        assert self.adapter._v2_to_v1_type("glue") == "glue"
+
+    def test_v2_to_v1_type_unknown_passthrough(self):
+        assert self.adapter._v2_to_v1_type("custom_type") == "custom_type"
```
iconara: This is testing an private method. If the method it meant to be used externally, it should be public, otherwise it should be tested through the public interface (which I think is what the next test below does, so these tests are not needed).

iconara (commented):
I'm not a dbt maintainer, I just work on Athena. Take my implementation comments as suggestions.

dbt-athena/src/dbt/adapters/athena/constants.py
```diff
@@ -39,3 +41,33 @@
...
+# Catalog integration (catalogs.yml v2) constants
+ICEBERG_TABLE_FORMAT = "iceberg"
+HIVE_TABLE_FORMAT = "hive"
+PARQUET_FILE_FORMAT = "parquet"
+GLUE_CATALOG_TYPE = "glue"
+INFO_SCHEMA_CATALOG_TYPE = "info_schema"
+
+# Default catalog registered for every model that does not reference a catalog.
+# Maps to Athena's standard Hive table_type.
+DEFAULT_INFO_SCHEMA_CATALOG = SimpleNamespace(
```
VersusFacit, line 54: just dropping by info schema is an abstraction I never really understand. Should this be hive if they are hive? I actually don't know, just asking 🍵

dbt-athena/src/dbt/include/athena/macros/materializations/models/helpers/resolve_table_type.sql
```diff
@@ -0,0 +1,18 @@
...
+
+  Precedence:
+    1. An explicit `table_type` model config always wins (backward compatible).
+    2. Otherwise, the `table_format` of the catalog referenced via `catalog_name`
```
VersusFacit, line 6: finally a use case for table format on catalog hooray!

Should this be hive or default. Based on info schema above default. But I lean hive

dbt-athena/src/dbt/adapters/athena/impl.py
aahel: intentionally kept this hook so that we can add other mappings later

dbt-athena/src/dbt/adapters/athena/impl.py
aahel: yeah. kept this for now so that we can extend it later if needed

dbt-athena/src/dbt/include/athena/macros/materializations/models/helpers/resolve_table_type.sql
aahel: dbt-athena only ever materializes hive or iceberg via CTAS; Hudi and Delta are read-only in Athena and aren't creatable by dbt, so they're out of scope here. resolve_table_type() only distinguishes iceberg vs hive and preserves the existing table_type config semantics. Anything that isn't iceberg falls back to hive, exactly as before

dbt-athena/tests/unit/test_catalogs.py
aahel: yeah. will refactor this

dbt-athena/src/dbt/adapters/athena/constants.py
aahel: hmm. i kept since it seemed to be the convention we are following for deafult catalog.

dbt-athena/src/dbt/include/athena/macros/materializations/models/helpers/resolve_table_type.sql
aahel: Fallback is hive because this value goes straight into Athena DDL (table_type='...'), where only hive/iceberg are valid.

aahel: have added _v2_table_format override.

```
_V2_TABLE_FORMAT = {"default": HIVE_TABLE_FORMAT, "iceberg": ICEBERG_TABLE_FORMAT}

def _v2_table_format(self, catalog: "CatalogV2") -> str:
    table_format = catalog.table_format.value
    try:
        return self._V2_TABLE_FORMAT[table_format]
    except KeyError:
        raise DbtRuntimeError(
            f"Catalog '{catalog.name}' has unsupported table_format '{table_format}'. "
            f"Athena supports: {', '.join(sorted(self._V2_TABLE_FORMAT))}."
        )

```

dbt-athena/tests/unit/test_catalogs.py
aahel: done

dbt-athena/tests/unit/test_catalogs.py
aahel: done

colin-k-rogers (approved):

PR #2018, opened 2026-06-10
feat(bigquery): support model-level location config to override query…

### LINKED ISSUE #2017: [Feature] BigQuery: support model-level location config to override query execution location

### Problem

In dbt-bigquery, the query execution location is fixed at the
profile level via the location credential. This means users
whose profile defaults to one region (e.g. EU) cannot write to
datasets in another region (e.g. US) without maintaining a
separate profile/target. BigQuery itself supports per-job
location overrides via the location parameter on `client.query()`,
but the adapter doesn't expose this.

### Proposed solution

Allow a location key in model config that overrides the
execution location for that model's queries:

```
models:
  my_project:
    my_dataset:
      +location: US
```

The adapter would pass this value as `location` to `client.query()`
for the duration of that model's execution, falling back to the
client default when not set.

### Motivation

Organisations operating in multiple regions (e.g.
EU-headquartered companies with US data) need cross-region
flexibility without duplicating profiles. [Global queries](https://docs.cloud.google.com/bigquery/docs/global-queries) (read from one region, write to another) are a supported
pattern — the adapter should expose the underlying capability.

### Implementation notes

- location is passed to `client.query()`, not `QueryJobConfig` — the
latter raises Property location is unknown.
- The override should be scoped to model execution only (set in
`pre_model_hook`, cleared in `post_model_hook`).

_No response_

### Who will this benefit?

Users that want fine grain control of location of query execution down to a model level. Very useful for transferring data from one region to another, without having to create an entirely new DBT profile + target and orchestrate two different dbt runs.

Yes

### Anything else?

_No response_

emilgq:
… execution location

Allows users to set `+location` in model config to override the BigQuery query execution location per model, independent of the profile default. The override is passed to `client.query()` via pre/post_model_hook.

resolves https://github.com/dbt-labs/dbt-adapters/issues/2017

### Problem
In dbt-bigquery, the query execution location is fixed at the
profile level via the location credential. This means users
whose profile defaults to one region (e.g. EU) cannot write to
datasets in another region (e.g. US) without maintaining a
separate profile/target. BigQuery itself supports per-job
location overrides via the location parameter on client.query(),
but the adapter doesn't expose this.

### Solution
Allow a location key in model config that overrides the
execution location for that model's queries:

```yaml
models:
  my_project:
    my_dataset:
      +location: US
```

The adapter would pass this value as location to client.query()
for the duration of that model's execution, falling back to the
client default when not set.

### Checklist


emilgq:
Hey @tauhid621  sorry for the direct ping. I see that you're an active maintainer and didn't know where else to turn.

This my first time contributing to this project and I'm not sure if I done everything expected in order to get a review. Should I just wait, fix something more or chase somebody down?

emilgq:
> Hey @tauhid621 sorry for the direct ping. I see that you're an active maintainer and didn't know where else to turn.
> 
> This my first time contributing to this project and I'm not sure if I done everything expected in order to get a review. Should I just wait, fix something more or chase somebody down?

Any comments? @tauhid621

PR #2026, opened 2026-06-17
fix(athena): support boolean partition columns on Iceberg in batch path

### LINKED ISSUE #2025: [Bug] dbt-athena: boolean partition column on an Iceberg table crashes the batch insert fallback ("Unsupported column type: boolean")

- [x] I believe this is a new bug in dbt-athena

### Current Behavior

On an **Iceberg** table partitioned by a **boolean** column, a run fails with:

```
ValueError: Unsupported column type: boolean
```

The error is raised by dbt (in Python, before any SQL is sent to Athena) on the partition-batching fallback path — the path dbt-athena takes when a single write opens more than `partitions_limit` (default 100) partitions. The table builds fine when that fallback isn't triggered, so a boolean-partitioned Iceberg model can work for a while and then fail the first time a single write crosses the limit (large backfill, a high-fan-out `bucket(N, ...)` write, etc.). Setting `force_batch=true` triggers the path deterministically.

### Expected Behavior

Boolean partition columns are **valid on Iceberg**, so dbt-athena should handle them on the batch path like the other supported partition types (`integer`, `string`, `date`, `timestamp`) instead of erroring.

This appears to be **Iceberg-specific**: boolean partition columns are **not** supported on **Hive** tables in Athena ([HIVE-6590](https://issues.apache.org/jira/browse/HIVE-6590) — Athena's engine rejects them at DDL time), which is likely why boolean was never handled here. The batch path is shared by both table types, so the fix should add boolean support for Iceberg while leaving Hive to raise a clear error.

### Steps To Reproduce

1. Create an Iceberg incremental model partitioned by a boolean column plus a high-cardinality bucket:

   ```sql
   {{ config(
       materialized='incremental',
       incremental_strategy='merge',
       table_type='iceberg',
       unique_key='my_key',
       partitioned_by=['is_current', 'bucket(id, 64)']
   ) }}
   select
       id,
       ...,
       cast(valid_to as date) = date '9999-12-31' as is_current  -- boolean
   from {{ ref('my_source') }}
   ```

2. Run it with enough data that a single write opens > 100 partitions (or set `force_batch=true` to force the path deterministically).
3. The run fails with `Unsupported column type: boolean`.

Workaround: type the partition column as `int` (0/1) instead of `boolean`.

### Relevant log output

```shell
... CREATE NON-PARTITIONED STAGING TABLE: ...
... ValueError: Unsupported column type: boolean
```

### Environment

```markdown
- OS: Linux
- Python: 3.10
- dbt-adapters: 1.24.2
- dbt-athena: 1.11.0b2 (also reproduced on 1.10.0.3; present on main as of 2026-06-15)
```

### Additional Context

Adjacent context: #1636 (merged 2026-02-16) fixed a *different* problem on this same batch path (query-size explosion from duplicated `IN (...)` clauses). This boolean issue is separate and currently unaddressed.

bxm156:
resolves #2025

Docs: N/A — no user-facing docs change.

### Problem

dbt-athena cannot write to an **Iceberg** table partitioned by a **boolean** column once the write is large enough to hit the partition-batching fallback — even though boolean partitions are valid on Iceberg. The run fails with:

```
ValueError: Unsupported column type: boolean
```

The error is raised by dbt (in Python, before any SQL reaches Athena) on the partition-batching fallback path — the path dbt-athena takes when a single write opens more than `partitions_limit` (default 100) partitions. `get_partition_batches` (and `process_bucket_column`) call `format_value_for_partition` for each distinct partition value, and that method has no `boolean` branch, so it hits the `else` and raises.

Because the crash only happens on the batch path, a boolean-partitioned Iceberg model can build fine for a while and then fail the first time a single write crosses the 100-partition limit (large backfill, a high-fan-out `bucket(N, ...)` write, `force_batch=true`, etc.).

Boolean partition columns are **valid on Iceberg** but **not on Hive** (Athena rejects them at DDL time, see [HIVE-6590](https://issues.apache.org/jira/browse/HIVE-6590)) — which is likely why boolean was never handled here. The batch path is shared by both table types, so the fix must add boolean support for Iceberg while leaving Hive to raise a clear error.

### Solution

Add a backward-compatible `table_type` parameter (default `"hive"`) to `format_value_for_partition`. For `table_type="iceberg"`, a boolean partition value is formatted as a bare `true`/`false` literal; for hive (the default), it still raises the existing clear error. `table_type` is threaded from `get_partition_batches` into both the non-bucketed call site and `process_bucket_column`. Existing two-argument callers are unaffected.

Tests are added at both levels: the Python method (`test_adapter.py` — iceberg formats; hive/unknown/default raise) and the macro end-to-end (`test_get_partition_batches.py` — iceberg formats, hive/default raise).

#### Alternatives considered

| # | Approach | Pros | Cons |
|---|----------|------|------|
| 1 | **Unconditional `boolean` branch in `format_value_for_partition`** (format boolean for every table type) | Smallest change (one branch); no signature change | Formats boolean for Hive too, which then fails *later* at Athena with a cryptic Hive DDL error — masks that Hive genuinely can't partition by boolean (HIVE-6590). Loses the clear early error. |
| 2 | **Scope the fix in Jinja only** — branch on `table_type` inside `get_partition_batches.sql`, format boolean inline for Iceberg, else fall through to `format_value_for_partition` | No Python signature change at all; Hive keeps the clear early raise | Splits partition-value formatting across two places — boolean logic lives in Jinja while `integer`/`string`/`date`/`timestamp` stay in Python. Harder to find/test; the bucketed call site (`process_bucket_column`) isn't covered by the same branch. |
| 3 | **`table_type` parameter on `format_value_for_partition` with a default** *(this PR)* | All partition-value formatting stays in one place (Python), directly unit-testable; Iceberg formats, Hive/default raises the clear error; the default keeps every existing 2-arg caller behaving exactly as before (boolean already raised) | Adds one optional parameter to an `@available` method (a minimal, backward-compatible interface change) |

**Why #3:** #1 hides a real Hive limitation behind a worse downstream error. #2 avoids touching Python but scatters the type-handling logic and leaves the bucketed path inconsistent. #3 keeps all type formatting cohesive in the single method that owns it, preserves the clear Hive error, and — thanks to the `"hive"` default — is non-breaking for any other caller, since boolean already raised under the old two-argument signature.

### Checklist

### DIFF
--- a/dbt-athena/.changes/unreleased/Fixes-20260616-222646.yaml
+++ b/dbt-athena/.changes/unreleased/Fixes-20260616-222646.yaml
@@ -1,0 +1,9 @@
+kind: Fixes
+body: Support boolean partition columns on Iceberg tables in the partition-batching
+    fallback path (force_batch or TOO_MANY_OPEN_PARTITIONS), which previously raised
+    "Unsupported column type: boolean". Hive tables still raise, since Hive does not
+    support boolean partition columns (HIVE-6590).
+time: 2026-06-16T22:26:46-07:00
+custom:
+    Author: bxm156
+    Issue: "2025"
--- a/dbt-athena/src/dbt/adapters/athena/impl.py
+++ b/dbt-athena/src/dbt/adapters/athena/impl.py
@@ -1614,8 +1614,16 @@ class AthenaAdapter(SQLAdapter):
         return int((hash_value & self.INTEGER_MAX_VALUE_32_BIT_SIGNED) % num_buckets)
 
     @available
-    def format_value_for_partition(self, value: Any, column_type: str) -> Tuple[str, str]:
-        """Formats a value based on its column type for inclusion in a SQL query."""
+    def format_value_for_partition(
+        self, value: Any, column_type: str, table_type: str = "hive"
+    ) -> Tuple[str, str]:
+        """Formats a value based on its column type for inclusion in a SQL query.
+
+        ``table_type`` defaults to ``"hive"`` so existing two-argument callers keep
+        their current behavior. Boolean partition columns are only valid on Iceberg
+        (Hive rejects them, see HIVE-6590), so booleans are formatted only when
+        ``table_type == "iceberg"`` and otherwise fall through to the error below.
+        """
         comp_func = "="  # Default comparison function
         if value is None:
             return "null", " is "
@@ -1629,6 +1637,10 @@ class AthenaAdapter(SQLAdapter):
             return f"DATE'{value}'", comp_func
         elif column_type == "timestamp":
             return f"TIMESTAMP'{value}'", comp_func
+        elif column_type == "boolean" and table_type == "iceberg":
+            # Athena/Presto accepts bare boolean literals (true/false, unquoted).
+            # The distinct partition query returns Python True/False, so lowercase.
+            return str(value).lower(), comp_func
         else:
             # Raise an error for unsupported column types
             raise ValueError(f"Unsupported column type: {column_type}")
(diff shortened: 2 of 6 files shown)


iconara:
This should probably go on top of #2016, which introduces more advanced catalog metadata – including table types.

bxm156:
> This should probably go on top of #2016, which introduces more advanced catalog metadata – including table types.

Ah good to know, will watch for that PR.

PR #1316, opened 2025-09-04
Adding support for Snowflake Workload Identity Federation

roryjbd:
resolves #1234
Issue raised for updating docs
https://github.com/dbt-labs/docs.getdbt.com/issues/7866

### Problem

Allows for authentication to Snowflake using Workload Identity Federation

### Solution

Allows the necessary parameters to be passed to the snowflake connector `connect` method. 
Updates to latest supported `snowflake-connector-python` package.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #2029, opened 2026-06-17
Add dynamic workload identity token providers for Snowflake WIF/OIDC

### LINKED ISSUE #2030: dbt-snowflake: mint a fresh Workload Identity (OIDC) token at connection open

### Describe the feature

`dbt-snowflake` supports Snowflake Workload Identity Federation (WIF) via #1316, but only with a **static** `token`:

```yaml
authenticator: workload_identity
workload_identity_provider: OIDC
token: "{{ env_var('SNOWFLAKE_TOKEN') }}"
```

The configured token must still be valid whenever dbt opens a Snowflake connection. For CI providers with short-lived OIDC tokens (e.g. GitHub Actions, ~5 min TTL) this is unsafe:

- a token minted at workflow start can expire before dbt opens a connection; and
- dbt can open or reopen Snowflake connections part-way through a long-running invocation, after the original token has expired.

The Snowflake connector does **not** mint OIDC tokens itself — for `workload_identity_provider: OIDC` it requires a caller-supplied `token` (unlike AWS/Azure/GCP, which it sources from the runtime). So token freshness is currently the caller's responsibility, and there is no way to mint a fresh token at connection-open time.

- **Pre-minting `SNOWFLAKE_TOKEN` before each dbt process.** Fixes stale-token-at-start, but not a connection opened/reopened mid-run.
- **`token_file_path` + a process that refreshes the file.** The connector re-reads `token_file_path` per connect, but this pushes token lifecycle management onto every user and every CI step.
- **An arbitrary `token_command` in `profiles.yml`.** Powerful but turns the profile into a code-execution surface; rejected on supply-chain/safety grounds.

### Proposed solution

A `workload_identity_token` provider config that mints a fresh token immediately before each connection open:

```yaml
authenticator: workload_identity
workload_identity_provider: OIDC
workload_identity_token:
  provider: github_actions
  audience: snowflakecomputing.com
```

First provider: `github_actions` (uses the standard `ACTIONS_ID_TOKEN_REQUEST_URL` / `ACTIONS_ID_TOKEN_REQUEST_TOKEN` env vars). Security-first: named adapter-owned providers (no shell), explicit opt-in, mutually exclusive with static `token`, host-pinned OIDC request with redirects disabled and a timeout, no token logging, fails closed, no cache in v1.

### Who will this benefit?

Teams running dbt-snowflake in CI with OIDC-based WIF (GitHub Actions today; the design generalizes to other providers), who currently hit intermittent expired-JWT failures.

Yes — implemented in #2029.

### Anything else?

Related: #1316 (static WIF support).

Daniel-Wiszowaty:
resolves #2030
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/# _(docs follow-up TBD)_

### Problem

`dbt-snowflake` supports Snowflake Workload Identity Federation (WIF) via #1316, but only with a **static** `token`:

```yaml
authenticator: workload_identity
workload_identity_provider: OIDC
token: "{{ env_var('SNOWFLAKE_TOKEN') }}"
```

The connector consumes this token only when a connection is **opened**. Snowflake then issues its own session/master tokens that keep an open connection alive (master token ~4h, auto-renewed), independent of the source OIDC token — so in-flight queries on already-open connections are unaffected by the source token expiring. The connector also does **not** re-mint OIDC tokens itself: for `workload_identity_provider: OIDC` it requires a caller-supplied `token` (unlike AWS/Azure/GCP, which it sources from the runtime).

So the failure mode is specific: with a short-lived CI OIDC token (e.g. GitHub Actions, ~5–15 min TTL), any connection dbt **opens or reopens after the token has expired** fails — late-starting threads, connection-pool misses, and reconnects on a long or highly-parallel run. A token minted once at workflow start cannot cover those later opens.

### When to use this (and when not to)

This is for teams that specifically want **keyless** WIF/OIDC — no long-lived secret at rest, federated trust to the CI identity. If you don't have that requirement, simpler options avoid the expiry window entirely:

- **Key-pair auth** sidesteps the problem with no new config behavior: the connector mints a fresh short-lived JWT from your private key on *every* connect and auto-renews the session. There is no caller-supplied token to expire. Prefer this unless a policy forbids storing a private key.
- **Static `token`** is sufficient for runs that comfortably finish within the OIDC TTL and don't open connections late.

dbt's default `reuse_connections: true` further narrows the window by pooling one open connection per thread, so dynamic minting earns its keep mainly on parallel/late-opening runs, reconnects, and `reuse_connections: false`.

### Solution

Add a `workload_identity_token` provider config. When set, dbt mints a **fresh** token immediately before each connection open:

```yaml
authenticator: workload_identity
workload_identity_provider: OIDC
workload_identity_token:
  provider: github_actions
  audience: snowflakecomputing.com   # defaults to snowflakecomputing.com
```

Minting happens inside `SnowflakeConnectionManager.open()`'s `connect()` closure, so every (re)connect and retry gets a newly minted token. Static `token` mode is unchanged and remains supported; the two modes are mutually exclusive.

The first provider is `github_actions`, which uses the standard GitHub Actions OIDC env vars (`ACTIONS_ID_TOKEN_REQUEST_URL`, `ACTIONS_ID_TOKEN_REQUEST_TOKEN`; workflow needs `permissions: id-token: write`).

**Design.** Providers live in a dedicated `workload_identity.py` module with a small dict dispatch (`provider name -> _mint_<provider>(config)`). Adding a provider (GCP, Azure, Vault, …) is one function + one dict entry and never touches connection code. I deliberately avoided a `Protocol`/registry/factory abstraction for what is currently one provider. (Note: env-var-injected CI tokens such as GitLab CI `id_tokens` and CircleCI `CIRCLE_OIDC_TOKEN` are *not* candidates for a provider — they're pre-minted and already covered by static `token` mode.)

**Security.**
- No arbitrary command execution — providers are adapter-owned Python, not shell hooks from `profiles.yml`.
- Explicit opt-in — nothing is minted unless `workload_identity_token` is configured.
- `token` and `workload_identity_token` are mutually exclusive (validation fails fast with a clear error), so there is no ambiguous precedence.
- The GitHub OIDC request is pinned to the `*.actions.githubusercontent.com` host family (rejecting `http`, foreign hosts, and `user@host` tricks), with redirects disabled and a 30s timeout — the bearer request token can itself mint identity tokens, so it must never be sent elsewhere.
- Token values and raw endpoint responses are never logged or placed in error messages (the OIDC success body *is* the token).
- Fails closed: a minting failure aborts the connection; there is no stale-token fallback and no token cache in v1, so each open mints fresh (also keeps it thread-safe — no shared mutable state).
- Snowflake-side WIF trust (issuer/subject/audience) remains the real authorization boundary.

### Operational security (deployment hardening)

Because the Snowflake-side WIF trust is the real authorization boundary, most of this feature's security lives in how the operator configures it. The setup guide (the test docstring) now spells this out:

- **Account-scoped audience.** Use an account-scoped `OIDC_AUDIENCE_LIST` (e.g. `https://<org>-<account>.snowflakecomputing.com`), not the bare `snowflakecomputing.com` default — Snowflake shares that value across all accounts, making a leaked token cross-account replayable. (The code keeps `snowflakecomputing.com` as a back-compat default; the docs steer to account-scoping.)
- **Tight SUBJECT.** Pin to a specific branch or, better, a protected GitHub Environment; never `:pull_request`, org-wide, or wildcard subjects; never trust forks.
- **Job-level `id-token: write`** (not workflow level); pin third-party actions by SHA; don't run on `pull_request_target` / `workflow_run` of untrusted code.
- **Least-privilege role** on the service user; revoke via `ALTER USER ... UNSET WORKLOAD_IDENTITY` (there is no GitHub-side revocation of an already-minted token).

A 5-way security review also caught and fixed a host-allowlist bypass: a URL such as `https://evil.com\.actions.githubusercontent.com/...` parsed as a trusted host by `urllib.parse` while `requests` dialed `evil.com`. The validator now rejects backslash / whitespace / control-character URLs and empty-label hosts before the request is made, so the bearer request token cannot be smuggled off-host. The `*.actions.githubusercontent.com` suffix allowlist is retained (the real request-URL host varies within that zone).

### Tests

- Unit: nested config parsing; default audience; the four validation rules (mutual exclusivity, requires `authenticator: workload_identity`, `github_actions` requires `workload_identity_provider: OIDC`, unknown provider); GitHub minting (env-var requirement, audience add/replace, returns `value`, missing value, non-JSON, non-200 does not leak the body, SSRF host/scheme/userinfo rejection); dispatch fails closed; and a regression proving two opens mint two different tokens.
- Functional: a dynamic-mode WIF/OIDC test alongside the existing static one, skipped when not running on a GitHub Actions OIDC runner.

`black`, `flake8`, and `mypy` pass; all unit tests pass locally.

### Checklist

- [x] This PR includes tests
- [ ] This PR has interface changes (a new `workload_identity_token` profile config) and may need Product/DX feedback

### DIFF
--- a/dbt-snowflake/.changes/unreleased/Features-20260617-191644.yaml
+++ b/dbt-snowflake/.changes/unreleased/Features-20260617-191644.yaml
@@ -1,0 +1,8 @@
+kind: Features
+body: Add dynamic workload identity token providers for Snowflake WIF/OIDC; the first
+  provider, `github_actions`, mints a fresh OIDC token immediately before each Snowflake
+  connection is opened
+time: 2026-06-17T19:16:44+02:00
+custom:
+    Author: Daniel-Wiszowaty
+    Issue: "2030"
--- a/dbt-snowflake/src/dbt/adapters/snowflake/connections.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/connections.py
@@ -59,6 +59,11 @@ from dbt.adapters.snowflake.record import SnowflakeRecordReplayHandle
 
 from dbt.adapters.snowflake.auth import private_key_from_file, private_key_from_string
 from dbt.adapters.snowflake.query_headers import SnowflakeMacroQueryStringSetter
+from dbt.adapters.snowflake.workload_identity import (
+    mint_workload_identity_token,
+    SUPPORTED_PROVIDERS as WORKLOAD_IDENTITY_TOKEN_PROVIDERS,
+    WorkloadIdentityTokenConfig,
+)
 
 if TYPE_CHECKING:
     import agate
@@ -132,6 +137,9 @@ class SnowflakeCredentials(Credentials):
     s3_stage_vpce_dns_name: Optional[str] = None
     workload_identity_provider: Optional[str] = None
     workload_identity_entra_resource: Optional[str] = None
+    # When set, dbt mints a fresh workload identity token before each connection
+    # open instead of using a static `token`. Mutually exclusive with `token`.
+    workload_identity_token: Optional[WorkloadIdentityTokenConfig] = None
     # Setting this to 0.0 will disable platform detection which adds query latency
     # this should only be set to a non-zero value if you are using WIF authentication
     platform_detection_timeout_seconds: Optional[float] = (
@@ -175,6 +183,42 @@ class SnowflakeCredentials(Credentials):
         if self.client_session_keep_alive is False and self.reuse_connections is None:
             self.reuse_connections = True
 
+        self._validate_workload_identity_token()
+
+    def _validate_workload_identity_token(self) -> None:
+        if self.workload_identity_token is None:
+            return
+
+        if self.token:
+            raise DbtConfigError(
+                "Invalid Snowflake profile: `token` and `workload_identity_token` cannot both "
+                "be configured. Use `token` for static token mode, or `workload_identity_token` "
+                "for dynamic token mode."
+            )
+
+        if not self.authenticator or self.authenticator.lower() != "workload_identity":
+            raise DbtConfigError(
+                "Invalid Snowflake profile: `workload_identity_token` requires "
+                "`authenticator: workload_identity`."
+            )
+
+        provider = self.workload_identity_token.provider
+        if provider not in WORKLOAD_IDENTITY_TOKEN_PROVIDERS:
+            raise DbtConfigError(
+                f"Invalid Snowflake profile: unknown workload identity token provider "
+                f"`{provider}`. Supported providers: "
+                f"{', '.join(sorted(WORKLOAD_IDENTITY_TOKEN_PROVIDERS))}."
+            )
+
+        if provider == "github_actions" and (
+            not self.workload_identity_provider
+            or self.workload_identity_provider.upper() != "OIDC"
+        ):
+            raise DbtConfigError(
+                "Invalid Snowflake profile: `workload_identity_token.provider: github_actions` "
+                "requires `workload_identity_provider: OIDC`."
+            )
+
     @property
     def type(self):
         return "snowflake"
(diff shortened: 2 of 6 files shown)


Daniel-Wiszowaty:
@colin-rogers-dbt PTAL

PR #2034, opened 2026-06-18
Fix Snowflake WIF (OIDC) auth tripping token/user profile validation

tauhid621:
Workload Identity Federation was missing from the `authenticator` allowlist in `SnowflakeCredentials.__post_init__`, so OIDC profiles (which pass a `token` and no `user`) hit a spurious token warning and the hard `'user' is a required property` error.

Adds `workload_identity` to the allowlist and a unit test guarding the regression.

### DIFF
--- a/dbt-snowflake/src/dbt/adapters/snowflake/connections.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/connections.py
@@ -147,19 +147,25 @@ class SnowflakeCredentials(Credentials):
                 )
             )
 
-        if self.authenticator not in ["oauth", "jwt", "programmatic_access_token"]:
+        if self.authenticator not in [
+            "oauth",
+            "jwt",
+            "programmatic_access_token",
+            "workload_identity",
+        ]:
             if self.token:
                 warn_or_error(
                     AdapterEventWarning(
                         base_msg=(
-                            "The token parameter was set, but the authenticator was "
-                            "not set to 'oauth', 'jwt', or 'programmatic_access_token'."
+                            "The token parameter was set, but the authenticator was not set "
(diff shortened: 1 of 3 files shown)


dbt-snowflake/src/dbt/adapters/snowflake/connections.py
```diff
@@ -147,19 +147,26 @@ def __post_init__(self):
...
             )
 
-        if self.authenticator not in ["oauth", "jwt", "programmatic_access_token"]:
+        authenticator = (self.authenticator or "").lower()
+        if authenticator not in [
+            "oauth",
+            "jwt",
+            "programmatic_access_token",
+            "workload_identity",
+        ]:
```
aahel: we currently only allow workload_identity to be case insensitive. others are not. So if someone configures `Jwt` they would not get warned any more

check line 258 in fun auth_args in same file

aahel:
```
  elif self.authenticator == "jwt":
                # If authenticator is 'jwt', then the 'token' value should be used
                # unmodified. We expose this as 'jwt' in the profile, but the value
                # passed into the snowflake.connect method should still be 'oauth'
                result["token"] = self.token
                result["authenticator"] = "oauth"

            elif self.authenticator == "programmatic_access_token":
                # Snowflake Programmatic Access Tokens (PATs) are passed directly
                # to the connector via the `token` field. The connector handles
                # PAT auth natively since snowflake-connector-python v3.12.0.
                result["token"] = self.token

            elif self.authenticator.lower() == "workload_identity":
                result["authenticator"] = WORKLOAD_IDENTITY_AUTHENTICATOR

```

aahel (approved):

PR #2035, opened 2026-06-18
fix(spark): default username to 'dbt' when not set for SASL PLAIN auth

aahel:
## Problem

When `use_ssl: true` and `auth: NONE` (or `LDAP`/`CUSTOM`) are set but `user` is not configured in the profile, `username=None` is passed into `build_ssl_transport`. This causes puresasl's `PlainMechanism.process()` to fail:

```python
# puresasl/mechanisms.py
def _fetch_properties(self, *properties):
    needed = [p for p in properties if getattr(self, p, None) is None]
    if needed and not self.sasl.callback:
        raise SASLError('The following properties are required, but a '
                        'callback has not been set: %s' % ', '.join(needed))
```

`PlainMechanism` lists `username` as required. With `None`, it goes into `needed`, no callback is registered, and `SASLError` is raised. `PureSASLClient.start()` catches this and returns `ret=False`, causing `TSaslClientTransport.open()` to raise `TTransportException: Could not start SASL: The following properties are required, but a callback has not been set: username` — before sending any bytes.

This was confirmed by customer NLB access logs showing side-by-side:
```
Local (success):  duration=5826ms  sent=765  received=537  tlsv13
dbt Cloud (fail): duration=6ms     sent=0    received=0    tlsv13
```

TLS completes on both sides, but dbt Cloud sends zero bytes and disconnects in 6ms — consistent with a failure inside `sasl_client_factory()` or `sasl.start()` before the SASL START frame is ever transmitted.

## Solution

Mirror the existing `password` guard — default `username` to `"dbt"` when not set. For `auth: NONE`, the server accepts any credentials, so the value doesn't matter.

```python
if username is None:
    # Username doesn't matter in NONE mode, just needs to be nonempty.
    username = "dbt"
if password is None:
    # Password doesn't matter in NONE mode, just needs to be nonempty.
    password = "x"
```

Note: this is not a puresasl bug — SASL PLAIN (RFC 4616) requires both `authcid` and `passwd` in the protocol token. puresasl correctly enforces this. The fix belongs in dbt-spark.

## Test plan

- [x] `dbt debug` with `use_ssl: true`, `auth: NONE`, and no `user` set now succeeds against a Spark ThriftServer with `hive.server2.authentication=NONE`
- [x] `dbt debug` with `user` explicitly set is unchanged
- [x] Verified locally against a real Spark ThriftServer (Docker, `apache/spark:3.5.4`) over TLS + SASL PLAIN via bore tunnel, both with and without `user` set
- [x] Unit test added verifying `build_ssl_transport` defaults `username` to `"dbt"` when `None`

### DIFF
--- a/dbt-spark/src/dbt/adapters/spark/connections.py
+++ b/dbt-spark/src/dbt/adapters/spark/connections.py
@@ -722,6 +722,10 @@ def build_ssl_transport(
             sasl_auth = "GSSAPI"
         else:
             sasl_auth = "PLAIN"
+            if username is None:
+                # Username doesn't matter in NONE mode, just needs
+                # to be nonempty.
+                username = "x"
             if password is None:
                 # Password doesn't matter in NONE mode, just needs
                 # to be nonempty.
--- a/dbt-spark/src/dbt/adapters/spark/connections.py
+++ b/dbt-spark/src/dbt/adapters/spark/connections.py
@@ -665,8 +665,9 @@ class SparkConnectionManager(SQLConnectionManager):
                     logger.warning(msg)
                     time.sleep(creds.connect_timeout)
                 else:
-                    logger.warning(f"Failed to connect: {type(e).__name__}: {e}")
-                    raise FailedToConnectError("failed to connect") from e
+                    raise FailedToConnectError(
+                        f"failed to connect: {type(e).__name__}: {e}"
+                    ) from e
         else:
             raise exc  # type: ignore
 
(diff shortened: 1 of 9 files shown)


dbt-spark/src/dbt/adapters/spark/connections.py
```diff
@@ -722,6 +722,10 @@ def build_ssl_transport(
             sasl_auth = "GSSAPI"
         else:
             sasl_auth = "PLAIN"
+            if username is None:
+                # Username doesn't matter in NONE mode, just needs
+                # to be nonempty.
+                username = "x"
```
colin-k-rogers:
```suggestion
                username = "dbt"
```
nit: `x` feels a little random

colin-k-rogers:
Can we add a unit test here?

colin-k-rogers (approved):

PR #2042, opened 2026-06-26
feat(snowflake): add interactive table materialization support

### LINKED ISSUE #1780: [Feature] Snowflake Interactive Table support

### Describe the feature

Hello,
Snowflake recently GAed interactive analytics workloads using a new table type (and a warehouse type).
We expect customers to want to materialize DBT models as Interactive tables and would appreciate your support in bringing this to the dbt-Snowflake adapter and documentation.
The Interactive table is fairly simple from a syntax POV - it's either dynamic and resembles a dynamic table in syntax or does a full rewrite, which is perfect for DBT CTAS.

Adding some documentation here:
[Snowflake interactive tables and interactive warehouses](https://docs.snowflake.com/en/user-guide/interactive)
[CREATE INTERACTIVE TABLE](https://docs.snowflake.com/en/sql-reference/sql/create-interactive-table)
Notice that (conveniently) there is no ALTER INTERACTIVE TABLE
[SHOW TABLES](https://docs.snowflake.com/en/release-notes/bcr-bundles/2026_01/bcr-2165) has an is_adaptive boolean field to detect interactive tables 

Thanks!

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

itsnamangoyal:
### Problem

Snowflake's interactive tables (now GA) — optimized for low-latency, high-concurrency queries — had no dbt materialization.

### Solution

This carries forward the great community contribution from @jimmyxie-figma in #1798, which added the original implementation. That PR stalled while awaiting review feedback, so we've picked it up and applied the requested fixes so it can land against `main`. Many thanks to @jimmyxie-figma for the original work! 🙏

Adds a `materialized='interactive_table'` materialization supporting static and dynamic (`target_lag`) variants. Since Snowflake has no `ALTER INTERACTIVE TABLE`, all rebuilds go through `CREATE OR REPLACE`:

- **Static** (`cluster_by` only): doesn't auto-refresh, and `SHOW INTERACTIVE TABLES` returns no query to diff — so it rebuilds via `CREATE OR REPLACE` on every run, like a normal `table`. A SQL change is reflected on the next run.
- **Dynamic** (`target_lag` + `snowflake_warehouse`): uses config-change detection on `cluster_by` / `target_lag` / `snowflake_warehouse` (matching `dynamic_table`). A SQL-only change is picked up with `--full-refresh`.

### Interactive warehouse attachment

Adds a `snowflake_interactive_warehouses` config (a warehouse name or a list). After each run the table is attached to every listed interactive warehouse via `ALTER WAREHOUSE <wh> ADD TABLES (<this>)` — modeled on `grants`, so it runs every run and the config stays authoritative.

```sql
{{ config(
    materialized='interactive_table',
    cluster_by='id',
    snowflake_interactive_warehouses=['my_iw'],
) }}
```

#### Behavior & limitations

- **Additive only.** Attach runs every run (idempotent). Removing a warehouse from the list detaches it only on a static rebuild — `CREATE OR REPLACE` drops all attachments; a dynamic table that no-ops will not detach a de-listed warehouse. Full reconcile (detaching de-listed warehouses) is intentionally out of scope.
- **No slot pre-check.** Snowflake enforces the 10-table-per-warehouse cap and the `MODIFY` requirement; an over-cap or unauthorized attach surfaces the Snowflake error directly rather than a dbt-side pre-check.
- dbt does not create/alter/drop warehouses — it only attaches to pre-existing ones, and the run's role must hold `MODIFY` on each.

Review fixes applied on top of the original branch:
- Metadata is now read from `SHOW INTERACTIVE TABLES` (using the `refresh_warehouse` and `target_lag` columns).
- Surrounding parens are stripped from `cluster_by` values to avoid spurious rebuilds on unchanged tables.
- The introspected `query` field is now optional — `SHOW INTERACTIVE TABLES` returns a NULL query for static tables, which previously crashed the second run. Verified end-to-end against a live Snowflake account, with unit + functional regression tests added.

Resolves #1780.


dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -421,8 +424,12 @@ def list_relations_without_caching(
         schema_functions = schema_functions.rename(
             column_names=[col.lower() for col in schema_functions.column_names]
         )
+        available_columns = [c.lower() for c in schema_objects.column_names]
```
colin-k-rogers: I think we can simplify this by just checking if the column exists and if not populate the column with None or False

dbt-snowflake/src/dbt/include/snowflake/macros/relations/interactive_table/describe.sql
```diff
@@ -0,0 +1,4 @@
+{% macro snowflake__describe_interactive_table(relation) %}
```
colin-k-rogers: Where is this used?

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/interactive_table.sql
```diff
@@ -0,0 +1,94 @@
...
+    {{ return({'relations': [target_relation]}) }}
+
+{% endmaterialization %}
+
+
+{% macro interactive_table_get_build_sql(existing_relation, target_relation) %}
+
+    {% set full_refresh_mode = should_full_refresh() %}
+
+    {% if existing_relation is none %}
```
colin-k-rogers, line 34: As I read it this logic would result in us not updating the table when the SQL changes. At the very least we need to look at if the SQL changed. More likely we should treat "interactive tables" like regular tables and re-build on every run.

Even more broadly: should "interactive" be a net new "materialization" or should it be an option on the table materialization?

colin-k-rogers (changes requested):
A couple things to consider but the biggest of which: is an interactive table a totally new materialization or is it a variation on the pre-existing table materialization?

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/interactive_table.py
```diff
@@ -0,0 +1,172 @@
+from dataclasses import dataclass
+from typing import Optional, Dict, Any, TYPE_CHECKING
+
+from dbt.adapters.relation_configs import RelationConfigChange, RelationResults
+from dbt.adapters.contracts.relation import RelationConfig
+from dbt.adapters.contracts.relation import ComponentName
+from dbt_common.exceptions import CompilationError
```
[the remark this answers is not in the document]
itsnamangoyal: It's used in line 45, 48 and 51.

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/interactive_table.sql
itsnamangoyal: Good questions — dug into this and here's where I landed. Sharing the reasoning; happy to change course if you see it differently.

On "should interactive be its own materialization or an option on the table materialization?" — I'm going with its own materialization. Here's why.

Interactive tables actually have two behaviors, and they need opposite handling on each run:

- Static (no target_lag) — doesn't auto-refresh. Snowflake only updates it on CREATE OR REPLACE. So dbt should rebuild it every run, exactly like a normal table.
- Dynamic (target_lag set) — Snowflake auto-refreshes it continuously. Rebuilding it every run would throw that away and pay a full recompute each time. So dbt should not rebuild unless the definition changed — exactly like a dynamic_table.

No single existing materialization does both. Making it an option on table gets the static case right but is wrong for the dynamic one (and table has nowhere to put the "skip unless changed" logic). A dedicated materialization can branch on target_lag and do the right thing for each.

That branch is also the real fix for your other comment: today the static case wrongly uses the dynamic "no-op if config unchanged" path, so it doesn't rebuild when the SQL/source changes. Fixing it: static → always CREATE OR REPLACE; dynamic → replace only when the query or config changed (interactive tables have no ALTER, so any change is just a replace — simpler than dynamic tables).

It also keeps the upcoming interactive warehouse work clean — that's a separate object that attaches to tables via TABLES(...), and the refresh warehouse on a dynamic interactive table has to be a standard warehouse, not an interactive one. Keeping interactive tables as their own relation type makes them easy to identify/attach later without entangling the two.

So I'm going ahead with the dedicated materialization and adding the static/dynamic branch. If you'd still rather make it an option on table, happy to switch — let me know.

itsnamangoyal: I ended up keeping the dedicated interactive_table materialization, and split the two variants by behavior:

- Static (cluster_by only, no target_lag): these don't auto-refresh and Snowflake returns no query text to diff, so there's nothing to compare against. I made them always CREATE OR REPLACE on every run — same as a normal table. That's what fixes the SQL-change case you flagged: a query edit is now reflected on the next run.
- Dynamic (target_lag + warehouse): these keep config-change detection on cluster_by/target_lag/warehouse.

One correction to what I said earlier: I'd mentioned diffing the query for dynamic tables. I dropped that — it matches how dynamic_table already behaves (config changes are detected; a SQL-only change needs --full-refresh). Parsing the SHOW DDL back into a comparable query is fragile, and I'd rather stay consistent with the existing materialization than invent a one-off.

There's an interactive-warehouse interaction I want your read on. CREATE OR REPLACE detaches the table from an interactive warehouse's TABLES set (serving still works, but the attachment is dropped — I verified this against Snowflake). So the open question is whether dbt should re-attach after a build at all. Doing so is awkward: dbt doesn't manage warehouse lifecycle today, it'd need a MODIFY grant on a pre-existing warehouse, there's a cap of 10 tables per interactive warehouse, and it couples a model build to external warehouse state. Do you think re-attaching is something dbt should own — and if so, this PR or a followup? My lean is a followup, keeping this PR to the materialization itself.

There's also an alternative to CREATE OR REPLACE: INSERT OVERWRITE refreshes the data in place and preserves the attachment (and grants). The catch is it can't change the schema, so we'd need to detect whether the columns changed — column changes still require a full CREATE OR REPLACE (and the detach that comes with it). That tradeoff feels like it belongs with the re-attach question rather than here.

Tests cover both variants: static rebuilds every run + reflects SQL changes; dynamic config-change apply/continue/fail + no-op reruns. All green against Snowflake.

Happy to revisit if you'd prefer a different split.

dbt-snowflake/src/dbt/include/snowflake/macros/relations/interactive_table/describe.sql
itsnamangoyal: Remove this. Claude created this to match dynamic table's pattern. describe macro for dynamic table is also unused. Shall I remove it as well in this PR?

dbt-snowflake/.changes/unreleased/Features-20260323-100000.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Features
+body: Add Snowflake interactive table materialization (`materialized='interactive_table'`), supporting static and dynamic (`target_lag`) variants.
+time: 2026-03-23T10:00:00-07:00
+custom:
+    Author: jimmyxie-figma
```
colin-k-rogers:
```suggestion
    Author: jimmyxie-figma itsnamangoyal
```

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/interactive_table.sql
colin-k-rogers: I think we should add a new "attach" model config similar to grants, if a user specifies it then we will run attach at the end of the DDL. This would also support users wanting to switch which interactive warehouse a model is attached to. I'm fine deferring that to a follow up issue or including in this PR.

itsnamangoyal: Thanks for this — the grants analogy was exactly the right call and made the design obvious. 🙏

Implemented it in this PR as a snowflake_interactive_warehouses config (a warehouse name or list), modeled on grants: after each run the table is attached to every listed warehouse via ALTER WAREHOUSE <wh> ADD TABLES (<this>), and like grants it runs every run so the config stays authoritative. Switching warehouses works just the way you described — edit the list and the next run reflects it.

Two things worth calling out from testing against a live account:

- A table can be attached to multiple interactive warehouses at once (verified), so the config is a list and attach is additive. CREATE OR REPLACE detaches from all warehouses, so a static rebuild cleanly re-establishes exactly the listed set. The one gap: a dynamic table that no-ops won't detach a warehouse you've removed from the list until it next rebuilds — full reconcile (actively detaching de-listed warehouses) is out of scope here, and it's noted in the PR body.
- No slot pre-check — Snowflake enforces the 10-table-per-warehouse cap and the MODIFY requirement, so an over-cap or unauthorized attach surfaces its error directly rather than a dbt-side guard. dbt never creates/alters/drops the warehouse itself; it only attaches to a pre-existing one.

Tests cover attaching to one and to multiple warehouses, plus re-attach on a dynamic no-op run.

Really appreciate the steer here — it landed cleanly.

colin-k-rogers:
Took another pass after the rework — the static/dynamic split looks great. Static interactive tables now `CREATE OR REPLACE` every run like a plain table (with the SQL-change and rebuild-every-run functional tests to prove it), and change-detection is cleanly scoped to the dynamic `target_lag` variant. The `is_interactive` column defaulting + single-form unpacking in `_parse_list_relations_result` is also much nicer than the earlier flag-threaded version. 👍

**Blocking:**

**Warehouse names are interpolated raw in the attach DDL.**
In `snowflake__attach_interactive_warehouses`:
```sql
alter warehouse {{ warehouse }} add tables ({{ relation }})
```
`{{ relation }}` renders as a proper quoted identifier, but `{{ warehouse }}` is dropped in verbatim from config with no identifier handling. Any warehouse name that isn't a bare uppercase-safe identifier — mixed case, spaces, reserved words — produces broken or wrong-target DDL (e.g. `my-wh` or a case-sensitive name silently resolves to the wrong/nonexistent warehouse). Since this value drives an `ALTER WAREHOUSE` against a user-supplied name, it should go through proper identifier rendering (e.g. `adapter.quote(...)`) rather than raw interpolation.

One thing to confirm when you fix it: quoting changes Snowflake's case semantics (an unquoted `foo` resolves to `FOO`, but `"foo"` is case-sensitive), so decide whether the config value is meant to be a verbatim identifier or a quoted one, and make the rendering + docs consistent. The current raw form is the worst of both — it neither quotes nor documents a verbatim contract.

**Non-blocking follow-ups:**

**1. `cluster_by` / `snowflake_warehouse` change-detection is format/case-sensitive (dynamic variant only).**
`parse_relation_results` strips only a bare surrounding `(...)` and doesn't normalize case or a `LINEAR(...)` wrapper, and the warehouse comparison is raw string equality. If `SHOW INTERACTIVE TABLES` ever returns clustering keys uppercased or `LINEAR(...)`-wrapped, a dynamic table would spuriously rebuild every run. The dynamic no-op / multicol functional tests guard this, so it holds as long as those run in CI against a live account — the unit tests mock the SHOW output, so they won't catch a format surprise.

**2. `requires_full_refresh` is dead code.**
Since there's no `ALTER INTERACTIVE TABLE`, every change replaces, and the materialization only branches on `configuration_changes is none` — it never reads `requires_full_refresh` on the changeset or the three `...ConfigChange` classes. Could drop it, or leave a one-line comment noting it exists only to satisfy the `RelationConfigChange` contract.

itsnamangoyal:
@colin-rogers-dbt Thanks for the reviewing pass. Addressed all three:

Blocking — raw warehouse interpolation. Good catch. I went with a verbatim identifier contract rather than adapter.quote(), to stay consistent with snowflake_warehouse (which is interpolated verbatim in the create/replace DDL two lines away) and to preserve Snowflake's uppercase folding — quoting analytics_wh → "analytics_wh" would make it case-sensitive and break the common bare-name case. The macro now documents the contract explicitly, and the PR body / config docs spell it out: write the name as Snowflake expects, and wrap case-sensitive or special-character names in double quotes in the config string (e.g. '"my-wh"'). (cbb2114b)

1 — format/case-sensitivity of change detection. Left the behavior as-is since it matches dynamic_table (same bare-paren strip, raw warehouse equality, no LINEAR(...)/case normalization), and added a comment at parse_relation_results recording exactly what you noted: the dynamic no-op / multicol functional tests are the guard against a live-account format surprise; the unit tests mock SHOW output so they won't catch one. (7a4e59b9)

2 — dead requires_full_refresh. Dropped the changeset-level property (nothing reads it — no ALTER INTERACTIVE TABLE, so the materialization only branches on configuration_changes is none). Kept the three ...ConfigChange properties since they satisfy the abstract RelationConfigChange contract, with a one-line comment saying so. (ce80bbc8)

colin-k-rogers (approved):

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -719,6 +739,63 @@ def _query_dynamic_table_transient_status(self, relation: SnowflakeRelation) ->
...
+        database = f'"{relation.database}"' if quoting.database else relation.database
+        show_sql = (
+            f"show interactive tables like '{relation.identifier}' in schema {database}.{schema}"
+        )
+        res, tables_table = self.execute(show_sql, fetch=True)
+        if res.code != "SUCCESS":
+            raise DbtRuntimeError(f"Could not get interactive table metadata: {show_sql} failed")
+
+        tables_table = tables_table.rename(
+            column_names=[name.lower() for name in tables_table.column_names]
```
xuliangs, line 765: I don't think we should lower case on the table names here since it preserves the original casing when those tables were created.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -719,6 +739,63 @@ def _query_dynamic_table_transient_status(self, relation: SnowflakeRelation) ->
...
+        # unquoted identifiers are stored uppercase in Snowflake metadata.
+        if quoting.identifier:
+            exact_match = tables_table.where(lambda row: row.get("name") == relation.identifier)
+        else:
+            _identifier_upper = (relation.identifier or "").upper()
+            exact_match = tables_table.where(
+                lambda row: row.get("name", "").upper() == _identifier_upper
+            )
+        if len(exact_match.rows) == 0:
+            raise DbtRuntimeError(f"Could not find interactive table: {relation.identifier}")
```
xuliangs, line 779: Does this mean changing a model from table (or another materialization) to an interactive table will fail?

xuliangs (approved):
My comments can be addressed in a separate PR. Thank you.

itsnamangoyal (commented):
(comment removed)

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
itsnamangoyal: Traced this — the .lower() here only renames the Agate result's column headers (NAME→name, etc.) for dict-style access; it never touches the identifier values. Verified live with a quoted mixed-case table ("ScenarioMixedCase"): SHOW INTERACTIVE TABLES returned the name with casing intact, and the exact-match lookup found it correctly. No casing bug here.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
itsnamangoyal: Tested this live — table → interactive_table conversion works for both the static and dynamic (target_lag) variants. The not existing_relation.is_interactive_table check in the materialization routes straight to CREATE OR REPLACE on the swap, bypassing describe_interactive_table entirely. A follow-up unchanged run on the dynamic variant also correctly no-op'd via config-change detection. No fix needed.

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/interactive_table.py
```diff
@@ -0,0 +1,167 @@
...
+            "cluster_by": cluster_by_val,
+            "target_lag": interactive_table.get("target_lag"),
+            "snowflake_warehouse": interactive_table.get("refresh_warehouse"),
+        }
+
+        return config_dict
+
+
+# There is no ALTER INTERACTIVE TABLE, so every change replaces and the materialization
+# only branches on `configuration_changes is none`. requires_full_refresh is never read; it
```
hope-wat, line 124: Did we pressure test the idempotent nature of dbt with how we are bringing in snowflake interactive tables? What is the intended `--full-refresh` behavior?

itsnamangoyal: **Static interactive tables** (`cluster_by` only):
- Always rebuilds via `CREATE OR REPLACE` every run. This is unavoidable  because Snowflake returns `NULL` for query text in `SHOW INTERACTIVE TABLES`, so there's no way to diff and check if it changed..
- Fully idempotent — running twice produces identical results.
- SQL changes are reflected on the next run because we rebuild every time anyway.
- Warehouse attachment: CREATE OR REPLACE automatically drops the old table and detaches it from all warehouses. Then dbt re-attaches to every listed warehouse via ALTER WAREHOUSE ... ADD TABLES(), making the config authoritative each run.

**Dynamic interactive tables** (`target_lag` + `warehouse`):
- Config-change detection on `cluster_by`, `target_lag`, `snowflake_warehouse`.
- No rebuild if config unchanged → skip `CREATE OR REPLACE`, preserve auto-refresh window.
- Idempotent when config unchanged; safe when config changes (all changes trigger rebuild).
- Warehouse attachment runs similarly to static.

**Test coverage**: All testing performed against live Snowflake accounts (both unit/functional tests in the PR and separate live project validation).

### **--full-refresh Behavior**

For dynamic tables: `--full-refresh` forces a rebuild regardless of config changes (bypasses the change-detection no-op). Intended behavior to keep it aligned with "normal" `dynamic_table`.

For static tables: `--full-refresh` has no additional effect since we are anyways rebuilding on every run.

hope-wat: I don't think this addresses what I am asking about because this response (some of it likely AI) is referring to dynamic tables. Dynamic tables and Interactive tables are fundamentally different -- dynamic is for real time data coming in while interactive is for low query latency.

itsnamangoyal: It is partially generated by AI. But interactive tables have 2 types: static and dynamic. Dynamic interactive table accepts target lag to keep the data up to date.

PR #2047, opened 2026-07-03
feat(athena): add S3 Tables catalog support

### LINKED ISSUE #1186: [Feature] Add S3 Tables support to dbt-athena adapter

aahel:
resolves #1186 

### Problem

dbt-athena had no way to write Iceberg models to AWS S3 Tables. S3 Tables is a managed Iceberg service where AWS controls storage — unlike Glue Iceberg where users specify an S3 path. The existing `glue` catalog type emits a `LOCATION` in DDL, deletes S3 data directly, and uses an `ALTER TABLE RENAME` swap for near-zero-downtime replacement — all three are rejected by S3 Tables.

### Solution

Adds a new `catalog_type: s3_tables` catalog integration, plus the adapter changes needed to make it work end-to-end (create, query, and idempotent re-runs).

**DDL (`create_table_as.sql`)** — when the catalog is `s3_tables`:
- No `LOCATION` in DDL (S3 Tables manages storage; AWS rejects `location=`)
- No `delete_from_s3()` (managed bucket forbids direct S3 access)
- No unique-location validation

**Metadata / re-run path** (discovered via live testing against a real S3 Tables catalog):
- `_get_data_catalog` recognizes `s3tablescatalog/<bucket>` as a Glue federated catalog and builds the descriptor directly, instead of calling the Athena `get_data_catalog` API (which 404s for these catalogs)
- `get_table_type` gives the iceberg `table_type` parameter precedence over the Glue `TableType`, so S3 Tables tables (which report `TableType="customer"`, not in `RELATION_TYPE_MAP`) classify as iceberg instead of raising
- Table replacement uses drop-and-recreate — S3 Tables does not support `ALTER TABLE RENAME` (`"Unsupported DDL query for S3 table buckets"`). The drop routes through the **Glue Data Catalog metadata delete** (`delete_from_glue_catalog` against the `<account>:s3tablescatalog/<bucket>` catalog id, which delegates to S3 Tables and removes the data), **not** a SQL `DROP TABLE` and **not** `delete_from_s3`. A SQL drop can't target a federated sub-catalog — Athena's Hive `DROP TABLE` is limited to `<schema>.<table>` in the query's default catalog (`awsdatacatalog`), so it silently no-ops and the re-run then fails `TABLE_ALREADY_EXISTS`; direct S3 delete is skipped because S3 Tables manages its own storage.

**User config:**
```yaml
# catalogs.yml (v2)
catalogs:
  - name: my_s3_tables
    type: s3_tables
    table_format: iceberg
    config:
      athena:
        catalog_database: s3tablescatalog/my-table-bucket
        file_format: parquet
```
```sql
{{ config(materialized='table', catalog_name='my_s3_tables', schema='my_namespace') }}
select 1 as id
```
`catalog_database` is the federated catalog name for the S3 Tables bucket (`s3tablescatalog/<bucket>`) — the same name Athena uses to address it. It routes the model's `database` via `generate_database_name` (the merged universal `catalog_database` field), exactly like every other catalog type. The model's `schema` is the S3 Tables namespace. Omitting `catalog_database` falls back to the model/profile `database`.

### Validation

Verified end-to-end against a live S3 Tables catalog in `us-east-2` (SQL engine v3, Glue federation via `aws glue create-catalog`):
- First `dbt run` creates the Iceberg table (correct DDL, no `location=`)
- Repeated `dbt run` (idempotency) succeeds — the existing table is dropped via the Glue Data Catalog and recreated
- `SELECT` returns expected data; the table is an S3 Tables managed table (`TableType="customer"`) and Glue reports `table_type=ICEBERG`
- Routing verified with `catalog_database` set in `catalogs.yml` and no `database` on the model
- Confirmed AWS rejects `location=` and `ALTER TABLE RENAME` for S3 Tables, justifying the special-casing

Regular Glue path is unchanged: the new `_get_data_catalog` branch only matches `s3tablescatalog/*`, and `get_table_type` is identical for `EXTERNAL_TABLE` iceberg tables.

Companion to #2016 (catalogs.yml v2 base support); builds on the merged universal `catalog_database` field for database routing.

### Checklist


dbt-athena/src/dbt/include/athena/macros/materializations/models/table/table.sql
```diff
@@ -99,6 +99,17 @@
           {{ query_result }}
         {% endcall %}
       {%- endif -%}
+    {%- elif adapter.is_s3_tables_database(database) -%}
```
colin-k-rogers, line 102: Do we need this same guard for incremental/snapshot materializations?

colin-k-rogers:
We need to add functional tests that actually validate this works with s3 tables

dbt-athena/src/dbt/include/athena/macros/materializations/models/table/table.sql
aahel: Thnx for pointing this out.
incremental needed a different guard. Its create/merge/full-refresh paths work, but it failed at the end at`expire_glue_table_versions` (Glue GetTableVersions 404s on the federated catalog), so I skipped it for s3_tables (S3 Tables manages its own versioning). Snapshots use MERGE and work as-is. Added functional tests covering table, incremental, and snapshot into a live S3 Tables catalog.

aahel:
> We need to add functional tests that actually validate this works with s3 tables

done

dbt-athena/src/dbt/adapters/athena/impl.py
```diff
@@ -917,14 +939,24 @@ def _get_data_catalog(self, database: str) -> Optional[DataCatalogTypeDef]:
...
+            # descriptor directly instead of calling athena.get_data_catalog (which 404s).
+            if database.lower() == "awsdatacatalog" or self.is_s3_tables_database(database):
                 with boto3_client_lock:
                     sts = client.session.client(
                         "sts",
                         region_name=client.region_name,
                         config=get_boto3_config(num_retries=creds.effective_num_retries),
                     )
-                catalog_id = sts.get_caller_identity()["Account"]
+                account_id = sts.get_caller_identity()["Account"]
```
colin-k-rogers: I realize this is a pre-existing issue so we can consider it out of scope but... can we just call this once when we initialize the connection instead of needing to do it on every call?

colin-k-rogers (approved):

dbt-athena/src/dbt/adapters/athena/impl.py
aahel: done [bd0acf0](https://github.com/dbt-labs/dbt-adapters/pull/2047/commits/bd0acf0ada3c706ca8d5aa5f78e4c3e2341102bc).

Issue #626, opened 2024-09-06
[Spike] Conflict with concurrent transaction


jennlw:
### Is this a new bug in dbt-redshift?

- [X] I believe this is a new bug in dbt-redshift
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Job runs sporadically hitting "conflict with concurrent transaction error." The failure is inconsistent as the customer cannot reproduce it on every execution. On one of their dbt Cloud environment (5 threads) it happened twice but on our their local DBT core instances(20+ threads) they were able to reproduce it every third execution. Their local execution has another model victim.

They checked their logs and were not able to locate any other thread running parallel execution that would cause conflict.

### Expected Behavior

Job run does not fail on concurrent transactions. 

### Steps To Reproduce

Not yet, but customer mentioned: 

> "Neither of the models are using incremental materialization, only thing that might be related to this failure is that these models being used in the materialized view downstream."

I had ran this past our team internally and it was mentioned:

> I suspect this could be the culprit. I remember that querying the object dependency tree in Redshift is not easy when using MVs. So I wonder if dbt is dropping the MV in parallel to the refresh of this table and that's attempting to drop this table at the same time. 

### Relevant log output

```shell
2024-09-03 00:40:17.026263 (Thread-3 (worker)): 00:40:17 On model.A_dbt.modelA: /* {"app": "dbt", "dbt_version": "1.7.17", "profile_name": "user", "target_name": "default", "node_id": "model.A_dbt.modelA"} */ drop table if exists "abcd"."dbt_1234"."modelA__dbt_backup" cascade

2024-09-03 00:40:22.922586 (Thread-3 (worker)): 00:40:22 Redshift adapter: Redshift error: could not complete because of conflict with concurrent transaction 2024-09-03 00:40:22.923500 (Thread-3 (worker)): 00:40:22 On model.A_dbt.modelA: ROLLBACK 2024-09-03 00:40:23.147961 (Thread-3 (worker)): 00:40:23 Redshift adapter: Error running SQL: macro drop_relation
```

### Environment

```markdown
- OS:
- Python:
- dbt-core: 1.7.17 (cloud)
- dbt-redshift: 1.7.7
```

### Additional Context

Please feel free to reach out to me and we can further discuss logs. I did not add them as they are customer logs and not dbt sandbox or test logs from dbt.

amit-jobadder:
[...]

PR #2049, opened 2026-07-03
fix(redshift): serialize rename swap to avoid losing __dbt_tmp on conflict

itsnamangoyal:
## Summary
- Adds a `rename_relation` override to `RedshiftAdapter` that wraps the rename in `fresh_transaction()`, mirroring the existing `drop_relation()` CASCADE-lock pattern from #626.
- Commits the intermediate (`__dbt_tmp`) relation before the target->backup rename is attempted, so a `conflict with concurrent transaction` error there can no longer discard it and mask itself as `relation ... does not exist` on the follow-up rename.
- Fixes CORE-724.

## Test plan
- [x] New regression test `test_rename_relation.py::TestRenameRelationLock::test_rename_relation_uses_fresh_transaction` (written RED-first, now GREEN)
- [x] Full `dbt-redshift` unit suite passes (170/170), no regressions
- [x] Validated against a live Redshift cluster: baseline `table` materialization builds and full-refreshes succeed with the fix; forced real timing overlap between a concurrent `VACUUM` and the rename step twice with no corruption

### DIFF
--- a/dbt-redshift/src/dbt/adapters/redshift/impl.py
+++ b/dbt-redshift/src/dbt/adapters/redshift/impl.py
@@ -178,6 +178,26 @@ class RedshiftAdapter(SQLAdapter):
         with self.connections.fresh_transaction():
             return super().drop_relation(relation)
 
+    def rename_relation(self, from_relation, to_relation):
+        """
+        The table materialization's target<->backup<->intermediate swap issues
+        two RENAMEs inside one build transaction. If Redshift aborts that
+        transaction on a "conflict with concurrent transaction" error during
+        the first rename, the intermediate relation created earlier in the
+        same transaction is rolled back with it. execute()'s retry then
+        resends only the failed RENAME statement, so the second RENAME
+        (intermediate -> target) fails with "relation ... does not exist",
+        masking the original conflict.
+
+        Wrapping the rename in fresh_transaction() commits the in-progress
+        transaction (and the intermediate relation with it) before attempting
+        the rename, so a conflict here can no longer take the intermediate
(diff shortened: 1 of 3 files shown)


dbt-redshift/src/dbt/adapters/redshift/impl.py
```diff
@@ -178,6 +178,26 @@ def drop_relation(self, relation):
...
+        (intermediate -> target) fails with "relation ... does not exist",
+        masking the original conflict.
+
+        Wrapping the rename in fresh_transaction() commits the in-progress
+        transaction (and the intermediate relation with it) before attempting
+        the rename, so a conflict here can no longer take the intermediate
+        relation down with it. Mirrors the CASCADE handling in drop_relation
+        above.
+        """
+        with self.connections.fresh_transaction():
```
tauhid621, line 198: I am wondering if we should fix whole model transaction retry logic.

This approach breaks the atomic part of renaming during models which can cause stale tables.

We can look at a way where when we abort due to conflict or other reason/ we start the transactions form beginning or maybe start the model materialization again 🤔 .

itsnamangoyal: Good point — this does give up some atomicity to fix the misleading error. It's the same tradeoff drop_relation() already makes for CASCADE (#626), so it's not a new risk, just the same pattern applied here.

Fixing the retry logic itself would be a much bigger change — it affects every SQL statement on every adapter, not just Redshift renames. I'd say we can merge this PR and think about broader transaction handling sometime down the line.

PR #2051, opened 2026-07-06
Add a macro to check the v2 runtime 

xuliangs:
### Problem

Today, the Jinja macros are duplicated across the v1 and v2 adapters, even though they are intended to remain largely consistent. In v2, some divergences are unavoidable and need to branch based on the runtime. At the same time, backward compatibility is required so the v2 parser can correctly consume manifest.json files generated by the v1 runtime.

### Solution

This macro will also be available in Core v2 adapters, allowing us to branch logic where v1 and v2 diverge.
https://github.com/dbt-labs/fs/pull/10687

### Checklist


dbt-adapters/src/dbt/include/global_project/macros/adapters/runtime.sql
```diff
@@ -0,0 +1,4 @@
+{% macro is_core_v2() %}
+    {%- set major_version = (dbt_version.split('.') | first) | int -%}
+    {{ return(major_version >= 2) }}
```
colin-k-rogers: Okay super pedantic since it's going to be a long time (if ever) till we get there but shouldn't this be `==`?
```suggestion
    {{ return(major_version == 2) }}
```

xuliangs: Makes sense!

colin-k-rogers (approved):

PR #2055, opened 2026-07-08
fix: guard build_catalog_relation against non-mapping node config (ExportConfig)

aahel:
resolves #

### Problem

After #2004 ("support catalog_database on all catalog types") landed, downstream CI (dbt-core, dbt-common) started failing with:

```
AttributeError: 'ExportConfig' object has no attribute 'get'
```

on any `dbt` run touching a **saved query with exports** (e.g. `tests/functional/saved_queries/*`, `tests/functional/list/*`).

**Root cause:** #2004 rewrote `default__generate_database_name` to call `adapter.build_catalog_relation(node)` for *every* node with a config:

```jinja
{%- if node is not none -%}
    {%- set catalog_relation = adapter.build_catalog_relation(node) -%}
```

`build_catalog_relation` (unchanged by #2004, but now reached for far more node types) does `config.config.get(...)`, assuming a mapping. A saved-query export node's `config` is a typed `ExportConfig` (`dbt_semantic_interfaces.protocols.export.ExportConfig`) — properties only, **no `.get`** — and it's truthy, so it passes the `if not config.config` check and then raises on `.get`.

The macro can't guard this itself: a check like `node.config.get('catalog_name')` would hit the *same* AttributeError. The only safe place to detect "config isn't mapping-like" is inside `build_catalog_relation`.

### Solution

Return `None` from `build_catalog_relation` when `config.config` has no callable `.get` — such nodes can't reference a catalog anyway, so `generate_database_name` falls through to its normal precedence (`custom_database_name` > `target.database`). No macro change needed; fixes all current and future callers.

```python
get = getattr(config.config, "get", None)
if not callable(get):
    return None
```

### Tests

Adds `TestBuildCatalogRelation`: non-mapping config (ExportConfig-like) → `None` (no raise); empty config → `None`; mapping config without `catalog_name` → `None`.

### Note

There is a separate, related version-skew issue from #2004: dbt-core's `core/dbt/config/catalogs.py` needs a matching `catalog_database` field on its `CatalogWriteIntegrationConfig` (mypy: "Cannot instantiate abstract class ... with abstract attribute catalog_database"). That's a dbt-core-side fix, tracked/handled separately.

### Checklist

### DIFF
--- a/dbt-adapters/src/dbt/adapters/base/impl.py
+++ b/dbt-adapters/src/dbt/adapters/base/impl.py
@@ -209,6 +209,17 @@ def _relation_name(rel: Optional[BaseRelation]) -> str:
         return str(rel)
 
 
+def _config_get(config: Any, key: str) -> Any:
+    """Read ``key`` from a node config that may not be dict-like.
+
+    Some node configs are typed, non-mapping objects (e.g. a saved-query export's
+    ``ExportConfig``, which has no ``.get``). Return None for those rather than
+    raising AttributeError.
+    """
+    get = getattr(config, "get", None)
+    return get(key) if callable(get) else None
+
+
 def log_code_execution(code_execution_function):
     # decorator to log code and execution time
     if code_execution_function.__name__ != "submit_python_job":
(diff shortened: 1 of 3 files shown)


dbt-adapters/src/dbt/adapters/base/impl.py
```diff
@@ -392,10 +392,16 @@ def build_catalog_relation(self, config: RelationConfig) -> Optional[CatalogRela
         if not config.config:
             return None
 
+        # Catalogs are referenced via a mapping-style `catalog_name`/`catalog` key.
+        # Some node types expose a typed, non-mapping config object (e.g. a saved-query
+        # export's `ExportConfig`, which has no `.get`). Those can't reference a catalog,
+        # so bail out instead of raising AttributeError on `.get`.
+        get = getattr(config.config, "get", None)
```
ash2shukla: Kinda hard to read can u do something like this ?

```python
def get_config(key):
     if not hasattr(config.config, "get"):
            return None

     return config.config.get(key)

# get_config(CATALOG_INTEGRATION_MODEL_CONFIG_NAME) or get_config("catalog") will return None then so we can remove L400, 401 too
```

ash2shukla (approved):
lgtm with a nit

ash2shukla (approved):

PR #1988, opened 2026-05-29
fix(snowflake): Replace invalid FIXED dtype with NUMBER(p, s)

rio-not-real:
resolves #1986
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#

### Problem

Currently, `SnowflakeAdapter` inherits `get_column_schema_from_query()` from `BaseAdapter`. However, this method returns 'FIXED' for fixed-point data types, which is not a valid data type in Snowflake SQL. Because of this, when macro `default__snapshot_staging_table` in [helpers.sql](https://github.com/dbt-labs/dbt-adapters/blob/63140d8b177fdcba82a8b5292276fe5eff4c763f/dbt-adapters/src/dbt/include/global_project/macros/materializations/snapshots/helpers.sql#L174) calls it, it fails with the following:

```shell
11:50:25  Failure in snapshot test_snapshot (snapshots/_test__snapshot.yml)
11:50:25    Database Error in snapshot test_snapshot (snapshots/_test__snapshot.yml)
  001003 (42000): SQL compilation error:
  syntax error line 181 at position 25 unexpected 'FIXED'.
```

### Solution

Propose overriding `BaseAdapter.get_column_schema_from_query()` in `SnowflakeAdapter` to normalise the data type names returned from the Snowflake connector's `SnowflakeCursorBase.description` for fixed-point data types.

When `data_type_code_to_name()` returns `FIXED`, it will now be converted to a valid Snowflake numeric type:

- `NUMERIC(precision, scale)` when both precision and scale are available.
- `NUMERIC` when they're unavailable (added as a safety net, though I'm not sure in what scenario this would occur).

Non-fixed-point data types (which do not return `FIXED`) will continue to use the existing dtype name returned from `data_type_code_to_name()` without modification.

Unit tests added for:

- converting the `FIXED` type to `NUMERIC`.
- preserving the existing behavior for non-fixed-point data types.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #2060, opened 2026-07-09
fix(snowflake): cast FIXED dtype to NUMBER in snapshot deletion_records

### LINKED ISSUE #1986: [Bug] Snapshot with `hard_deletes: new_record` fails on Snowflake when relation adds a numeric column
### Is this a new bug?

- [x] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [x] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

When a snapshot uses `hard_deletes: new_record` and a new numeric column is added to the relation model, the snapshot fails with:

```shell
11:50:25  Failure in snapshot test_snapshot (snapshots/_test__snapshot.yml)
11:50:25    Database Error in snapshot test_snapshot (snapshots/_test__snapshot.yml)
  001003 (42000): SQL compilation error:
  syntax error line 181 at position 25 unexpected 'FIXED'.
```

This happens in the `deletion_records` CTE generated by the `default__snapshot_staging_table` macro in [helpers.sql](https://github.com/dbt-labs/dbt-adapters/blob/63140d8b177fdcba82a8b5292276fe5eff4c763f/dbt-adapters/src/dbt/include/global_project/macros/materializations/snapshots/helpers.sql#L188).

The generated SQL:

```logs/dbt.log
    deletion_records as (

        select
            'insert' as dbt_change_type,
            snapshotted_data."PK",snapshotted_data."UPDATED_AT",snapshotted_data."COL1",
    
        try_cast(NULL as FIXED)  -- `FIXED` is not a valid Snowflake SQL data type
     as "COL2",snapshotted_data.dbt_unique_key as dbt_unique_key,
```

### Expected Behavior

The generated SQL should use a valid Snowflake [data type for fixed-point numbers](https://docs.snowflake.com/en/sql-reference/data-types-numeric#data-types-for-fixed-point-numbers) instead of `FIXED`.

For example with `dbt-snowflake`

```sql
{{ safe_cast('NULL', col.dtype) }}
```

should be parsed into:

```sql
try_cast(NULL as NUMBER(38, 0))
-- or
try_cast(NULL as NUMBER)
```

### Steps To Reproduce

1. Create a test model with at least one row:

    ```sql
    # models/test.sql
    {{ config(materialized='ephemeral') }}
    select
      'pk1' as pk
      , '2026-05-28T00:00:00Z'::timestamp_ntz as updated_at
      , 1 as col1
    /*
    * | pk  | updated_at          | col1 |
    * | --- | ------------------- | ---- |
    * | pk1 | 2026-05-28 00:00:00 | 1    |
    */
    ```

2. Create a snapshot with `hard_deletes: new_record` for the test model:

    ```yaml
    # snapshots/_test__snapshot.yml
    snapshots:
      - name: test_snapshot
        relation: ref('test')
        config:
          strategy: timestamp
          updated_at: updated_at
          unique_key: pk
          hard_deletes: new_record
    ```

3. Run `dbt snapshot -s test_snapshot` once to create the snapshot table

4. Add a new numeric column to the test model:

    ```sql
    # models/test.sql
    {{ config(materialized='ephemeral') }}
    select
      'pk1' as pk
      , '2026-05-28T00:00:00Z'::timestamp_ntz as updated_at
      , 1 as col1
      , null as col2
    union all
    select
      'pk2' as pk
      , '2026-05-29T00:00:00Z'::timestamp_ntz as updated_at
      , 1 as col1
      , 2 as col2
    /*
    * | pk  | updated_at          | col1 | col2 |
    * | --- | ------------------- | ---- | ---- |
    * | pk1 | 2026-05-28 00:00:00 | 1    | null |
    * | pk2 | 2026-05-29 00:00:00 | 1    | 2    |
    */
    ```

5. Run `dbt snapshot -s test_snapshot` again to reproduce the error
> Also noticed that, in contrast, adding a non-numeric column does not produce this error

### Relevant log output

```shell
11:50:25  Completed with 1 error, 0 partial successes, and 0 warnings:
11:50:25  
11:50:25  Failure in snapshot test_snapshot (snapshots/_test__snapshot.yml)
11:50:25    Database Error in snapshot test_snapshot (snapshots/_test__snapshot.yml)
  001003 (42000): SQL compilation error:
  syntax error line 181 at position 25 unexpected 'FIXED'.
```

The compiled SQL in logs/dbt.log:

```shell
    deletion_records as (

        select
            'insert' as dbt_change_type,
            snapshotted_data."PK",snapshotted_data."UPDATED_AT",snapshotted_data."COL1",
    
        try_cast(NULL as FIXED)
     as "COL2",snapshotted_data.dbt_unique_key as dbt_unique_key,
```

### Environment

```markdown
- OS: macOS Tahoe (Version 26.5)
- Python: 3.12.13
- dbt-adapters: 1.24.2
- dbt-snowflake: 1.11.5
```

### Additional Context

Currently `SnowflakeAdapter` inherits `get_column_schema_from_query()` from `BaseAdapter`. Within that, [data_type_code_to_name()](https://github.com/dbt-labs/dbt-adapters/blob/63140d8b177fdcba82a8b5292276fe5eff4c763f/dbt-snowflake/src/dbt/adapters/snowflake/connections.py#L733) maps `type_code` to a string data type name from [FIELD_ID_TO_NAME](https://github.com/snowflakedb/snowflake-connector-python/blob/9c7f5ca4c9d25f9272e7d7dc87eb20aa7408fed8/src/snowflake/connector/constants.py#L199-L211), which is populated by [FIELD_TYPES](https://github.com/snowflakedb/snowflake-connector-python/blob/9c7f5ca4c9d25f9272e7d7dc87eb20aa7408fed8/src/snowflake/connector/constants.py#L128-L130) (where `FIXED` is assigned).

itsnamangoyal:
## Summary

Snowflake's `DESCRIBE`-based cursor introspection reports fixed-point numeric columns as `"FIXED"`, which is not valid Snowflake DDL. This breaks snapshots using `hard_deletes: new_record` when a new numeric column is added to the source relation, since the `deletion_records` CTE emits `try_cast(NULL as FIXED)`.

Fixes this at the `snowflake__safe_cast` macro — the exact call site that emits invalid DDL — rather than in `get_column_schema_from_query()`, which is shared by three other call sites (contract equivalence checks, Iceberg/Glue CLD table creation, snapshot timestamp lookup) that already have their own handling of the `FIXED` string and would regress if that method's return contract changed.

Closes #1986. Supersedes #1988, which fixed `get_column_schema_from_query()` directly — I verified that approach breaks 15 constraint-equivalence functional tests (`test_constraints.py`), since `assert_columns_equivalent` compares two independent calls to that method and previously relied on both sides coincidentally returning the same `"FIXED"` string. Changing that method's output broke the coincidence without fixing the underlying comparison logic.

Thanks to @rio-not-real for reporting #1986, root-causing it down to `snowflake-connector-python`'s `FIELD_ID_TO_NAME` mapping, and opening #1988 with the first fix attempt — this PR builds directly on that investigation, just relocates the fix to avoid the regression above.

## Test plan
- [x] Full `dbt-snowflake` unit test suite (265 tests) — passing
- [x] Full `dbt-snowflake` functional test suite against live Snowflake (469 tests) — no failures attributable to this change; remaining failures are pre-existing environment/credential gaps (missing Iceberg external volume, missing role grants) present on `main` too
- [x] `test_constraints.py` (15 tests that failed under #1988's approach) — all passing with this fix
- [x] `test_ephemeral_snapshot_hard_deletes.py` and `test_simple_snapshot.py` — passing
- [x] Live end-to-end repro in a real dbt project: reproduced #1986 exactly on unpatched `main`, confirmed fixed with this patch, confirmed contract enforcement on a numeric column still passes
- [x] `black`/`mypy` clean against CI's exact pinned versions

### DIFF
--- a/dbt-snowflake/.changes/unreleased/Fixes-20260710-120000.yaml
+++ b/dbt-snowflake/.changes/unreleased/Fixes-20260710-120000.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix invalid 'FIXED' dtype in snapshot deletion_records cast to NUMBER
+time: 2026-07-10T12:00:00.000000-04:00
+custom:
+    Author: itsnamangoyal
+    Issue: "1986"
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/utils/safe_cast.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/utils/safe_cast.sql
@@ -3,6 +3,11 @@
         try_to_geometry({{field}})
     {% elif type|upper == "GEOGRAPHY" -%}
         try_to_geography({{field}})
+    {% elif type|upper == "FIXED" -%}
+        {#-- 'FIXED' is snowflake-connector-python's internal name for fixed-point
+        numeric types (surfaced by get_column_schema_from_query's cursor
+        introspection) -- it is not valid Snowflake DDL, so cast to NUMBER instead --#}
+        try_cast({{field}} as NUMBER)
     {% elif type|upper != "VARIANT" -%}
         {#-- Snowflake try_cast does not support casting to variant, and expects the field as a string --#}
         {% set field_as_string =  dbt.string_literal(field) if field is number else field %}


tauhid621 (commented):
Looks good. let's add some integration tests for this?

itsnamangoyal:
Closing this PR because #1781 fixes this with a better approach.

PR #2064, opened 2026-07-10
fix(redshift): restore transaction_open after execute() retry reconnect

itsnamangoyal:
## Summary
- Root-caused via issue history (#626, #1698) and code tracing: `execute()`'s retry mechanism (`_handle_execute_exception`) is the only place in dbt-redshift that closes and reopens a connection mid-query on a retryable error (`conflict with concurrent transaction`, `could not open relation with OID`). It never restored `connection.transaction_open` or re-issued `BEGIN` on the new connection, desyncing dbt's bookkeeping from the real connection state. Depending on where in a materialization the retry landed, this produced either a hard crash (`Tried to commit transaction... but it does not have one open!`, #1698) or silent loss of earlier-in-transaction work (CORE-724).
- Fix: after the retry's `close()`/`open()`, if a transaction was open before the reconnect, re-issue `begin()` so the flag and the real connection state stay consistent. This is a generic fix at the single choke point, so it covers every multi-statement in-transaction sequence (table/view rename swaps, grants, incremental/snapshot merges), not just one macro.
- Removed now-dead code: `fresh_transaction()`/`fresh_transaction_without_lock()` had their own local flag-restoration (added in #1841) for this exact problem, scoped only to `drop_relation`. Since the fix above is the only place that can desync the flag, and it now restores it itself, that local patch is provably unreachable — removed along with its now-invalid regression test.
- This PR replaces its earlier contents (a full statement-log/replay approach) that reviewers flagged correctness gaps in and that never actually fixed the `transaction_open` desync. Complements #2049, which wraps `rename_relation` in `fresh_transaction()` for the CORE-724 data-loss case specifically — that PR needs no changes and stays independent.

## Test plan
- [x] New unit tests in `test_connection.py`: transaction restored after retry when one was open, not restored when none was open
- [x] Full `dbt-redshift` unit suite passes (169/169)
- [x] `hatch run code-quality` (black, flake8, mypy, pre-commit) passes
- [x] Blast-radius review: confirmed `begin()` can't raise post-`close()`, acts on the same reopened connection, is a no-op-but-consistent in autocommit/skip-transactions mode, doesn't double-BEGIN/double-commit when composed with `fresh_transaction()`, and can't deadlock (the shared lock is a re-entrant `RLock`)
- [x] Local live-cluster verification of the underlying mechanism (not included here — requires live credentials, not CI-runnable)


dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -556,6 +590,7 @@ def begin(self) -> None:
             super().begin()
 
         connection.transaction_open = True
+        self._clear_statement_log()
```
tauhid621: This clears the log when we are replaying. We may want to maintain a state where we do not clear log while replyaing

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -717,6 +754,9 @@ def _handle_execute_exception(
             # otherwise no queries will succeed on this connection
             self.close(self.get_thread_connection())
             self.open(self.get_thread_connection())
+            # the reopened connection lost whatever this transaction had done
+            # before the failing statement; rebuild that state before retrying it
+            self._replay_statement_log()
```
tauhid621: We need to honor `_should_skip_transaction_statements`. In that scenario, we do not need to replay

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -541,6 +546,35 @@ def _is_autocommit_enabled(self) -> bool:
...
+        Python won't redo that decision, and IDENTITY/sequence values are not
+        transactional in Redshift, so replayed inserts using them may not
+        get the same values as the first, rolled-back attempt.
+        """
+        log = self._statement_log.get(threading.get_ident(), [])
+        if not log:
+            return
+        logger.debug(f"Replaying {len(log)} statement(s) after connection reset")
+        for sql, auto_begin in log:
+            self.add_query(sql, auto_begin)
```
tauhid621: We should use same roboustness while running the queries again. any failure in this should be retried

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -541,6 +546,35 @@ def _is_autocommit_enabled(self) -> bool:
         connection = self.get_thread_connection()
         return connection.credentials.autocommit is True
 
+    def _clear_statement_log(self) -> None:
+        self._statement_log.pop(threading.get_ident(), None)
```
tauhid621: `threading.get_ident()` can be used by other threads if current thread dies and a new one has same ident

dbt-redshift/tests/unit/test_connection.py
```diff
@@ -701,3 +702,121 @@ def test_raises_after_retries_exhausted(self):
             "could not complete because of conflict with concurrent transaction",
             always_fail=True,
         )
+
+
+class TestStatementReplayOnRetry(TestCase):
```
tauhid621: We should add:

- a two-conflicts-in-one-transaction test
- an autocommit + skip-transactions test asserting no replay happens;
- a test where a replayed statement itself fails.

tauhid621 (commented):
Left comments. This is a very high risk PR. We should ensure that we have all areas covered and all areas tested manually.

dbt-redshift/src/dbt/adapters/redshift/connections.py
```diff
@@ -715,8 +707,15 @@ def _handle_execute_exception(
             retries -= 1
             # we need to actually close and open to get a new connection
             # otherwise no queries will succeed on this connection
-            self.close(self.get_thread_connection())
-            self.open(self.get_thread_connection())
+            connection = self.get_thread_connection()
+            was_transaction_open = connection.transaction_open
+            self.close(connection)
+            self.open(connection)
+            if was_transaction_open:
```
tauhid621, line 714: This is just changing how we handle open transactions. The PR is not changing any behaviour or fixing the issue it was initially solving.

tauhid621 (changes requested):
The PR now is just a refactor and does not solve any scenario around temp tables not available

PR #2077, opened 2026-07-16
fix(snowflake): detect query body changes in dynamic table config changeset

troyharvey:
## Summary

`SnowflakeDynamicTableConfigChangeset.has_changes` checks `target_lag`, `snowflake_warehouse`, `refresh_mode`, etc. but **not the query body**. When only the SQL of a dynamic table model changes (no config param diff), dbt logs `No configuration changes were identified` and skips the DDL — the Snowflake dynamic table silently keeps its old query.

- Adds `SnowflakeDynamicTableQueryConfigChange` with `requires_full_refresh = True` (query changes require `CREATE OR REPLACE`, not `ALTER`)
- Adds `query` field to `SnowflakeDynamicTableConfigChangeset`
- Wires up the comparison in `dynamic_table_config_changeset` in `relation.py`
- Exports the new class from `relation_configs/__init__.py`

## Reproduction

1. Create a dbt dynamic table model and run it
2. Change only the SQL body (e.g. add a row to an allowlist)
3. Run `dbt run --select <model>` — observe `skip` with "No configuration changes were identified"
4. The Snowflake dynamic table still has the old query

## Test plan

- [ ] Unit test: changeset detects a query diff and sets `query` change with `requires_full_refresh=True`
- [ ] Unit test: changeset returns `None` when query is unchanged
- [ ] Integration: `dbt run` on a model with a SQL-only change triggers `CREATE OR REPLACE DYNAMIC TABLE`

### DIFF
--- a/dbt-snowflake/src/dbt/adapters/snowflake/relation.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/relation.py
@@ -28,6 +28,7 @@ from dbt.adapters.snowflake.relation_configs import (
     SnowflakeDynamicTableImmutableWhereConfigChange,
     SnowflakeDynamicTableClusterByConfigChange,
     SnowflakeDynamicTableTransientConfigChange,
+    SnowflakeDynamicTableQueryConfigChange,
     SnowflakeQuotePolicy,
     SnowflakeRelationType,
 )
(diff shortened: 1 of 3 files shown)


troyharvey:
@cla-bot signed

PR #2092, opened 2026-07-22
fix(snowflake): make Iceberg timestamp-precision error (091385) actionable

### LINKED ISSUE #2090: [Bug] Iceberg timestamp-precision error (091385) gives no guidance on how to fix it
## Is this a new bug?

- [x] dbt-snowflake

## Current Behavior

When you create an Iceberg table that includes a `TIMESTAMP_NTZ(9)` column, Snowflake rejects the `CREATE` with error 091385:

```
Database Error in model my_model (models/my_model.sql)
  091385 (42601): Invalid time type scale specified for column 'TS'
  with data type 'TIMESTAMP_NTZ(9)'.
```

The error message gives no indication of what the valid range is or how to fix it.

## Expected Behavior

The error should explain that Iceberg (format version 2) only supports timestamps up to microsecond precision (scale 6), and tell the user what they can do: cast to `TIMESTAMP_NTZ(6)` / `TIMESTAMP_LTZ(6)`, or use a format-version 3 table with nanosecond support enabled.

## Steps To Reproduce

1. Configure a Snowflake Iceberg model (BUILT_IN or ICEBERG_REST catalog)
2. Write a model that selects a `TIMESTAMP_NTZ(9)` expression
3. Run `dbt run`
4. Observe the raw 091385 with no actionable guidance

## Relevant log output

```
Database Error in model my_model (models/my_model.sql)
  091385 (42601): Invalid time type scale specified for column 'TS'
  with data type 'TIMESTAMP_NTZ(9)'.
```

## Environment

- OS:
- Python:
- dbt-adapters: 1.10.x
- dbt-snowflake: 1.11.0 (latest stable; likely affects any version with Iceberg support)
- Snowflake Iceberg table (BUILT_IN or ICEBERG_REST catalog)

sfc-gh-nijain:
resolves #2090
docs: N/A

### Problem

When a dbt model targets an Iceberg table and a column uses `TIMESTAMP_NTZ(9)`, Snowflake returns error 091385 whose message says "Please consult the Snowflake documentation" but gives no actionable guidance on what to change or why.

### Solution

Iceberg v2 doesn't support nano-second precision (ntz(9)). We started by looking into getting the precision from 9 to 6 either internally within snowflake or in the dbt-adapter itself as the initial fix. But that would mean silently dropping the precision on customer data (either within the adapter or within Snowflake) without customer having any indication of this happening. Snowflake rejecting the DDL with 091385 is actually the correct behavior since it protects the csutomer from silent data loss.
That's where we pivoted to an approach of making the error message more actionable to the customer instead of silent transformations.

In `SnowflakeConnectionManager.exception_handler`, detect when the raised `ProgrammingError` contains `"091385"` and append an actionable hint:

> Iceberg supports nanosecond timestamps only on format-version 3 with nanosecond timestamp support enabled. Either cast the column to TIMESTAMP_NTZ(6) / TIMESTAMP_LTZ(6), or use an Iceberg v3 table with nanosecond timestamp support enabled.

The hint fires reactively — only when Snowflake rejects the statement — so it is never shown where nanosecond timestamps are actually supported (Iceberg v3 with the feature enabled).

### Checklist

### DIFF
--- a/dbt-snowflake/.changes/unreleased/Fixes-20260709-120000.yaml
+++ b/dbt-snowflake/.changes/unreleased/Fixes-20260709-120000.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: Append an actionable hint to the Iceberg timestamp-precision error (091385), pointing
+  users to cast to TIMESTAMP_NTZ(6)/TIMESTAMP_LTZ(6) or use an Iceberg v3 table with nanosecond
+  support enabled
+time: 2026-07-09T12:00:00.000000-07:00
+custom:
+  Author: sfc-gh-nijain
+  Issue: "2090"
--- a/dbt-snowflake/src/dbt/adapters/snowflake/connections.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/connections.py
@@ -401,6 +401,20 @@ class SnowflakeConnectionManager(SQLConnectionManager):
             for regex_pattern, replacement_message in ERROR_REDACTION_PATTERNS.items():
                 msg = re.sub(regex_pattern, replacement_message, msg)
 
+            # Iceberg tables reject timestamp precision greater than microseconds (6) unless the
+            # table is Iceberg format-version 3 with nanosecond timestamp support enabled. The
+            # raw error (091385) does not explain how to resolve it, so append an actionable hint.
+            # This is reactive: it fires only when Snowflake actually rejects the statement, so it
+            # never triggers where nanosecond timestamps are supported (v3 with the feature on).
+            if "091385" in msg:
+                msg = (
+                    f"{msg}\n\n"
+                    "Iceberg supports nanosecond timestamps only on format-version 3 with "
+                    "nanosecond timestamp support enabled. Either cast the column to "
+                    "TIMESTAMP_NTZ(6) / TIMESTAMP_LTZ(6), or use an Iceberg v3 table with "
+                    "nanosecond timestamp support enabled."
+                )
+
             logger.debug("Snowflake query id: {}".format(e.sfqid))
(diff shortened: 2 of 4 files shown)


sfc-gh-nijain:
CLA is signed. Could a maintainer add the `ci:approve-public-fork-ci` label to unblock CI? Thanks!

colin-k-rogers:
This should really be fixed in Snowflake itself, this fix will be brittle and is likely to result in dead code that we'll need to clean up down the line.

sfc-gh-nijain:
Fair point on the long-term concern. A few reasons I think this still makes sense here:

* **Precedent in the same file.** The existing `exception_handler` already does this for "This session does not have a current database", it appends a note about invalid credentials and role access. This PR follows that established pattern exactly.

* **Stability of the check.** The code keys on error code `"091385"`, not on message text, so it is not tied to any particular wording.

* **Low cleanup cost if the upstream behavior changes.** If nanosecond timestamp support becomes more broadly available and 091385 stops firing in practice, the code becomes dead but harmless, and removal is a small, clearly-motivated delete. Happy to add an inline comment flagging exactly that to make it easy to find later.

* **Bridging the gap for users today.** Iceberg v3 with nanosecond support is not yet the common case. In the meantime, surfacing the path forward (cast to `TIMESTAMP_NTZ(6)` or use v3 with the feature enabled) at the adapter layer helps users resolve the issue without having to dig through documentation.

colin-k-rogers:
Those precedents cover a much more generic case that's never going to be patched in Snowflake this is a hyper specific case that definitely _should_ be addressed there and opens the door to a never ending amount of exception handling. 

Also, it does not inspire confidence you've given this much thought when you reply with a clearly AI generated response.

PR #2093, opened 2026-07-22
fix(snowflake): raise early dbt error for bare NUMBER in Iceberg contracts (099200)

### LINKED ISSUE #2091: [Bug] Model contract with bare NUMBER/NUMERIC/DECIMAL on an Iceberg table fails with opaque Snowflake error 099200
## Is this a new bug?

- [x] dbt-snowflake

## Current Behavior

If a model contract declares a column with a bare `number`, `numeric`, or `decimal` type (no precision or scale) and the model is an Iceberg table, `dbt run` fails deep into execution with a Snowflake error:

```
Database Error in model my_model (models/my_model.sql)
  099200 (P0000): Iceberg does not support the type for column 'ID': numeric
```

Regular Snowflake (non-Iceberg) tables silently coerce bare `NUMBER` to `NUMBER(38,0)`, so this error only surfaces when switching to Iceberg. The raw Snowflake error does not tell you that the fix is simply to add explicit precision and scale in the contract.

## Expected Behavior

dbt should fail fast with a clear, actionable error before the `CREATE` statement is sent to Snowflake. The message should name the offending column(s) and tell the user to add explicit precision and scale (e.g., `number(38, 0)`).

## Steps To Reproduce

1. Configure a Snowflake Iceberg model (BUILT_IN or ICEBERG_REST catalog) with `contract: {enforced: true}`
2. In `schema.yml`, declare a column with `data_type: number` (no precision/scale)
3. Run `dbt run`
4. Observe the opaque 099200 from Snowflake

**schema.yml:**
```yaml
models:
  - name: my_model
    config:
      contract:
        enforced: true
    columns:
      - name: id
        data_type: number
```

## Relevant log output

```
Database Error in model my_model (models/my_model.sql)
  099200 (P0000): Iceberg does not support the type for column 'ID': numeric
```

## Environment

- OS:
- Python:
- dbt-adapters: 1.10.x
- dbt-snowflake: 1.11.0 (latest stable; likely affects any version with Iceberg support)
- Snowflake Iceberg table (BUILT_IN or ICEBERG_REST catalog, format version 2 or 3)

sfc-gh-nijain:
resolves #2091
docs: N/A

### Problem

When a dbt model targets an Iceberg table with a contract that declares a numeric column as bare `NUMBER`, `NUMERIC`, or `DECIMAL` (no precision or scale), Snowflake rejects the generated DDL at runtime with opaque error 099200 that does not name the offending column or explain the fix. Regular Snowflake tables silently coerce bare `NUMBER` to `NUMBER(38,0)`, so users migrating from non-Iceberg to Iceberg tables encounter this unexpectedly.

### Solution

Override `snowflake__get_table_columns_and_constraints` to inspect the model contract before the `CREATE` reaches Snowflake. If the target is an Iceberg relation (BUILT_IN or ICEBERG_REST) and any contract column declares a bare numeric type, raise a clear dbt compiler error that names the column(s) and directs the user to set explicit precision and scale (e.g. `number(38, 0)`).

The user's declared type is not rewritten — we ask them to be explicit. Non-Iceberg contracts are unaffected.

### Checklist

### DIFF
--- a/dbt-snowflake/.changes/unreleased/Fixes-20260709-123000.yaml
+++ b/dbt-snowflake/.changes/unreleased/Fixes-20260709-123000.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: Raise a clear dbt error, before the CREATE reaches Snowflake, when a model contract
+  declares a bare NUMBER/NUMERIC/DECIMAL (no precision/scale) on an Iceberg table, instead of
+  the opaque 099200 Snowflake returns at runtime. The user's declared type is not rewritten
+time: 2026-07-09T12:30:00.000000-07:00
+custom:
+  Author: sfc-gh-nijain
+  Issue: "2091"
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/relations/column/columns_spec_ddl.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/relations/column/columns_spec_ddl.sql
@@ -1,0 +1,38 @@
+{% macro snowflake__get_table_columns_and_constraints() -%}
+  {#-
+    Snowflake + Iceberg: a model contract that declares a numeric column without precision
+    and scale (a bare NUMBER / NUMERIC / DECIMAL) renders DDL that Snowflake rejects on Iceberg
+    tables with error 099200. Regular Snowflake tables silently coerce bare NUMBER to
+    NUMBER(38,0), but Iceberg requires the precision and scale to be explicit.
+
+    Fail fast with a clear dbt error before the CREATE is sent to Snowflake, instead of the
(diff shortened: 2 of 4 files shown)


sfc-gh-nijain:
CLA is signed. Could a maintainer add the `ci:approve-public-fork-ci` label to unblock CI? Thanks!

aahel (commented):
Thnx for investigating this issue.
Snowflake's error message seems to be clear now.
```
099200 (42601): SQL Compilation error:
  data type 'NUMBER/DECIMAL' without precision and scale is not supported for Iceberg tables. 
```
i don't think we should add this snowflake side validation logic in dbt.

 Closing this, but thanks again for raising the issue and pr!

Issue #1991, opened 2026-06-01
sync_all_columns drops all columns when datasharing: true on Redshift


gseda19:
### Is this a new bug?

- [x] I believe this is a new bug
- [x] I have searched the existing issues, and I could not find an existing issue for this bug

### Which packages are affected?

- [x] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [x] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

When a Redshift connection is configured with `datasharing: true`, `information_schema.columns` returns zero rows when queried against a session-local temp table (e.g. <model>__dbt_tmp<session_id>). The query executes successfully with no error — it simply returns no rows.
With on_schema_change='sync_all_columns', dbt interprets this as "source has 0 columns, target has all N columns" and issues `ALTER TABLE ... DROP COLUMN` for every column in the target table. All drops succeed at the Redshift layer. The subsequent DELETE then fails immediately because the unique key columns were just dropped:
`Redshift error: column "<unique_key>" does not exist in <model>`

### Expected Behavior

`sync_all_columns` should correctly resolve the schema of the temp table when `datasharing: true` is configured, and should not drop columns that exist in the source.

### Steps To Reproduce

Steps To Reproduce

1. Configure a Redshift connection with datasharing: true
2. Create an incremental model with on_schema_change='sync_all_columns' and one or more unique_key columns
3. Run the model incrementally 
4. Observe that information_schema.columns returns 0 rows for the temp table, all target columns are dropped, and the DELETE fails

### Relevant log output

```shell
-- Step 1: information_schema.columns on the temp table — succeeds but returns 0 rows
with bound_views as (
  select ordinal_position, table_schema, column_name, data_type, ...
  from information_schema."columns"
  where table_name = '<model>__dbt_tmp<session_id>'
)
...
-- SQL status: SUCCESS (~55 seconds, 0 rows returned)

-- Step 2: sync_all_columns diff
Schema changed: True
Source columns not in target: []
Target columns not in source: [col1, col2, ..., <unique_key>, ...]  -- all N columns

-- Step 3: dbt drops all columns
alter table "<db>"."<schema>"."<model>" drop column "col1"  -- SUCCESS
alter table "<db>"."<schema>"."<model>" drop column "col2"  -- SUCCESS
[...]

Issue #1947, opened 2026-05-19
[Bug] Connecting to Redshift (datashare) external database breaks on_schema_change


zachary-povey:
### Is this a new bug?

- [x] I believe this is a new bug
- [x] I have searched the existing issues, and I could not find an existing issue for this bug

### Which packages are affected?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [x] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

If in the profiles.yml if you set up a redshift connection using an "external" (datasharing) database for `dbname` the on_schema_change functionality does not work:
- `on_schema_change=fail` fails, even if there are no changes
- `on_schema_change=append_new_columns` ignores new columns
- `on_schema_change=ignore` sidesteps the issue

### Expected Behavior

I would expect either on_schema_change to work as per the docs, or a helpful error to be raised explaining the combined set of options (external db, on_schema_change != ignore) is not supported.

### Steps To Reproduce

1. Create two redshift serverless namespace + workgroup pairs
2. Set up a producer -> consumer datashare between the two, enabling the consumer to create/modify tables in the producer using [multi-warehouse writes](https://docs.aws.amazon.com/redshift/latest/dg/getting-started-datashare-writes.html)
3. Connect to the *consumer* workgroup in your dbt profiles.yml, with dbname set to the external datashare db
4. Attempt to process incremental changes for a model with on_schema_change != ignore

### Relevant log output

```shell
11:40:33  2 of 2 ERROR creating sql incremental model public.test_latest ................. [ERROR in 2.25s]
11:40:34  
11:40:34  Finished running 1 incremental model, 1 table model in 0 hours 0 minutes and 11.24 seconds (11.24s).
11:40:34  
11:40:34  Completed with 1 error, 0 partial successes, and 0 warnings:
11:40:34  
11:40:34  Failure in model test_latest (models/public/test_latest.sql)
11:40:34    Compilation Error in model test_latest (models/public/test_latest.sql)
  
                The source and target schemas on this incremental model are out of sync!
                They can be reconciled in several ways:
                  - set the `on_schema_change` config to either append_new_columns or sync_all_columns, depending on your situation.
                  - Re-run the incremental model with `full_refresh: True` to update the target schema.
                  - update the schema manually and re-run the process.
[...]

PR #2096, opened 2026-07-28
fix: raise instead of dropping all columns when temp relation introspection returns empty

will-sargent-dbtlabs:
## Problem

`default__check_for_schema_changes` accepts an empty `source_columns` list as a legitimate schema:

```jinja
{%- set source_columns = adapter.get_columns_in_relation(source_relation) -%}   {# -> [] #}
{%- set target_columns = adapter.get_columns_in_relation(target_relation) -%}   {# -> N columns #}
{%- set target_not_in_source = diff_columns(target_columns, source_columns) -%} {# -> all N #}
```

Every caller of `process_schema_changes` passes a freshly-built temp relation as the source
(`incremental.sql`, and the snowflake / spark / athena materializations), so that relation always has
columns. An empty result is therefore always an introspection failure — and continuing makes
`sync_all_columns` conclude that every target column was removed and emit
`ALTER TABLE … DROP COLUMN` for all of them.

This is reachable in production, not theoretical. On Redshift, connecting with `dbname` set to a
datashare consumer database (`svv_redshift_databases.database_type = 'shared'`) leaves temporary
relations **invisible to `information_schema.columns`, `pg_attribute` and `svv_columns`** while the
relation itself stays fully queryable — so `redshift__get_columns_in_relation_legacy` returns zero
rows with statement status `FINISHED` and no error. Reported in #1947 and #1991, with a reproduction
and evidence tables in [this comment](https://github.com/dbt-labs/dbt-adapters/issues/1947#issuecomment-5108260475).

The user-visible symptom is confusing, because the failure surfaces one step later as
`column "<unique_key>" does not exist` from the incremental `DELETE`, which sends people looking at
their unique key rather than at metadata.

### What the unguarded path actually does

Traced against a live Postgres with introspection stubbed to return `[]`, `on_schema_change='sync_all_columns'`:

```
Schema changed: True
Target columns not in source: [id, field_1, field_2]
Columns removed: [id, field_1, field_2]
alter table "…"."incremental_preguard" drop column "id", drop column "field_1", drop column "field_2"
SQL status: ALTER TABLE in 0.001 seconds
… then: syntax error at or near ")"  ->  insert into "…"."incremental_preguard" ()
```

Worth recording precisely, because it corrects an assumption I had made in the first version of this
PR. The `DROP COLUMN` **is** issued and succeeds; the run then dies on the empty insert column list.
Postgres has transactional DDL, so that failed transaction rolls the drop back and the columns
survive **on Postgres**. On a warehouse that commits the DDL the columns stay gone — which is what
#1947 and #1991 report. This has a direct consequence for testing: comparing column lists before and
after cannot distinguish a guarded run from an unguarded one on Postgres, so the test asserts that
**no `drop column` is issued at all** instead.

## Fix

Guard in `default__process_schema_changes`, immediately after `check_for_schema_changes` returns.
Because `on_schema_change` is in scope there, the guard can be scoped to what is actually
destructive rather than failing every path:

| `on_schema_change` | unguarded behaviour | guarded behaviour |
|---|---|---|
| `sync_all_columns` | issues `drop column` for every target column, then fails on `insert into t ()` | **raises** |
| `fail` | raises the generic "out of sync" error listing every target column as removed | **raises**, naming the metadata failure |
| `append_new_columns` | no-op `ALTER`, returns `[]`, caller falls back to the target's columns — run passes | **warns**, run still passes |
| `ignore` | returns before the guard | unchanged |

The error names both relations and the active `on_schema_change` value, states that this is a
metadata failure rather than an empty schema, and lists causes and workarounds.

**`append_new_columns` deliberately warns rather than raises.** Every incremental materialization in
this repo (core, snowflake, athena, bigquery) has this immediately after the call:

```jinja
{% set dest_columns = process_schema_changes(...) %}
{% if not dest_columns %}
  {% set dest_columns = adapter.get_columns_in_relation(existing_relation) %}
{% endif %}
```

so an empty return is already absorbed and the run completes correctly using the target's columns.
Raising there would convert currently-passing runs into hard failures for no safety benefit — nothing
can be dropped under `append_new_columns`. The real cost is that a genuinely new column goes
undetected, which is what the warning says. `--warn-error` escalates it for anyone who wants the
failure.

Two honest caveats on that choice:

- `exceptions.warn` fires with `force_warn_or_error_handling=True`, so **`--warn-error` turns this
  into an error**. That is the desired escape hatch, but it does mean strict CI setups will see a
  new failure. Covered by a test.
- The warning appears in the log body but does **not** increment the run summary's `WARN` count —
  a warned run still reports `PASS=1 WARN=0`. That is how `exceptions.warn` behaves generally, not
  something this PR introduces, but it makes the warning easier to miss than the wording implies.

## Scope

- **No behaviour change on healthy runs** — the guard only fires when introspection returns nothing.
- **`default__check_for_schema_changes` is byte-identical to `main`.** The guard originally lived
  there; it moved per review. The net diff to the macro file is +41 lines, 0 deletions.
- **`on_schema_change='ignore'` is unaffected** — it returns from `process_schema_changes` before
  reaching the guard. Covered by a test.
- **BigQuery is untouched** — it defines both `bigquery__check_for_schema_changes` **and**
  `bigquery__process_schema_changes`, and the latter dispatches to the former, so it never reached
  the default guard before this change and still does not. Adapter coverage is therefore identical
  in both placements. It has the same unguarded shape and may want the same treatment, but I've left
  it out of scope rather than change an adapter I can't test here.
- Deliberately **not** guarding empty `target_columns`: that direction is additive (columns get added,
  not dropped), so it isn't destructive and the correct behaviour is less obvious.

The underlying Redshift introspection gap is now fixed on `main` by #2097, which describes temp
relations from the driver's cursor when the catalog cannot see them. So the condition this guard
catches is no longer reachable on Redshift through the reported path, and this PR is now
defense-in-depth: it keeps a metadata failure in **any** adapter from turning into a dropped schema,
rather than being the fix for a live outage.

## Tests

`dbt-postgres/tests/functional/incremental_schema_tests/test_empty_source_columns.py` reproduces the
condition adapter-agnostically by overriding `postgres__get_columns_in_relation` to return `[]` for
`*__dbt_tmp*` relations only. One class per `on_schema_change` value:

| test | asserts |
|---|---|
| `TestEmptySourceColumnsRaises` | raises, **and no `drop column` is issued** (see above for why that, not a column comparison, is the load-bearing check) |
| `TestEmptySourceColumnsFailRaises` | raises the metadata error and **not** the generic `out of sync` one |
| `TestEmptySourceColumnsAppendNewColumnsWarns` | run **passes**, warning is emitted, and the new column is not appended — the accepted limitation |
| `TestEmptySourceColumnsAppendNewColumnsWarnError` | `--warn-error` escalates the warning to a failure |
| `TestEmptySourceColumnsIgnoreStillWorks` | `ignore` is unaffected |

Verified against a local Postgres 16:

- **5/5 pass** with the guard.
- **4/5 fail** with the macro reverted to `main`. The `ignore` test passes either way, which is its
  purpose — it exists to prove the guard is *not* reached on that path.
- Full `dbt-postgres` functional suite: **570 passed, 12 skipped**, plus 2 errors in
  `test_simple_copy.py::TestSimpleCopyUppercase` and `test_mixed_case_db.py` that are
  `FailedToConnectError` against the mixed-case database in my local container and **reproduce
  identically with this change reverted** — environmental, not caused by this PR.

## Review response

@tauhid621 — both of your comments are addressed by the same move, and you were right that
`default__check_for_schema_changes` was the wrong home: it's an inspect-and-return macro, and raising
from it changed the contract for anyone calling it for introspection. The guard now sits beside the
existing `on_schema_change` error in `default__process_schema_changes`.

I checked whether moving it costs coverage, since either macro could be overridden — it doesn't;
BigQuery overrides both and is the only override in the repo, so the two placements protect exactly
the same set of adapters.

Your second point about `on_schema_change` being available is what led to the per-config table above.
Scoping by it is what makes it possible to keep `append_new_columns` runs working instead of failing
them, which turned out to matter more than I expected once I traced the `dest_columns` fallback.

## Checklist

- [x] Changelog entry added (`Fixes-20260728-114754.yaml`, updated to mention the `append_new_columns` warning)
- [x] `black` clean at line-length 99 (flake8's `exclude` covers `dbt-postgres/tests` and `dbt-redshift/tests`, so it does not lint these files)
- [x] Regression tests verified to fail without the fix
- [x] Rebuilt on top of `main` post-#2097

### DIFF
--- a/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
+++ b/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
@@ -27,6 +27,36 @@
 
   {%- set source_columns = adapter.get_columns_in_relation(source_relation) -%}
   {%- set target_columns = adapter.get_columns_in_relation(target_relation) -%}
+
+  {#-
+    An empty source column list is a metadata failure, not a real schema. This materialization
+    just built and populated `source_relation`, so it necessarily has columns. If column
+    introspection comes back empty it means the adapter could not see the relation -- and
+    continuing would compute "every target column is missing from the source", which under
+    on_schema_change='sync_all_columns' drops every column in the target table.
+    Fail loudly instead of silently destroying the target schema.
+  -#}
+  {%- if source_columns | length == 0 -%}
+    {%- set empty_source_columns_msg -%}
+      Could not read any columns for {{ source_relation }} while checking for schema changes on
+      incremental model {{ target_relation }}.
+
+      This is a metadata failure rather than an empty schema: {{ source_relation }} was just built
+      by this run, so it does have columns. Continuing would treat every column in
+      {{ target_relation }} as removed and, with on_schema_change='sync_all_columns', drop them.
+
+      Common causes:
+        - the warehouse does not expose this intermediate/temporary relation to the catalog views
+          the adapter reads column metadata from
+        - the connection's database differs from the one the temporary relation was created in
+
+      Workarounds:
+        - set on_schema_change='ignore' to skip this check
+        - run the model with --full-refresh to rebuild it
+    {%- endset -%}
+    {%- do exceptions.raise_compiler_error(empty_source_columns_msg) -%}
+  {%- endif -%}
+
   {%- set source_not_in_target = diff_columns(source_columns, target_columns) -%}
   {%- set target_not_in_source = diff_columns(target_columns, source_columns) -%}
 
--- a/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
+++ b/dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
@@ -27,36 +27,6 @@
 
   {%- set source_columns = adapter.get_columns_in_relation(source_relation) -%}
   {%- set target_columns = adapter.get_columns_in_relation(target_relation) -%}
-
-  {#-
-    An empty source column list is a metadata failure, not a real schema. This materialization
-    just built and populated `source_relation`, so it necessarily has columns. If column
-    introspection comes back empty it means the adapter could not see the relation -- and
-    continuing would compute "every target column is missing from the source", which under
-    on_schema_change='sync_all_columns' drops every column in the target table.
-    Fail loudly instead of silently destroying the target schema.
-  -#}
-  {%- if source_columns | length == 0 -%}
-    {%- set empty_source_columns_msg -%}
-      Could not read any columns for {{ source_relation }} while checking for schema changes on
-      incremental model {{ target_relation }}.
-
-      This is a metadata failure rather than an empty schema: {{ source_relation }} was just built
-      by this run, so it does have columns. Continuing would treat every column in
-      {{ target_relation }} as removed and, with on_schema_change='sync_all_columns', drop them.
-
-      Common causes:
-        - the warehouse does not expose this intermediate/temporary relation to the catalog views
-          the adapter reads column metadata from
-        - the connection's database differs from the one the temporary relation was created in
-
-      Workarounds:
-        - set on_schema_change='ignore' to skip this check
-        - run the model with --full-refresh to rebuild it
-    {%- endset -%}
-    {%- do exceptions.raise_compiler_error(empty_source_columns_msg) -%}
-  {%- endif -%}
-
   {%- set source_not_in_target = diff_columns(source_columns, target_columns) -%}
   {%- set target_not_in_source = diff_columns(target_columns, source_columns) -%}
 
@@ -153,6 +123,47 @@
     {% else %}
 
       {% set schema_changes_dict = check_for_schema_changes(source_relation, target_relation) %}
+
+      {#-
+        An empty source column list is an introspection failure, not a real schema -- this run just
+        built source_relation. sync_all_columns would drop every target column and fail would report
+        a bogus diff, so both refuse. append_new_columns never drops, and the caller substitutes the
+        target's columns for our empty return, so it warns instead of breaking a working run.
+      -#}
+      {%- if schema_changes_dict['source_columns'] | length == 0 -%}
+        {%- set empty_source_columns_msg -%}
+          Could not read any columns for {{ source_relation }} while checking for schema changes on
+          incremental model {{ target_relation }} (on_schema_change='{{ on_schema_change }}').
+
+          This is a metadata failure rather than an empty schema: {{ source_relation }} was just
+          built by this run, so it does have columns.
+
+          Common causes:
+            - the warehouse does not expose this temporary relation to the catalog views the
+              adapter reads column metadata from
+            - the connection's database differs from the one the temporary relation was created in
+
+          {% if on_schema_change == 'append_new_columns' -%}
+          Continuing, because on_schema_change='append_new_columns' never drops columns and dbt
+          will insert using {{ target_relation }}'s existing columns. No new column can be
+          detected this run, so if the model added one it stays missing until introspection works
+          or you run with --full-refresh.
+          {%- else -%}
+          Refusing to continue: dbt cannot determine which columns changed, so it cannot safely
+          apply on_schema_change='{{ on_schema_change }}'. With 'sync_all_columns' this would
+          silently drop every column in {{ target_relation }}.
+
+          Workarounds:
+            - set on_schema_change='ignore' to skip this check
+            - run the model with --full-refresh to rebuild it
+          {%- endif %}
+        {%- endset -%}
+        {%- if on_schema_change == 'append_new_columns' -%}
+          {%- do exceptions.warn(empty_source_columns_msg) -%}
+        {%- else -%}
+          {%- do exceptions.raise_compiler_error(empty_source_columns_msg) -%}
+        {%- endif -%}
+      {%- endif -%}
 
       {% if schema_changes_dict['schema_changed'] %}
 
--- a/dbt-adapters/.changes/unreleased/Fixes-20260728-114754.yaml
+++ b/dbt-adapters/.changes/unreleased/Fixes-20260728-114754.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Raise a clear error instead of dropping all target columns when column introspection
+    of an incremental model's temporary relation returns no columns
+time: 2026-07-28T11:47:54.242865-07:00
+custom:
+    Author: will-sargent-dbtlabs
+    Issue: "1947 1991"
--- a/dbt-adapters/.changes/unreleased/Fixes-20260728-114754.yaml
+++ b/dbt-adapters/.changes/unreleased/Fixes-20260728-114754.yaml
@@ -1,6 +1,7 @@
 kind: Fixes
 body: Raise a clear error instead of dropping all target columns when column introspection
-    of an incremental model's temporary relation returns no columns
+    of an incremental model's temporary relation returns no columns. on_schema_change='append_new_columns'
+    warns and continues instead, since it cannot drop columns
 time: 2026-07-28T11:47:54.242865-07:00
 custom:
     Author: will-sargent-dbtlabs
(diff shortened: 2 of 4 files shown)


dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
```diff
@@ -27,6 +27,36 @@
 
   {%- set source_columns = adapter.get_columns_in_relation(source_relation) -%}
   {%- set target_columns = adapter.get_columns_in_relation(target_relation) -%}
+
+  {#-
```
tauhid621: This is code path for all `on_schema_change`. Maybe we should move this to `default__process_schema_changes` ?

dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
tauhid621, line 126: Better to have the error here. We also have `on_schema_change` variable to identify the behaviour.

tauhid621 (commented):
Changes look good. Left a comment around palcement of the error

dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
will-sargent-dbtlabs: Agreed, and done in ae19ba8a — you were right that this was the wrong macro. `check_for_schema_changes` is an inspect-and-return macro, so raising from it changed the contract for anyone calling it just to introspect. `default__check_for_schema_changes` is now byte-identical to `main` and the guard lives in `default__process_schema_changes`.

One thing I checked before moving it, since either macro could be overridden and a guard only protects adapters that don't override its host: **coverage is identical either way.** BigQuery is the only override in this repo and it defines *both* `bigquery__check_for_schema_changes` and `bigquery__process_schema_changes`, with the latter dispatching to the former — so it never reached the default guard before this change and still doesn't. Nothing lost by moving, and the placement is cleaner.

dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/on_schema_change.sql
will-sargent-dbtlabs: Moved here in ae19ba8a. Your point about `on_schema_change` being available turned out to be the more consequential half of the review — having it in scope let me scope the guard to what is actually destructive instead of failing every non-`ignore` path:

| `on_schema_change` | unguarded | now |
|---|---|---|
| `sync_all_columns` | issues `drop column` for every target column, then fails on `insert into t ()` | **raises** |
| `fail` | raises the generic "out of sync" error listing every target column as removed | **raises**, naming the metadata failure |
| `append_new_columns` | no-op `ALTER`, returns `[]`, caller falls back to the target's columns — run passes | **warns**, run still passes |
| `ignore` | returns before the guard | unchanged |

`append_new_columns` warns rather than raises because every incremental materialization here (core, snowflake, athena, bigquery) does this right after the call:

```jinja
{% set dest_columns = process_schema_changes(...) %}
{% if not dest_columns %}
  {% set dest_columns = adapter.get_columns_in_relation(existing_relation) %}
{% endif %}
```

so the empty return is already absorbed and the run completes correctly. Raising would convert currently-passing runs into hard failures with no safety benefit — nothing can be dropped under `append_new_columns`. The cost is that a genuinely new column goes undetected, which is what the warning says, and `--warn-error` escalates it for anyone who wants the failure.

Two caveats I'd rather flag than leave you to find:

- `exceptions.warn` fires with `force_warn_or_error_handling=True`, so `--warn-error` **does** turn this into an error. Intended as the escape hatch, but strict CI setups will see a new failure. There's a test for it.
- The warning doesn't increment the summary's `WARN` count — a warned run still reports `PASS=1 WARN=0`. That's general `exceptions.warn` behaviour rather than anything this PR introduces, but it makes the warning easier to miss than the wording suggests.

Also worth knowing, since it changed how the test is written: I traced the unguarded path on a live Postgres and the `DROP COLUMN` **is** issued and succeeds — the run then dies on the empty insert list, and Postgres's transactional DDL rolls the drop back. So on Postgres the columns survive either way, and a before/after column comparison can't tell a guarded run from an unguarded one. The test asserts no `drop column` is issued instead. On a warehouse that commits the DDL the columns stay gone, which is what #1947/#1991 report.

will-sargent-dbtlabs:
### CI is red from a GitHub Actions incident, not from this change

Flagging this so the red X isn't read as a broken PR. Run [31116580642](https://github.com/dbt-labs/dbt-adapters/actions/runs/31116580642) on `ae19ba8a` finished **25 pass / 12 fail**. I checked all 12 — none failed in a test assertion, build step, or lint check:

| Failure mode | Count | Jobs |
|---|---|---|
| Died in `Set up job`, never reached checkout | 5 | redshift, bigquery (1), spark `apache_spark` / `spark_http_odbc`, unit-tests (dbt-postgres) |
| `cancelled` with zero steps executed | 3 | snowflake, unit-tests (dbt-spark), verify-build (dbt-postgres) |
| `cancelled` mid-run | 3 | bigquery (2), `databricks_cluster`, `databricks_sql_endpoint` |
| `alls-green` aggregator reflecting the above | 1 | Pull request checks |

The `Set up job` failures are all the same thing:

```
Getting action download info
Failed to resolve action download info. Error: Service Unavailable
Retrying in 28.338 seconds
Failed to resolve action download info. Error: Internal Server Error
##[error]Failed to resolve action download info.
```

GitHub couldn't serve `actions/checkout` / `actions/setup-python` / `pre-commit/action`. This was the third consecutive round with the same signature — the first attempt lost `affected-packages` and `code-quality` the same way, which then gated the whole matrix.

**Everything that covers this change executed and passed:**

```
affected-packages                                ✅
code-quality / code-quality                      ✅
changelog-entry-check (dbt-adapters)             ✅
integration-tests / integration-tests-postgres   ✅  572 passed, 12 skipped
integration-tests-postgres-psycopg2-check        ✅
unit-tests (dbt-adapters)                        ✅
unit-tests (dbt-redshift)                        ✅
verify-build (dbt-adapters)                      ✅
verify-build (dbt-redshift)                      ✅
```

The guard is core Jinja shipped in `dbt-adapters`, and the five new tests are in the `dbt-postgres` functional suite — `unit-tests (dbt-adapters)`, `verify-build (dbt-adapters)` and `integration-tests-postgres` cover both. The jobs that never ran are warehouse suites requiring credentials, none of which touch this macro.

One useful detail from that green postgres run: **572 passed** in CI vs **570 passed + 2 errors** locally. The two extra are `test_simple_copy.py::TestSimpleCopyUppercase` and `test_mixed_case_db.py`, which error against the mixed-case database in my local container and pass in CI — confirming they were environmental, as noted in the description.

Happy to re-run once GitHub settles if you'd rather see a fully green board before merging, but I've stopped re-running for now since two further attempts produced no new information about the code.

will-sargent-dbtlabs:
@tauhid621 — one thing I'd like your explicit call on before merge, because I described it as settled in the thread above and on reflection it isn't mine to settle alone.

**Should `on_schema_change='append_new_columns'` warn or raise when source columns come back empty?**

I chose **warn**. The reasoning is in the description, but the short version is that every incremental materialization already absorbs an empty return via the `{% if not dest_columns %}` fallback, so those runs currently *pass* — raising would convert working pipelines into hard failures, and nothing can be dropped under `append_new_columns` anyway.

What that costs, stated plainly:

- A genuinely new column goes **undetected**, silently, until introspection works or someone runs `--full-refresh`. The warning says so, but a warning is easier to ignore than an error.
- The warning does **not** increment the run summary's `WARN` count — a warned run still prints `PASS=1 WARN=0`. It's in the log body only.
- Anyone on `--warn-error` gets a **new hard failure** on a config that previously passed. That's the opt-in escape hatch working as intended, but it is still new breakage for strict CI setups.

So it's a trade between "never break a working run" and "never let schema drift go unnoticed." I weighted the first, partly because #2097 has since removed the only path we know reaches this condition, which makes the guard defense-in-depth rather than an active fix — but that weighting is a product call as much as a technical one.

If you'd rather it raise for every non-`ignore` value, that's a smaller diff than what's here: delete the `append_new_columns` branch from the message and the `if/else` around `warn` vs `raise_compiler_error`, and drop two of the five tests. Say which you prefer and I'll make it match — I have no attachment to the current shape.

will-sargent-dbtlabs:
@tauhid621 Just wanted to follow up for your re-review on my changes.

tauhid621 (approved):

PR #2097, opened 2026-07-28
fix(redshift): describe temp relations from the driver when the catalog cannot see them

### LINKED ISSUE #1947: [Bug] Connecting to Redshift (datashare) external database breaks on_schema_change

will-sargent-dbtlabs:
Fixes #1947, #1991.

## Root cause

When the connection's database is a **datashare consumer database**
(`svv_redshift_databases.database_type = 'shared'`), temporary relations are created and remain
fully queryable, but they are absent from **every** catalog view. Verified on a purpose-built
producer/consumer datashare (Redshift 1.0.358853), one session each, identical statements:

| Catalog queried | control `dev` (local) | `ds_consumer` (datashare) |
|---|---|---|
| `information_schema.columns` | 3 columns | **0 rows** |
| `pg_attribute` via `pg_class` | 3 columns | **0 rows** |
| `svv_columns` | 3 columns | **0 rows** |
| `select count(*)` from the relation | 1 row | **1 row** |

The session's temp schema belongs to the *default* database, while catalog views resolve against
the consumer database. `SHOW COLUMNS` can't address the relation either — it requires a database
and schema, and temp relations are built with `relation.include(database=False, schema=False)`.

So `redshift__get_columns_in_relation_legacy` returned zero rows **with statement status
`FINISHED`** — a silent empty result. `on_schema_change='sync_all_columns'` read that as "the
source has no columns" and issued `ALTER TABLE … DROP COLUMN` for every column in the target. The
incremental `DELETE` then failed on the now-missing unique key, which is the error users report.

**There is no alternative catalog to switch to.** I checked all of these from a consumer-db session:

| Attempt | Result |
|---|---|
| `pg_namespace` temp schemas | 0 rows (despite `search_path` = `{pg_temp_4,…}`) |
| `SHOW COLUMNS` w/ default db + temp schema | 0 rows |
| `SHOW COLUMNS` w/ consumer db + `public` | 0 rows |
| `dev.information_schema.columns` (3-part) | `ERROR: Only user perm tables supported in producer views` |
| `dev.pg_catalog.pg_attribute` (3-part) | `ERROR: Schema pg_catalog does not exist in the database` |

Reading the driver's cursor description is the only route that doesn't depend on database context.
Full write-up: https://github.com/dbt-labs/dbt-adapters/issues/1947#issuecomment-5108260475

## Fix

`RedshiftAdapter.get_columns_in_temp_relation()` runs `select * from <identifier> limit 0` and builds
columns from `cursor.description`.

Two deliberate design choices:

**1. Fallback only, never a replacement.** The macro keeps using the legacy catalog query and only
falls back when it returns *zero* columns, **and only for a relation with no database or schema**
— i.e. a temp relation. The same `else` branch is taken for fully qualified relations whenever
`datasharing` is off, and the driver query is unqualified, so without that gate a qualified relation
that legitimately returns zero rows would be described from whatever `search_path` resolves the bare
identifier to. Setups that work today are bit-for-bit unaffected; the driver path engages exactly
where the catalog has already failed.

**2. Map by type OID, and follow whichever catalog described the target.** Schema comparison diffs
these columns against the target relation, so the reported data type has to match *exactly* or
`diff_column_data_types` reports a change on every run and `ALTER COLUMN TYPE` churns forever.
`TYPE_OID_TO_DATA_TYPE` is keyed on OID (stable) rather than `redshift_connector`'s labels. Two
values are easy to get wrong:

| OID | pg type | catalog reports | naive guess |
|---|---|---|---|
| 1042 | `bpchar` | **`character`** | `character varying` ✗ |
| 6551 | `varbyte` | **`binary varying`** | `varbyte` ✗ |

`bpchar` mistaken for `character varying` yields `character varying(n)` vs `character(n)`; `varbyte`
mistaken for `varbyte` mismatches outright. Either produces a type change that never converges.

**Which catalog describes the target depends on `datasharing`**: `SHOW COLUMNS` when it is on,
`information_schema` when it is off — and it defaults to `False`, with nothing forcing it on when
`dbname` points at a consumer database. Those two do not agree everywhere. `information_schema`
reports `intervaly2m`/`intervald2s` where `SHOW COLUMNS` reports `interval year to month`/`interval
day to second`, so one static map cannot satisfy both.
`TYPE_OID_TO_INFORMATION_SCHEMA_DATA_TYPE` carries exactly those two, selected on `use_show_apis()`;
a unit test pins its contents so it cannot widen silently. Every other mapped type agrees across both
catalogs — established by the probe below rather than assumed.

Size fields are only populated where they actually affect `Column.data_type` — `char_size` for
string types, precision/scale for `numeric`/`decimal`. `is_numeric()` covers only `numeric`/`decimal`,
so for ints and floats the size fields are left unset regardless of source.

**3. Size comes from `type_modifier`, not `cursor.description`.** An earlier version of this fix read
`precision`/`internal_size` off the PEP 249 `description` tuple. Verified directly against the
installed `redshift_connector` source (`Cursor._getDescription`): those fields are hardcoded to `None`
on every version dbt-redshift depends on — not merely undocumented, dead code. Caught this by running
the new gated suite live: a `numeric(18,2)` column got silently altered to `numeric(18,0)` and moved
to the end of the table on the very next run, because an unset precision/scale rendered as Redshift's
bare-numeric default, which didn't match the target's declared `numeric(18,2)` and triggered exactly
the "type change that never converges" failure this fix exists to prevent.

The real value is still on the wire, just not exposed through the public API: the driver parses
`cursor.ps["row_desc"]`'s `type_modifier` (`pg_attribute.atttypmod`) per column but never surfaces it.
Reading it directly and decoding with the standard Postgres convention (`length = type_modifier - 4`
for strings; `precision, scale = (m - 4) >> 16, (m - 4) & 0xFFFF` for numeric) is the only way to get
real sizes. `cursor.ps` is undocumented driver internals, not a stable public API — there's no
supported alternative, and a future driver version could restructure it. `get_columns_in_temp_relation`
falls back to no size info rather than raising if it does.

An OID the map doesn't cover falls back to the driver's own label with a debug log. If
`redshift_connector` cannot name it either — its lookup is `RedshiftOID(oid).name`, an `IntEnum` call
that raises — then the column cannot be described at all, so it raises with the offending type code
rather than inventing a name that only resurfaces later as invalid DDL.

## Snapshots are also affected, not just incremental models

The zero-columns condition isn't specific to `on_schema_change='sync_all_columns'`. Snapshots hit
the same root cause through a different door: `snapshot.sql` builds a `__dbt_tmp` staging relation
and calls `adapter.get_columns_in_relation`/`get_missing_columns` on it directly — there's no
`on_schema_change` involved, so #2096's guard doesn't apply to this path and isn't meant to.

A zero-columns read on the staging relation empties `snapshot_merge_sql`'s insert column list,
producing:

```sql
insert into <target> ()
select
from <staging>__dbt_tmp... as DBT_INTERNAL_SOURCE
where DBT_INTERNAL_SOURCE.dbt_change_type::text = 'insert'::text;
```

— a Redshift syntax error at the empty `()`, rather than incremental's silent column drop. Louder
failure, same root cause. This is not a synthetic scenario: it's the shape of a real production
snapshot job failing against a datashare consumer connection.

This fix covers snapshots too without any snapshot-specific change, since it's implemented at the
`redshift__get_columns_in_relation` macro level — every caller benefits, not just the incremental
materialization. Added `TestSnapshotDatashareConsumer`/`TestSnapshotCheckDatashareConsumer` to the
gated suite below to actually exercise that rather than rely on the trace.

## Tests

**Unit** (`tests/unit/test_temp_relation_columns.py`, 32 tests, all passing) — pins all OID
mappings against the values captured from `SHOW COLUMNS`, with explicit regression tests for the
`bpchar` and `varbyte` cases, plus column-construction tests covering which size fields get
populated, unknown type codes, an empty description, and a missing/malformed `cursor.ps` (the
stub mirrors the real driver's split: dead size fields on `description`, real `type_modifier` on
`ps["row_desc"]`, so a regression here fails the same way the real driver would).

**Functional — describer agreement** (`tests/functional/test_type_oid_mapping.py`, 5 tests) — builds
a table with one column per mapped type and requires `SHOW COLUMNS`, `information_schema` and the
driver to report the same `Column.data_type`, once with `datasharing` on and once off. It needs no
datashare, so unlike the gated suite below it runs in ordinary integration CI. A third test fails if
an OID is added to the map without either a probe or an explicit exemption, so the agreement claim
can't rot.

**Functional — datashare consumer** (`tests/functional/adapter/datashare_consumer/`) — a gated suite for the topology
that currently has **no coverage at all**:

| test | connection `dbname` | model database | covered before |
|---|---|---|---|
| `TestIncrementalOnSchemaChangeWithDatasharing` | default | same | yes |
| `TestIncrementalCrossDatabase*` | default | cross-db via `+database` | yes |
| **`datashare_consumer/`** | **consumer database** | same | **no** |

`CrossDatabaseMixin` keeps the connection on the default database and only retargets models via
`+database` — which is the documented *workaround*. That's why this shipped with green CI. The new
`DatashareConsumerMixin` points `dbname` straight at the consumer database. Gated behind
`REDSHIFT_TEST_DATASHARE_DBNAME` (skips when unset), following the existing `cross_database/` pattern;
`fixtures.py` documents the full producer/consumer setup SQL.

Beyond asserting columns survive, `test_incremental_on_schema_change.py` runs the model three times
and asserts the column set *and* dtypes are stable, which is what would catch a type-mapping
regression. `test_snapshot.py` adds `TestSnapshotDatashareConsumer` (timestamp strategy) and
`TestSnapshotCheckDatashareConsumer` (check strategy), reusing dbt's existing `BaseSimpleSnapshot`/
`BaseSnapshotCheck` suites — 12 tests total in the directory now.

## Verified live against a real datashare cluster

Stood up a fresh producer/consumer pair (Redshift Serverless, `1.0.384821`) with a write-enabled
datashare (`GRANT CREATE, USAGE ON SCHEMA ... TO DATASHARE`) and ran the full gated suite over the
real Python driver (`redshift_connector==2.1.16`), not just the Data API used to trace the original
root cause. Result: **12/12 pass**, including `TestIncrementalTempRelationColumnsPreserved`'s 3-run
stability check and both snapshot strategies. Confirmed via `SHOW COLUMNS` that a `numeric(18,2)` and
a `character(5)` column keep their exact declared type, size, and ordinal position across three runs.

This live run is what caught the `cursor.description` issue described above — the fix in this PR
reflects that finding, not the version that was originally run. One unrelated flake surfaced and was
run down separately: `test_deletes_are_captured_by_snapshot` hit an internal Redshift engine assertion
(`federation_api.cpp:1341`, `CanLinkStorageOid`) — but only on a schema that had gone through
`DROP SCHEMA CASCADE` + recreate churn from my own repeated test iterations. The identical test passes
cleanly on a genuinely untouched schema, and this is a storage-bookkeeping quirk of the write-datashare
feature itself under drop/recreate churn, not something a real environment's stable schemas would hit.

A second live run, against a plain Serverless workgroup with no datashare, exercised the
describer-agreement probe: **5/5 pass** in both `datasharing` modes. That run is what surfaced the
interval divergence described above, and the geometry finding now split out to #2108. CI's
`integration-tests-redshift` job is green on the head commit, so the probe has since run in CI too.

One caveat remains: `cursor.ps["row_desc"]` is undocumented driver internals, so if a future
`redshift_connector` release restructures it, size extraction silently degrades to "unset" rather than
erroring (by design — see `get_columns_in_temp_relation`'s try/except). Worth a periodic sanity check
against new driver releases; not blocking.

## Related

#2096 adds an adapter-agnostic guard so an empty source column list raises instead of dropping
columns. That's the backstop; this is the actual repair. They're independent and complementary — with
both, a warehouse that hides temp relations gets fixed here, and any *other* adapter hitting the same
class of metadata failure fails loudly instead of silently.

#2108 records a pre-existing bug this PR deliberately does **not** fix. A `geometry` column reports
OID 3999 (`GEOMETRYHEX`) on the wire while both catalogs report `geometry`, so any driver-derived type
name disagrees with the target. It breaks `sync_all_columns` on `main` independently of this fallback
and needs its own coverage, so it is recorded in the probe's `KNOWN_DIVERGENT_OIDS` with a one-line
reproduction instead of being fixed here.

### DIFF
--- a/dbt-redshift/src/dbt/adapters/redshift/impl.py
+++ b/dbt-redshift/src/dbt/adapters/redshift/impl.py
@@ -253,32 +253,58 @@ class RedshiftAdapter(SQLAdapter):
         database context. Type codes are mapped back to the names ``SHOW COLUMNS`` reports
         so that columns described this way compare equal to columns of the target relation
         in ``check_for_schema_changes``.
+
+        Size/precision/scale cannot come from ``cursor.description``: ``redshift_connector``
+        (the version dbt-redshift depends on) hardcodes those PEP 249 fields to ``None`` --
+        verified directly against its source (``Cursor._getDescription``), not merely
+        undocumented. The real values are still on the wire: the driver parses the
+        Postgres-protocol row description into ``cursor.ps["row_desc"]``, which carries
+        ``type_modifier`` (``pg_attribute.atttypmod``) per column but never surfaces it
+        through the public API. Reading it directly is the only way to get real sizes;
+        without it, every sized string/numeric column looks "changed" against the target
+        on the very next run, and Redshift's add/copy/drop/rename response to that
+        reorders the column and can silently narrow its scale.
         """
         sql = f"select * from {self.quote(relation.identifier)} limit 0"
         _, cursor = self.connections.add_select_query(sql)
 
+        # `cursor.ps["row_desc"]` is an internal, undocumented structure of
+        # `redshift_connector`, not a public API -- there is no supported alternative for
+        # reaching `type_modifier`. `_getDescription` builds `cursor.description` from this
+        # same list, in the same order, so index-aligned lookup here is safe.
+        try:
+            row_descriptions = cursor.ps["row_desc"]
+        except (AttributeError, KeyError, TypeError):
+            row_descriptions = []
+
         columns = []
-        for description in cursor.description or []:
+        for i, description in enumerate(cursor.description or []):
             # PEP 249: (name, type_code, display_size, internal_size, precision, scale, null_ok)
             column_name = description[0]
             type_code = description[1]
-            internal_size = description[3] if len(description) > 3 else None
-            precision = description[4] if len(description) > 4 else None
-            scale = description[5] if len(description) > 5 else None
+
+            type_modifier = None
+            if i < len(row_descriptions):
+                type_modifier = row_descriptions[i].get("type_modifier")
 
             data_type = self._temp_relation_data_type(type_code)
 
             # Only string and exact-numeric types feed size into Column.data_type, which is
             # what schema comparison comes down to. For every other type the size fields are
-            # ignored, so leave them unset rather than propagate the driver's display widths.
+            # ignored, so leave them unset.
             char_size = None
             numeric_precision = None
             numeric_scale = None
-            if data_type in ("character varying", "character"):
-                char_size = next((s for s in (precision, internal_size) if s and s > 0), None)
-            elif data_type == "numeric":
-                numeric_precision = precision
-                numeric_scale = scale
+            if type_modifier is not None and type_modifier >= 0:
+                # Postgres wire-protocol convention: VARHDRSZ (4 bytes) is added to both a
+                # string's declared length and a numeric's packed (precision, scale) pair.
+                # -1 means "no modifier" (unconstrained/default size) and is left unset.
+                raw_modifier = type_modifier - 4
+                if data_type in ("character varying", "character"):
+                    char_size = raw_modifier if raw_modifier > 0 else None
+                elif data_type == "numeric":
+                    numeric_precision = raw_modifier >> 16
+                    numeric_scale = raw_modifier & 0xFFFF
 
             columns.append(
                 self.Column(
--- a/dbt-redshift/src/dbt/adapters/redshift/impl.py
+++ b/dbt-redshift/src/dbt/adapters/redshift/impl.py
@@ -42,6 +42,44 @@ for package in packages:
 GET_RELATIONS_MACRO_NAME = "redshift__get_relations"
 SHOW_TABLES_FROM_SCHEMA_MACRO_NAME = "redshift__show_tables_from_schema"
 
+# Redshift type OID -> the data type name SHOW COLUMNS reports for that type.
+#
+# Used to describe temporary relations from the driver's cursor description when the
+# catalog cannot see them (see RedshiftAdapter.get_columns_in_temp_relation). Keying on the
+# OID rather than on redshift_connector's own type labels keeps this independent of driver
+# label changes, and lets the values match SHOW COLUMNS exactly -- necessary because
+# incremental schema comparison diffs columns described this way against columns of the
+# target relation, which are described by SHOW COLUMNS.
+#
+# Verified against `SHOW COLUMNS FROM TABLE` on Redshift 1.0.358853. Note in particular
+# that bpchar reports as "character" (not "character varying") and varbyte reports as
+# "binary varying" (not "varbyte"); getting either wrong produces a spurious type change
+# on every run.
+TYPE_OID_TO_DATA_TYPE: Dict[int, str] = {
+    16: "boolean",  # bool
+    20: "bigint",  # int8
+    21: "smallint",  # int2
+    23: "integer",  # int4
+    25: "character varying",  # text
+    700: "real",  # float4
+    701: "double precision",  # float8
+    1042: "character",  # bpchar
+    1043: "character varying",  # varchar
+    1082: "date",  # date
+    1083: "time without time zone",  # time
+    1114: "timestamp without time zone",  # timestamp
+    1184: "timestamp with time zone",  # timestamptz
+    1188: "interval year to month",  # intervaly2m
+    1190: "interval day to second",  # intervald2s
+    1266: "time with time zone",  # timetz
+    1700: "numeric",  # numeric
+    2935: "hllsketch",  # hllsketch
+    3000: "geometry",  # geometry
+    3001: "geography",  # geography
+    4000: "super",  # super
+    6551: "binary varying",  # varbyte
+}
+
 REDSHIFT_SKIP_AUTOCOMMIT_TRANSACTION_STATEMENTS = BehaviorFlag(
     name="redshift_skip_autocommit_transaction_statements",
     default=False,
@@ -198,6 +236,78 @@ class RedshiftAdapter(SQLAdapter):
         Returns True when the ``datasharing`` profile config is enabled.
         """
         return bool(self.config.credentials.datasharing)
+
+    @available
+    def get_columns_in_temp_relation(self, relation: BaseRelation) -> List[Any]:
+        """Describe a temporary relation without consulting the catalog.
+
+        When the connection's database is a datashare consumer database
+        (``svv_redshift_databases.database_type = 'shared'``), temporary relations are
+        created and remain queryable but are invisible to ``information_schema.columns``,
+        ``pg_attribute`` and ``svv_columns`` alike -- the session's temp schema belongs to
+        the default database while catalog views resolve against the consumer database.
+        ``SHOW COLUMNS`` cannot address them either, since it requires a database and
+        schema and temporary relations carry neither.
+
+        Reading the driver's cursor description is the only route that does not depend on
+        database context. Type codes are mapped back to the names ``SHOW COLUMNS`` reports
+        so that columns described this way compare equal to columns of the target relation
+        in ``check_for_schema_changes``.
+        """
+        sql = f"select * from {self.quote(relation.identifier)} limit 0"
+        _, cursor = self.connections.add_select_query(sql)
+
+        columns = []
+        for description in cursor.description or []:
+            # PEP 249: (name, type_code, display_size, internal_size, precision, scale, null_ok)
+            column_name = description[0]
+            type_code = description[1]
+            internal_size = description[3] if len(description) > 3 else None
+            precision = description[4] if len(description) > 4 else None
+            scale = description[5] if len(description) > 5 else None
+
+            data_type = self._temp_relation_data_type(type_code)
+
+            # Only string and exact-numeric types feed size into Column.data_type, which is
+            # what schema comparison comes down to. For every other type the size fields are
+            # ignored, so leave them unset rather than propagate the driver's display widths.
+            char_size = None
+            numeric_precision = None
+            numeric_scale = None
+            if data_type in ("character varying", "character"):
+                char_size = next((s for s in (precision, internal_size) if s and s > 0), None)
+            elif data_type == "numeric":
+                numeric_precision = precision
+                numeric_scale = scale
+
+            columns.append(
+                self.Column(
+                    column=column_name,
+                    dtype=data_type,
+                    char_size=char_size,
+                    numeric_precision=numeric_precision,
+                    numeric_scale=numeric_scale,
+                )
+            )
+
+        return columns
+
+    def _temp_relation_data_type(self, type_code: Any) -> str:
+        """Map a driver type code onto the data type name ``SHOW COLUMNS`` reports."""
+        data_type = TYPE_OID_TO_DATA_TYPE.get(type_code)
+        if data_type is not None:
+            return data_type
+
+        # Unrecognised type code: the driver's own label is closer than nothing, but it is
+        # not guaranteed to match SHOW COLUMNS, so surface it.
+        fallback = str(self.connections.data_type_code_to_name(type_code)).lower()
+        logger.debug(
+            f"No known Redshift data type for type code {type_code}; "
+            f"falling back to {fallback!r}. Schema comparison for this column may be "
+            f"inaccurate -- please report this at "
+            f"https://github.com/dbt-labs/dbt-adapters/issues"
+        )
+        return fallback
 
     @available
     def drop_without_cascade(self) -> bool:
--- a/dbt-redshift/src/dbt/adapters/redshift/impl.py
+++ b/dbt-redshift/src/dbt/adapters/redshift/impl.py
@@ -69,7 +69,6 @@ TYPE_OID_TO_DATA_TYPE: Dict[int, str] = {
     2935: "hllsketch",  # hllsketch
     3000: "geometry",  # geometry
     3001: "geography",  # geography
-    3999: "geometry",  # geometryhex -- what the wire returns for a geometry column
     4000: "super",  # super
     6551: "binary varying",  # varbyte
 }
--- a/dbt-redshift/src/dbt/adapters/redshift/impl.py
+++ b/dbt-redshift/src/dbt/adapters/redshift/impl.py
@@ -42,19 +42,12 @@ for package in packages:
 GET_RELATIONS_MACRO_NAME = "redshift__get_relations"
 SHOW_TABLES_FROM_SCHEMA_MACRO_NAME = "redshift__show_tables_from_schema"
 
-# Redshift type OID -> the data type name SHOW COLUMNS reports for that type.
-#
-# Used to describe temporary relations from the driver's cursor description when the
-# catalog cannot see them (see RedshiftAdapter.get_columns_in_temp_relation). Keying on the
-# OID rather than on redshift_connector's own type labels keeps this independent of driver
-# label changes, and lets the values match SHOW COLUMNS exactly -- necessary because
-# incremental schema comparison diffs columns described this way against columns of the
-# target relation, which are described by SHOW COLUMNS.
-#
-# Verified against `SHOW COLUMNS FROM TABLE` on Redshift 1.0.358853. Note in particular
-# that bpchar reports as "character" (not "character varying") and varbyte reports as
-# "binary varying" (not "varbyte"); getting either wrong produces a spurious type change
-# on every run.
+# Redshift type OID -> SQL data type name, as reported by both the legacy path's
+# information_schema.columns.data_type and SHOW COLUMNS. Used to describe temp relations
+# from the driver's cursor description (see get_columns_in_temp_relation); the names have
+# to match whichever of those two paths described the target relation, or schema comparison
+# reports a type change on every run. Verified against SHOW COLUMNS on Redshift 1.0.358853
+# -- note bpchar reports as "character", and varbyte as "binary varying".
 TYPE_OID_TO_DATA_TYPE: Dict[int, str] = {
     16: "boolean",  # bool
     20: "bigint",  # int8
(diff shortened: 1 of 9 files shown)


dbt-redshift/src/dbt/include/redshift/macros/adapters.sql
```diff
@@ -130,7 +130,26 @@
...
+      information_schema.columns, pg_attribute and svv_columns, because the session's temp
+      schema belongs to the default database while those views resolve against the consumer
+      database. SHOW COLUMNS cannot address it either -- it needs a database and schema, and
+      temporary relations have neither.
+
+      Left unhandled, an empty list makes on_schema_change='sync_all_columns' treat every
+      column in the target as removed and drop it. Fall back to describing the relation
+      through the driver, which does not depend on database context.
+    -#}
+    {%- if columns | length == 0 -%}
```
tauhid621: Let's also gate this for `relation.database and relation.schema` for temp relations.

dbt-redshift/src/dbt/adapters/redshift/impl.py
```diff
@@ -42,6 +42,44 @@
 GET_RELATIONS_MACRO_NAME = "redshift__get_relations"
 SHOW_TABLES_FROM_SCHEMA_MACRO_NAME = "redshift__show_tables_from_schema"
 
+# Redshift type OID -> the data type name SHOW COLUMNS reports for that type.
```
tauhid621: I am not sure if this should be mapped to SHOW COLUMNS. 
The legacy path uses information schema. so we should map it so legacy path. 

Ideally it should be the same.

dbt-redshift/src/dbt/adapters/redshift/impl.py
```diff
@@ -42,6 +42,44 @@
...
+# catalog cannot see them (see RedshiftAdapter.get_columns_in_temp_relation). Keying on the
+# OID rather than on redshift_connector's own type labels keeps this independent of driver
+# label changes, and lets the values match SHOW COLUMNS exactly -- necessary because
+# incremental schema comparison diffs columns described this way against columns of the
+# target relation, which are described by SHOW COLUMNS.
+#
+# Verified against `SHOW COLUMNS FROM TABLE` on Redshift 1.0.358853. Note in particular
+# that bpchar reports as "character" (not "character varying") and varbyte reports as
+# "binary varying" (not "varbyte"); getting either wrong produces a spurious type change
+# on every run.
```
tauhid621: Let's reduce the comments.

dbt-redshift/src/dbt/adapters/redshift/impl.py
```diff
@@ -199,6 +237,104 @@ def use_show_apis(self) -> bool:
...
+        (the version dbt-redshift depends on) hardcodes those PEP 249 fields to ``None`` --
+        verified directly against its source (``Cursor._getDescription``), not merely
+        undocumented. The real values are still on the wire: the driver parses the
+        Postgres-protocol row description into ``cursor.ps["row_desc"]``, which carries
+        ``type_modifier`` (``pg_attribute.atttypmod``) per column but never surfaces it
+        through the public API. Reading it directly is the only way to get real sizes;
+        without it, every sized string/numeric column looks "changed" against the target
+        on the very next run, and Redshift's add/copy/drop/rename response to that
+        reorders the column and can silently narrow its scale.
+        """
```
tauhid621: Let's reduce the comments

dbt-redshift/src/dbt/adapters/redshift/impl.py
```diff
@@ -199,6 +237,104 @@ def use_show_apis(self) -> bool:
...
+
+    def _temp_relation_data_type(self, type_code: Any) -> str:
+        """Map a driver type code onto the data type name ``SHOW COLUMNS`` reports."""
+        data_type = TYPE_OID_TO_DATA_TYPE.get(type_code)
+        if data_type is not None:
+            return data_type
+
+        # Unrecognised type code: the driver's own label is closer than nothing, but it is
+        # not guaranteed to match SHOW COLUMNS, so surface it.
+        fallback = str(self.connections.data_type_code_to_name(type_code)).lower()
```
tauhid621: This can be an exception. we should handle it.

tauhid621 (commented):
Overall approach looks sound. Left some comments

dbt-redshift/src/dbt/include/redshift/macros/adapters.sql
will-sargent-dbtlabs: Done in `de53e789` — gated on `not relation.database and not relation.schema`.

Worth flagging why this matters beyond tidiness: this `else` branch is also taken for **fully qualified** relations whenever `datasharing` is off. The driver query is unqualified (`select * from <identifier>`), so without the gate, a schema-qualified relation that legitimately returns zero rows (dropped, or a permissions gap) would get described from whatever `search_path` resolves that bare identifier to — a wrong answer rather than a recovery.

Confirmed the gate doesn't narrow the fix: Redshift inherits `postgres__make_temp_relation`, which ends in `incorporate(path={"schema": none, "database": none})`, so temp relations carry only an identifier. Both the incremental path and the snapshot staging path (`snapshots/helpers.sql`, same `make_temp_relation`) still reach it.

dbt-redshift/src/dbt/adapters/redshift/impl.py
will-sargent-dbtlabs: Good catch, and it's a live bug rather than an unhandled edge: `data_type_code_to_name` is `RedshiftOID(oid).name` — an `IntEnum` call, so it raises `ValueError` for any OID missing from the driver's own table. That is exactly the case this fallback was written for, so the old code raised where it meant to degrade.

It went uncaught because the unit stub returned `"UNKNOWN"` where the real driver raises. Fixed the stub to mirror the driver, which split the coverage in two:

- OIDs the driver **can** name but this adapter doesn't map (`1015` varchar_array, `28` xid, …) still degrade to the driver's label with a debug log.
- If neither side can name it, it now raises `DbtRuntimeError` naming the type code. Describing the column with an invented type name would only resurface later as invalid DDL (`alter table ... add column <nonsense>`), so failing here with something actionable is strictly better.

Both branches are tested now. `de53e789`.

dbt-redshift/src/dbt/adapters/redshift/impl.py
will-sargent-dbtlabs: You were right, and it's worth more than a reword — `datasharing` defaults to `False` and nothing forces it on when `dbname` points at a consumer database, so information_schema really can be the describer, not just `SHOW COLUMNS`.

Rather than argue it, I probed it. `tests/functional/test_type_oid_mapping.py` builds a table with one column per mapped type, describes it through `get_columns_in_relation` (which dispatches on `datasharing`) and through the driver, and requires `Column.data_type` to agree — one subclass per mode. It needs no datashare, so unlike the gated `datashare_consumer/` suite it runs in ordinary integration CI.

I ran it against a live Serverless cluster. **The 14 standard types agree in both modes. Two things did not:**

| type | OID | `SHOW COLUMNS` | `information_schema` | driver (before) |
|---|---|---|---|---|
| `interval year to month` | 1188 | `interval year to month` | **`intervaly2m`** | `interval year to month` |
| `interval day to second` | 1190 | `interval day to second` | **`intervald2s`** | `interval day to second` |
| `geometry` | 3000 | `geometry` | `geometry` | **`geometryhex`** |

So exactly your point: information_schema reports its internal name for the interval types, and one static map can't satisfy both describers. `TYPE_OID_TO_INFORMATION_SCHEMA_DATA_TYPE` now holds just those two, selected on `use_show_apis()`. A unit test pins that dict's exact contents so it can't widen silently.

**The probe also caught something I wasn't looking for.** A `geometry` column reports OID **3999** (`GEOMETRYHEX`) in the cursor description, never 3000 — 3999 is a wire representation, not a declarable type. The `3000` entry was therefore dead on the driver path, and every geometry column was falling through to the unknown-type fallback and reporting `geometryhex` against a target reporting `geometry`: a type change that never converges. Worth noting this misbehaves on `main` too, for any `sync_all_columns` model with a geometry column — it isn't introduced by this PR, just newly visible.

`hllsketch`, `geography`, `super` and `varbyte` all agree across both catalogs. OID 25 (`text`) is exempted in `UNPROBEABLE_OIDS`: Redshift aliases `TEXT` to `VARCHAR(256)`, so no stored column carries it and there's nothing to compare against.

Live result: **5 passed in both `datasharing` modes**, plus 205 unit tests. `79739907`.

will-sargent-dbtlabs:
Correcting scope on one of the two probe findings.

The interval divergence **stays** — it's required for this fix to work reliably. `datasharing` decides which catalog describes the target, information_schema reports `intervaly2m`/`intervald2s` where `SHOW COLUMNS` reports the SQL names, and without following the right one a temp relation with an interval column reads as changed on every run. Narrow, proven live in both modes.

The **geometry finding is now out of scope**, filed as #2108. Mapping OID 3999 → `geometry` is a behaviour change for a type with no other coverage in this PR, and the bug pre-dates the fallback entirely: it hits any `sync_all_columns` model with a geometry column regardless of whether the catalog can see the temp relation. It also raises questions I haven't tested — whether `geometryhex` is ever legitimately the right name, and whether `geography` has an equivalent wire OID (it reports 3001 and agrees with both catalogs, but only checked for a plain column). That belongs in its own PR with its own coverage rather than riding along here.

It's recorded rather than dropped: `KNOWN_DIVERGENT_OIDS` in the probe points at #2108 and notes the one-line addition to `REDSHIFT_TYPES` that reproduces it.

So this PR is back to exactly the scoped fix: describe temp relations from the driver when the catalog can't see them, and name the columns the way whichever catalog describes the target does. `56f2799f`.

One note on verification: 204 unit tests and the probe's coverage guard pass locally, and the full probe passed live in both `datasharing` modes at `79739907`. This last commit only *removes* a probe row and a map entry, so it can't affect the assertions that passed — but I'd already torn down the test cluster when I made it, so the live suite hasn't been re-run against this exact commit. CI will cover it.

will-sargent-dbtlabs:
@tauhid621 Looks like we are ready for a re-review!

tauhid621 (approved):

dbt-redshift/src/dbt/adapters/redshift/impl.py
tauhid621: thanks

will-sargent-dbtlabs:
Thanks @tauhid621 !  I know at some point the customer is going to want to understand when they can test these changes in platform. Any idea on then they will be included in the platform release track?

tauhid621:
@will-sargent-dbtlabs It should be released mid next week in `latest` release track

PR #2105, opened 2026-08-03
fix(bigquery): normalize a custom api_endpoint

### LINKED ISSUE #2103: [dbt-bigquery] Custom api_endpoint causes literal "https" hostname resolution failure ("Reopening connection" loop)
### Describe the bug

Jobs on an account with a custom BigQuery `api_endpoint` configured are failing continuously with the adapter attempting to resolve the literal string `https` as a DNS hostname, rather than the actual configured endpoint host. This breaks every affected run and the adapter enters a retry loop without recovering.

### Error signature

```
BigQuery adapter: Reopening connection after ConnectionError(MaxRetryError('HTTPSConnectionPool(host=\'https\', port=443): Max retries exceeded with url: //<custom-endpoint-host>/bigquery/v2/projects/<project>/datasets?maxResults=10000&prettyPrint=false (Caused by NameResolutionError("HTTPSConnection(host=\'https\', port=443): Failed to resolve \'https\' ([Errno -2] Name or service not known)"))'))
```

and

```
Encountered an error:
Database Error
  HTTPSConnectionPool(host='https', port=443): Max retries exceeded with url: //<custom-endpoint-host>/bigquery/v2/projects/<project>/datasets?maxResults=10000&prettyPrint=false (Caused by NameResolutionError("HTTPSConnection(host='https', port=443): Failed to resolve 'https' ([Errno -2] Name or service not known)"))
```

Note the request line: `url: //<custom-endpoint-host>/bigquery/v2/projects/...` — the scheme (`https:`) has been separated from the host during URL construction, so the HTTP client ends up trying to open a connection to a host literally named `https` on port 443, instead of the configured endpoint host.

### Steps to reproduce / observed context

- A dbt Cloud account has a custom BigQuery `api_endpoint` configured pointing at a third-party proxy in front of BigQuery (e.g. `https://<proxy-host>`).
- The stored `api_endpoint` value was visually confirmed (via screenshot in support tooling) to be a single, correctly-formed URL — no duplicated scheme, no stray `https:` line.
- Despite the stored config being clean, the error recurred continuously in production logs, roughly every 5 minutes, over a multi-hour window.
- **Confirmed root cause / repro:** removing the `https://` scheme prefix from the configured `api_endpoint` field — leaving just the bare host (e.g. `<proxy-host>` instead of `https://<proxy-host>`) — resolved the issue entirely; jobs ran clean afterward. This confirms the adapter is unconditionally prepending its own `https://` scheme at request-construction time, without checking whether the configured value already includes a scheme. When the configured value already has `https://`, the resulting URL construction/parsing drops the real host and leaves `https` as the literal hostname.
- Reported via an internal dbt Labs support ticket; happy to share account/connection details privately with a maintainer if needed for reproduction, rather than posting them publicly here.

### Expected behavior

When a custom `api_endpoint` is configured, the adapter should construct request URLs using the full scheme+host as configured, without splitting or dropping the scheme, regardless of whether the user's input already includes a scheme. Suggested fix direction: check for an existing `http://`/`https://` prefix before prepending one (or normalize by stripping any existing scheme first, then prepending consistently), so both scheme-included and bare-host input work correctly.

### Is this related to the known Fusion issue dbt-core#14617?

No — confirmed this is **not** a duplicate of https://github.com/dbt-labs/dbt-core/issues/14617. That issue is Fusion-only (engine:v2), affects the Storage Read API gRPC path, and produces a distinct error (`"[BigQuery] Storage API is not available for query"`). This issue's error comes from the classic Python `BigQuery adapter: Reopening connection...` code path (dbt Core v1 style adapter, in `dbt-bigquery`), and Fusion is not involved in the affected jobs' execution path for the account in question.

tauhid621:
resolves #2103
docs N/A

### Problem

`api_endpoint` is handed to the BigQuery client verbatim. A value that picks up an extra scheme upstream (`https://https://host`) makes the client resolve the literal host `https`, and a bare host yields a schemeless URL.

### Solution

Normalize to a single scheme-qualified base URL, defaulting to `https` and keeping the innermost scheme when one repeats. Unusable values fall back to the client default with a warning. Unit tests cover each shape.

### Checklist

### DIFF
--- a/dbt-bigquery/src/dbt/adapters/bigquery/clients.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/clients.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from typing import Optional, TYPE_CHECKING
 
 # Lazy-loaded inside create_notebook_client() to avoid slowing down every
@@ -27,6 +28,11 @@ from dbt.adapters.bigquery.credentials import (
 
 
 _logger = AdapterLogger("BigQuery")
+
+# splits an endpoint into scheme and host. the scheme is optional and may repeat, in which
+# case the group holds the last (innermost) one, i.e. the scheme the user actually meant;
+# slashes around the host are dropped so they don't double up against the client's path
+_API_ENDPOINT = re.compile(r"(?:(?P<scheme>https?)://)*/*(?P<host>.*?)/*", re.IGNORECASE)
 
 
 def create_bigquery_client(credentials: BigQueryCredentials) -> BigQueryClient:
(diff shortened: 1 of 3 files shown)


dbt-bigquery/src/dbt/adapters/bigquery/clients.py
```diff
@@ -72,11 +78,33 @@ def _create_bigquery_client(credentials: BigQueryCredentials) -> BigQueryClient:
...
+        return None
+
+    match = _API_ENDPOINT.fullmatch(endpoint)
+    if not match or not (host := match["host"]):
+        _logger.warning(f"Ignoring api_endpoint {api_endpoint}: could not parse a host")
+        return None
+
+    normalized = f"{(match['scheme'] or 'https').lower()}://{host}"
+    _logger.debug(f"Using api_endpoint {normalized}")
+    return normalized
```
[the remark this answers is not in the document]
tauhid621: addressed

dbt-bigquery/src/dbt/adapters/bigquery/clients.py
```diff
@@ -72,11 +80,40 @@ def _create_bigquery_client(credentials: BigQueryCredentials) -> BigQueryClient:
...
+
+    google-cloud-bigquery uses this value verbatim as the base of every REST URL it
+    builds, so a bare host yields a schemeless URL and a value that picked up an extra
+    scheme somewhere upstream (`https://https://host`) resolves the literal host `https`.
+    Accept both forms instead. See https://github.com/dbt-labs/dbt-adapters/issues/2103
+    """
+    if not (endpoint := (api_endpoint or "").strip()):
+        return None
+
+    match = _API_ENDPOINT.fullmatch(endpoint)
```
VersusFacit: when I did this in Fusion I used a uri library rather than handroll regexes. Python have something similar?

tauhid621: Yes using library is much better. I have updated the code.

PR #2112, opened 2026-08-07
Compare BigQuery clustering fields case-insensitively in is_replaceab…

### LINKED ISSUE #2111: [Bug] BigQuery _clusters_match use case-sensitive comparison and causes an unnecessary table drop
### Is this a new bug?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [x] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

The method `_clusters_match` compares the configured and the stored clustering fields with a case-sensitive test (`table.clustering_fields == conf_cluster`).

A model can set `cluster_by` in a different letter case than the stored column. This was an issue we had, where the cluster was defined as `ExternalCustomerId`, but the BigQuery column was saved as `externalcustomerid`. 

The create or replace itself still works in BigQuery in this case without the need to drop the table beforehand. But in this condition, `is_replaceable()` returns `False` on every run, although the clustering did not change.

The table materialization then does a hard refresh. It calls `tables.delete` to drop the table before the rebuild query. 

So when eventually the model ran into a compilation issue, the table from the previous run dissapeared, and the atomicity of the operation was lost.

Overall, it causes two problems:

- dbt drops and recreates the table on every run. This adds cost. It also creates a period when the table does not exist.

- dbt loses atomicity. dbt drops the table before the rebuild query runs. If the model SELECT fails, no table remains. An atomic `create or replace` keeps the previous table after a failure.

The related method `_partitions_match` do changes both sides to lower case.

### Expected Behavior

BigQuery column names are not case-sensitive. Therefore `externalcustomerid` and `ExternalCustomerId` refer to the same clustering column.

`is_replaceable()` must return `True` in this condition. dbt must replace the model with an atomic `create or replace table`. dbt must not drop the table first.

If the rebuild query fails, the previous table must remain.

### Steps To Reproduce

1. Create a model with `materialized='table'` that aliases a column in lower case, for example `select customer_id as externalcustomerid`.
2. Set the clustering config in a different case: `cluster_by=["ExternalCustomerId"]`.
3. Run the model two times with dbt run.
4. Read the dbt debug log or the BigQuery audit log.
5. Find the hard-refresh message and the `tables.delete` operation before each rebuild.

### Relevant log output

```shell
Hard refreshing <database>.<schema>.<model> because it is not replaceable
```

### Environment

```markdown
- OS: Linux
- Python: 3.11
- dbt-core: 1.11.10
- dbt-adapters: 1.23.0
- dbt-bigquery: 1.11.3
```

### Additional Context

The cause is in `src/dbt/adapters/bigquery/impl.py`:

```
@staticmethod
def _clusters_match(table, conf_cluster) -> bool:
    if isinstance(conf_cluster, str):
        conf_cluster = [conf_cluster]
    return table.clustering_fields == conf_cluster   # case-sensitive comparison
```

`table.clustering_fields` returns the stored column name without a change, for example `["externalcustomerid"`]. `conf_cluster` holds the raw config value, for example `["ExternalCustomerId"]`. Therefore the comparison returns `False`.

This problem is not a regression. The same case-sensitive comparison exists in previous versions. In each of these versions, `_partitions_match` already changes both sides to lower case.

Proposed fix. Change both sides to lower case. Keep the field order, because clustering uses an ordered list.

ddamart:
resolves #2111

docs N/A

### Problem

The static method `_clusters_match` compares the configured and the stored clustering fields with a case-sensitive test (`table.clustering_fields == conf_cluster`).

A model can set `cluster_by` in a different letter case than the stored column. For example, the config gives `ExternalCustomerId`, but BigQuery stores the column as `externalcustomerid`.

In this condition, `is_replaceable()` returns `False` on every run, although the clustering did not change. The `table` materialization then drops the table with `tables.delete` before the rebuild query. It does not use an atomic `create or replace table`.

This behavior causes two problems:

1. dbt drops and recreates the table on every run. This adds cost. It also creates a period when the table does not exist.
2. dbt loses atomicity. dbt drops the table before the rebuild query runs. If the model SELECT fails, no table remains. An atomic `create or replace` keeps the previous table after a failure.

The related method `_partitions_match` already changes both sides to lower case. The method `_clusters_match` did not. This difference is an oversight, not an intentional design. The same case-sensitive comparison exists back to v1.4.0.

### Solution

Change both sides to lower case in `_clusters_match`. Keep the field order, because clustering uses an ordered list. Accept an empty list or `None`.

BigQuery column names are not case-sensitive. Therefore `externalcustomerid` and `ExternalCustomerId` refer to the same column, and `is_replaceable()` returns `True`. dbt then replaces the model with an atomic `create or replace table` and does not drop it first.

I added a unit test (`TestClustersMatch`) that covers the case-only difference, an exact match, a true mismatch, the order of the fields, an empty list, and a `None` value.

Alternatives and tradeoffs: I considered case normalization at config-parse time. I kept the change in the comparison, because `_partitions_match` uses the same approach. This choice keeps the fix small and does not change the config path.

### Checklist

### DIFF
--- a/dbt-bigquery/.changes/unreleased/Fixes-20260807-000000.yaml
+++ b/dbt-bigquery/.changes/unreleased/Fixes-20260807-000000.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: Compare clustering fields case-insensitively in `is_replaceable` so a `cluster_by`
+    config that differs only by letter case no longer forces an unnecessary drop and
+    recreate of the table
+time: 2026-08-07T00:00:00.000000+00:00
+custom:
+    Author: ddamart
+    Issue: "2111"
--- a/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
@@ -724,12 +724,15 @@ class BigQueryAdapter(BaseAdapter):
         """
         Check if the actual and configured clustering columns for a table
         are a match. BigQuery tables can be replaced if clustering columns
-        match exactly.
+        match exactly. Names are compared case-insensitively because BigQuery
+        column identifiers are not case-sensitive.
         """
         if isinstance(conf_cluster, str):
             conf_cluster = [conf_cluster]
 
-        return table.clustering_fields == conf_cluster
+        return [c.lower() for c in (table.clustering_fields or [])] == [
+            c.lower() for c in (conf_cluster or [])
+        ]
 
     @available.parse(lambda *a, **k: True)
     @record_function(
--- a/dbt-bigquery/tests/unit/test_bigquery_adapter.py
+++ b/dbt-bigquery/tests/unit/test_bigquery_adapter.py
@@ -1304,3 +1304,24 @@ class TestPartitionConfigRenderWrappedInt64Range:
         """int64 without range config should return the raw field name."""
         config = PartitionConfig(field="id", data_type="int64")
         assert config.render_wrapped() == "id"
+
+
+class TestClustersMatch:
+    @pytest.mark.parametrize(
+        "table_fields,conf_cluster,expected",
+        [
+            # case-only difference must still match (BigQuery identifiers are case-insensitive)
+            (["A"], ["a"], True),
+            (["a", "b"], ["a", "b"], True),
+            (["a"], "a", True),  # str config is wrapped into a list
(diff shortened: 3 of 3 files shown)


ddamart:
@cla-bot check

PR #2116, opened 2026-08-11
fix(snowflake): preserve NUMBER precision/scale in catalog.json

### LINKED ISSUE #2114: dbt-snowflake: catalog.json reports every numeric column as bare NUMBER, discarding precision and scale the same query already returns
## Summary

`snowflake__get_catalog_columns_sql` selects `data_type` from
`INFORMATION_SCHEMA.COLUMNS` and nothing else about the type. For Snowflake, `data_type` is
`NUMBER` for **every** numeric column regardless of its declared precision and scale, so
`catalog.json` records `"NUMBER"` for a `NUMBER(12,2)` and for a `NUMBER(38,0)` alike.

`numeric_precision` and `numeric_scale` are columns of the very row being selected. They
are available, correct, and dropped.

Anything that builds on `catalog.json` to reproduce a table's shape — a local engine, a
type-diffing tool, a generated DDL — gets a column with **scale 0** and silently rounds
every value.

## The macro

`dbt-snowflake/src/dbt/include/snowflake/macros/catalog.sql`, in the
`dbt-labs/dbt-adapters` monorepo — verified against `main` on 7 August 2026:

```sql
{% macro snowflake__get_catalog_columns_sql(information_schema) -%}
    select
        table_catalog as "table_database",
        table_schema as "table_schema",
        table_name as "table_name",

        column_name as "column_name",
        ordinal_position as "column_index",
        data_type as "column_type",          -- <- NUMBER, for every numeric column
        comment as "column_comment"
    from {{ information_schema }}.columns
{%- endmacro %}
```

## Reproduction

Any model with a decimal column will do.

```sql
-- models/example.sql
select cast(901.75 as number(12,2)) as retail_price
```

```bash
dbt run && dbt docs generate
```

`target/catalog.json`:

```json
{ "column_name": "RETAIL_PRICE", "column_type": "NUMBER" }
```

What Snowflake already returned for that same row:

```sql
select column_name, data_type, numeric_precision, numeric_scale
from   INFORMATION_SCHEMA.COLUMNS
where  table_name = 'EXAMPLE';
```

```
COLUMN_NAME    DATA_TYPE   NUMERIC_PRECISION   NUMERIC_SCALE
RETAIL_PRICE   NUMBER      12                  2
```

Executed on a real project, this is not an occasional case. Across 16 models with 89
numeric columns, **not one** carried precision or scale in `catalog.json`:

```
NUMBER 89 · TEXT 70 · DATE 16 · TIMESTAMP_NTZ 3 · BOOLEAN 3 · ARRAY 1 · OBJECT 1
```

It also erases scale the author wrote down explicitly: a column defined as
`cast(avg(x) as number(38,4))` is still reported as bare `NUMBER`.

## Why it matters beyond documentation

For the docs site the loss is cosmetic. For anything *consuming* the artefact it is not,
and `catalog.json` is the documented way to learn a model's column types without querying
the warehouse.

Concretely, in a tool that materialises Snowflake data locally: `NUMBER` maps to a
scale-0 decimal, and the landing table then rounds.

```
duckdb> CREATE TABLE landing ("v" DECIMAL(38,0));   -- what "NUMBER" implies
duckdb> INSERT INTO landing VALUES (901.75);
duckdb> SELECT v FROM landing;                       -- 902
```

It rounds. It does not raise. Every money column in the project was silently altered, and
the defect was only found by fingerprinting the same relation on both engines and noticing
that one column disagreed while the row counts matched exactly.

Anyone building on `catalog.json` has this and has no way to know.

## Suggested fix

Include the columns that are already in the result set:

```sql
        data_type as "column_type",
        numeric_precision as "numeric_precision",
        numeric_scale as "numeric_scale",
        character_maximum_length as "character_maximum_length",
```

If adding fields to the catalog schema is a larger change than it looks, an alternative
that keeps the existing shape is to compose the type string in SQL:

```sql
        case
            when data_type = 'NUMBER' and numeric_precision is not null
                then 'NUMBER(' || numeric_precision || ',' || numeric_scale || ')'
            else data_type
        end as "column_type",
```

That is what a reader of `column_type` almost certainly expects it to mean, and it matches
what `describe` on a cursor already returns for the same column.

## Scope

Snowflake specifically, because `data_type` is where Snowflake puts the unparameterised
name. Other adapters may or may not have the equivalent issue depending on what their
information schema returns in that column — worth a look, but this report is about the one
that was executed.

## Workaround, for anyone who finds this before it is fixed

Do not take types from `catalog.json`. `cursor.describe()` on the Snowflake connector
returns `precision` and `scale` per column and costs no scan.

dgvj-work:
resolves #2114

### Problem

`snowflake__get_catalog_columns_sql` selected bare `data_type` from `INFORMATION_SCHEMA.COLUMNS`. For Snowflake, that value is `NUMBER` for every fixed-point numeric column, so `catalog.json` dropped precision and scale even though `numeric_precision` / `numeric_scale` were already on the same row. Downstream consumers that rebuild tables from the catalog then treat decimals as scale 0 and silently round.

### Solution

Compose `NUMBER(p,s)` in the catalog columns SQL when precision is present, matching what `DESCRIBE` returns. Also use expanded numeric `data_type` in `get_catalog_for_single_relation` so the single-relation catalog path stays consistent. Adds unit coverage for the macro, a functional regression test, and updates docs/catalog expectations that previously asserted bare `NUMBER`.

### Checklist

### DIFF
--- a/dbt-snowflake/.changes/unreleased/Fixes-20260811-032000.yaml
+++ b/dbt-snowflake/.changes/unreleased/Fixes-20260811-032000.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Include precision and scale in Snowflake catalog column types for NUMBER
+  columns so catalog.json reports NUMBER(p,s) instead of bare NUMBER
+time: 2026-08-11T03:20:00.000000-05:00
+custom:
+  Author: dgvj-work
+  Issue: "2114"
--- a/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
@@ -349,7 +349,11 @@ class SnowflakeAdapter(SQLAdapter):
         }
 
         catalog_columns = {
-            c.column: ColumnMetadata(type=c.dtype, index=i + 1, name=c.column)
+            c.column: ColumnMetadata(
+                type=c.data_type if c.is_numeric() else c.dtype,
+                index=i + 1,
+                name=c.column,
+            )
             for i, c in enumerate(self.get_columns_in_relation(relation))
         }
 
--- a/dbt-snowflake/src/dbt/include/snowflake/macros/catalog.sql
+++ b/dbt-snowflake/src/dbt/include/snowflake/macros/catalog.sql
@@ -81,7 +81,12 @@
 
         column_name as "column_name",
         ordinal_position as "column_index",
-        data_type as "column_type",
+        -- data_type is bare NUMBER; compose precision/scale for catalog consumers
+        case
+            when data_type = 'NUMBER' and numeric_precision is not null
+                then 'NUMBER(' || numeric_precision || ',' || coalesce(numeric_scale, 0) || ')'
+            else data_type
+        end as "column_type",
         comment as "column_comment"
     from {{ information_schema }}.columns
 {%- endmacro %}
--- a/dbt-snowflake/tests/functional/adapter/catalog_tests/files.py
+++ b/dbt-snowflake/tests/functional/adapter/catalog_tests/files.py
@@ -30,3 +30,14 @@ MY_DYNAMIC_TABLE = """
 ) }}
 select * from {{ ref('my_seed') }}
 """
+
+
+MY_NUMERIC_TABLE = """
+{{ config(
+    materialized='table',
+) }}
+select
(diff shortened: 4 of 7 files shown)


dgvj-work:
CLA has been signed by user: @dgvj-work
let me know if anything else needs to be done.

PR #2127, opened 2026-08-14
fix: respect an explicitly empty cache_schemas set in relation-cache population

### LINKED ISSUE #2126: [Bug] `--cache-selected-only` is a no-op for `dbt test`: empty `cache_schemas` treated as `None`
### Is this a new bug in dbt-adapters?

- [x] I believe this is a new bug in dbt-adapters

### Current Behavior

`BaseAdapter._relations_cache_for_schemas` treats an empty `cache_schemas` set the same as `cache_schemas=None`:

https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-adapters/src/dbt/adapters/base/impl.py#L593-L602

```python
def _relations_cache_for_schemas(
    self,
    relation_configs: Iterable[RelationConfig],
    cache_schemas: Optional[Set[BaseRelation]] = None,
) -> None:
    """Populate the relations cache for the given schemas. Returns an
    iterable of the schemas populated, as strings.
    """
    if not cache_schemas:
        cache_schemas = self._get_cache_schemas(relation_configs)
    ...
```

For `dbt test`, dbt-core's `get_model_schemas` legitimately returns an *empty set*, because test nodes aren't `is_relational`. With `--cache-selected-only` on, that empty set means "no schemas are needed for this selection" — but `if not cache_schemas:` is truthy for `set()`, so it falls through to `_get_cache_schemas(relation_configs)`, which re-expands to every schema in the project. The flag becomes a no-op for `dbt test`.

### Expected Behavior

An explicitly empty `cache_schemas` set should be respected (introspect nothing), and the "compute the default schemas" fallback should only trigger when `cache_schemas is None`.

### Steps To Reproduce

Reported downstream in dbt-labs/dbt-core#12944 (see [this comment](https://github.com/dbt-labs/dbt-core/issues/12944#issuecomment-5222202417)):

1. A project with models spread across many schemas, `cache_selected_only` enabled.
2. `dbt test -s <one model's tests>`
3. Expected: only the selected model's schema is introspected for the relation cache.
4. Actual: every schema in the project is introspected, because `get_model_schemas` returns an empty set for the test selection and `_relations_cache_for_schemas` treats that as "compute the default (all schemas)".

### Relevant log output

_No response_

### Environment

- dbt-adapters: 1.24.5 (also present on `main`)

other (adapter-agnostic — logic lives in `BaseAdapter`)

### Additional Context

Proposed fix (from the linked dbt-core issue comment): change the falsy check to an explicit `is None` check:

```python
if cache_schemas is None:
    cache_schemas = self._get_cache_schemas(relation_configs)
```

This is a companion issue to dbt-labs/dbt-core#12944, which covers the same `--cache-selected-only` gap for `compile`/`show`/`docs generate` (fixed in dbt-labs/dbt-core#15915). This one is being tracked separately because the fix here changes `dbt test` behavior and was flagged as needing a maintainer's call rather than a drive-by fix.

dataders:
resolves #2126

### Problem

`BaseAdapter._relations_cache_for_schemas` uses `if not cache_schemas:` to decide whether to fall back to computing the default (project-wide) set of schemas to introspect. That treats an explicitly empty `cache_schemas` set the same as `cache_schemas=None`.

For `dbt test` with `--cache-selected-only` on, dbt-core's `get_model_schemas` legitimately returns an *empty set* for the selection, because test nodes aren't `is_relational`. That empty set means "no schemas are needed for this selection" — but the falsy check re-expands it to every schema in the project via `_get_cache_schemas(relation_configs)`, making `--cache-selected-only` a no-op for `dbt test`.

### Fix

Use `cache_schemas is None` instead of `not cache_schemas`, so an explicitly empty set is respected as "nothing to cache," while `None` (the default, when the caller didn't specify anything) still falls back to computing the full set.

### Scope

This is a companion fix to dbt-labs/dbt-core#15915, which covers the same `--cache-selected-only` gap for `compile`/`show`/`docs generate` in `dbt-core`. That fix didn't touch this file because the `dbt test` gap lives here in `dbt-adapters` and changes behavior specifically for `dbt test`, which the original reporter flagged as needing a maintainer's call (see [dbt-labs/dbt-core#12944 (comment)](https://github.com/dbt-labs/dbt-core/issues/12944#issuecomment-5222202417)).

`postgres` and `redshift` override `_relations_cache_for_schemas` but only pass `cache_schemas` through to `super()` unmodified, so they pick up this fix automatically without needing their own change.

### Testing

- Added a changelog entry
- Added `TestRelationsCacheForSchemas` in `tests/unit/test_base_adapter.py` asserting: an empty `cache_schemas` set does *not* trigger `_get_cache_schemas`, and `cache_schemas=None` still does
- Full `tests/unit/` suite: 164 passed
- `black --check --line-length=99 --target-version=py39` on the changed files: clean

### DIFF
--- a/dbt-adapters/tests/unit/test_base_adapter.py
+++ b/dbt-adapters/tests/unit/test_base_adapter.py
@@ -494,3 +494,85 @@ class TestGrantsMacroQuotesGrantees:
         # Verify the SQL structure is correct
         assert result.startswith("revoke select on")
         assert "from" in result
+
+
+class TestRelationsCacheForSchemas:
+    @pytest.fixture
+    def adapter(self):
+        config = MagicMock()
+
+        class TestAdapter(BaseAdapter):
+            def convert_boolean_type(self, *args, **kwargs):
+                return None
+
+            def convert_date_type(self, *args, **kwargs):
+                return None
+
+            def convert_datetime_type(self, *args, **kwargs):
+                return None
+
+            def convert_number_type(self, *args, **kwargs):
+                return None
+
+            def convert_text_type(self, *args, **kwargs):
+                return None
+
+            def convert_time_type(self, *args, **kwargs):
+                return None
+
+            def create_schema(self, *args, **kwargs):
+                return None
+
+            def date_function(self, *args, **kwargs):
+                return None
+
+            def drop_relation(self, *args, **kwargs):
+                return None
+
+            def drop_schema(self, *args, **kwargs):
+                return None
+
+            def expand_column_types(self, *args, **kwargs):
+                return None
+
+            def get_columns_in_relation(self, *args, **kwargs):
+                return None
+
+            def is_cancelable(self, *args, **kwargs):
+                return False
+
+            def list_relations_without_caching(self, *args, **kwargs):
+                return []
+
+            def list_schemas(self, *args, **kwargs):
+                return []
+
+            def quote(self, *args, **kwargs):
+                return ""
+
+            def rename_relation(self, *args, **kwargs):
+                return None
+
+            def truncate_relation(self, *args, **kwargs):
+                return None
+
+        return TestAdapter(config, MagicMock())
+
+    def test_empty_cache_schemas_is_not_expanded_to_default(self, adapter):
+        """An explicitly empty `cache_schemas` means "nothing to cache" and
+        must not be re-expanded to every schema in the project (the bug
+        behind dbt-labs/dbt-adapters#2126)."""
+        with patch.object(adapter, "_get_cache_schemas") as mock_get_cache_schemas:
+            adapter._relations_cache_for_schemas([], cache_schemas=set())
+
+        mock_get_cache_schemas.assert_not_called()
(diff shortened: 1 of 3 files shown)


dbt-adapters/tests/unit/test_base_adapter.py
```diff
@@ -494,3 +494,85 @@ def test_revoke_macro_quotes_grantees(
         # Verify the SQL structure is correct
         assert result.startswith("revoke select on")
         assert "from" in result
+
+
+class TestRelationsCacheForSchemas:
+    @pytest.fixture
+    def adapter(self):
```
tauhid621, line 501: Instead of creating a new adapter, we can use existing one

`return BaseAdapterStub(MagicMock(), MagicMock())`

We will need to import it though
`from tests.unit.fixtures.adapter import BaseAdapterStub`

tauhid621 (approved):
Loog good. Left a small comment around test.

PR #2133, opened 2026-08-21
fix(snowflake): map insert_overwrite values by column name, not position

### LINKED ISSUE #1511: [Bug] Wrong generated SQL in enforced contract models that have different column order between yml and sql files
### Is this a new bug?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [x] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

With Snowflake adapter, column order is not always respected in the generated SQL queries, compared to model yml and sql files, in the context of incremental model with contract enforced at least.

### Expected Behavior

Having different ordering of columns between model.yml and model.sql should not be an issue, or should results in an error, depending on your vision.

### Steps To Reproduce

foo.yml

```yaml
version: 2

models:
- name: foo
  config:
      contract:
        enforced: true
  columns:
    - name: column1
      data_type: date
    - name: column2
      data_type: varchar
    - name: column3
      data_type: number
    - name: column4
      data_type: number
```

foo.sql

```sql
{{ config(
  materialized="incremental",
  incremental_strategy="insert_overwrite",
  on_schema_change="append_new_columns"
) }}

with raw as (
  select *
  from {{ ref("bar") }}
)

select
  column1,
  column2,
  column4,
  column3,
from raw
```

As you can see, the order of columns 3 and 4 are swapped between the yml file and the sql final query.

When our production job got triggered by the merge of the PR that introduced the model, the following query get generated in Snowflake : 

```sql
create or replace table database.schema.foo
(
  column1 date,
  column2 varchar,
  column3 number,
  column4 number
)
copy grants
as (
  select column1, column2, column3, column4
  from (
    with raw as (
      select *
      from database.schema.bar
    )
    select
      column1,
      column2,
      column4,
      column3
    from raw
  ) as model_subq
)
```

The query inserts column4 into column3, and vice-versa. This SQL is valid and won't throw any error in Snowflake as column3 and column4 have the same data type. So this results in having the model materialized without any error but inserting wrong data in those columns.

Also, once the model get materialized for the first time, here is the SQL generated by the job run that updates it : 

```sql
create or replace temporary view database.schema.foo__dbt_tmp
(
"COLUMN1",
"COLUMN2", 
"COLUMN4", 
"COLUMN3", 
)
copy grants
as (
  with raw as (
    select *
    from database.schema.bar
  )
  select
    column1
    column2
    column4
    column3
  from raw
)
```

So, the view that dbt creates that will be used for the insert overwrite statement later on follows the column order of the model SQL file, while, as seen above, the deploy initial job follows the column order of the yml file.

Finally, and as a side note as it is not directly related, the insert overwrite final SQL query uses a select * instead of explicit columns selection ; I am not 100% convinced it is totally safe.

### Relevant log output

```shell

```

### Environment

```markdown
dbt Cloud latest version
```

### Additional Context

Originally posted as a support issue through my cloud subscription, copy pasting the answer : 

> Based in the [docs](https://eur02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fdocs.getdbt.com%2Fdocs%2Fmesh%2Fgovern%2Fmodel-contracts%23how-to-define-a-contract&data=05%7C02%7Chugo.epicier%40rolex.com%7C95e5f8f10072458667cc08de3e600a78%7Cf2460eca756e4a3fbd14d2a84590fc31%7C0%7C0%7C639016779408213801%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=DHtReHbjcDzf5Ht6UaYfqrYCeA4lIPJoAFADWNLNIZk%3D&reserved=0) mentioned below, you've identified a gap between the documented behavior and the implementation. The docs state that preflight is "order agnostic" and that DDL will follow contract order—which sets an expectation that order shouldn't matter. But the data insertion step doesn't account for this reordering, resulting in the mismatch you observed.
> When building a model with a defined contract, dbt will do two things differently:
>  
> 1.	dbt will run a "preflight" check to ensure that the model's query will return a set of columns with names and data types matching the ones you have defined. This check is agnostic to the order of columns specified in your model (SQL) or YAML spec.
> 2.	dbt will include the column names, data types, and constraints in the DDL statements it submits to the data platform, which will be enforced while building or updating the model's table, and order the columns per the contract instead of your dbt model.

tauhid621:
resolves #1511

### Problem

The Snowflake `insert_overwrite` strategy built `insert overwrite into <target> select * from <tmp>`, mapping values by **position**. Target and tmp relation don't always agree on column order:

- **Enforced contract:** the target is created via `get_select_subquery` (yaml column order), but this strategy defaults to `tmp_relation_type='view'`, and `snowflake__create_view_as_with_temp_flag` passes the compiled SQL through unchanged (model SQL order). `get_assert_columns_equivalent` validates by name, so a reordered select list passes the contract check and nothing warns.
- **`on_schema_change='append_new_columns'`:** `alter table add column` appends to the end of the target regardless of where the column sits in the model.

Either way values land in the wrong column — silently when the types are compatible, as an opaque cast error when they aren't.

### Solution

Name the columns on both sides so the mapping is by name:

```sql
insert overwrite into <target> ("COL_A", "COL_B") select "COL_A", "COL_B" from <tmp>
```

Columns come from `dest_columns` via `get_quoted_csv` — the same pattern `default__get_delete_insert_merge_sql` and the merge strategies already use. An explicit `overwrite_columns` config still takes precedence and is joined as given, so callers passing pre-quoted identifiers are unaffected.

Two cases change behaviour, both pinned by tests:

1. **Column removed from the model under `append_new_columns`** — used to fail on the `select *` count mismatch, now succeeds and leaves the column NULL. Broader than for `append`/`delete+insert`, since `insert overwrite` truncates first: the column is NULL for *every* row, not only new ones. Raising a compiler error instead is a reasonable alternative — happy to add it, as this combination already failed before.
2. **Renamed column under the default `on_schema_change='ignore'`** — now fails with `invalid identifier` instead of silently writing into the old column.

New tests in `test_insert_overwrite_column_order.py` cover the contract reordering (fails without this change), the appended column, `overwrite_columns` precedence, and the two changes above. Each ordering test runs all three `tmp_relation_type` values; only `view` (this strategy's default) actually diverges, since `table` and `transient` apply `get_select_subquery` under a contract.

### Checklist

- [ ] I have run this code in development and it appears to resolve the stated issue — functional tests need Snowflake credentials and haven't been run yet
- [ ] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX — the two behaviour changes above need a Product/DX look


dbt-snowflake/src/dbt/include/snowflake/macros/materializations/incremental/insert_overwrite.sql
```diff
@@ -24,10 +24,19 @@
...
+    {#-- Name both sides so values map by column name; target and tmp relation do not always
+         agree on column order. https://github.com/dbt-labs/dbt-adapters/issues/1511 --#}
+    {%- if overwrite_columns -%}
+        {#-- joined as given, for callers already passing pre-quoted identifiers --#}
+        {%- set columns_csv = ', '.join(overwrite_columns) -%}
+    {%- elif dest_columns -%}
+        {%- set columns_csv = get_quoted_csv(dest_columns | map(attribute='name')) -%}
+    {%- else -%}
+        {%- set columns_csv = none -%}
+    {%- endif -%}
```
aahel: The `{%- endif -%}` (combined with the `{#--` above) strips every character between `sql_header` and `insert overwrite`, so the two run together. Verified against dbt's own Jinja environment:

```
-- normalize session
alter session set timezone = 'UTC';insert overwrite into DB.SC.TGT ("COL_A", "COL_B")
```

A `;`-terminated header still splits fine, so this only bites when the header's last line is a `--` comment — which then swallows the entire insert. Wrapped in `snowflake_dml_explicit_transaction`, that becomes a `begin; commit;` that writes nothing. Narrow, but it's a one-character fix:

```sql
    {%- endif %}

    insert overwrite into {{ target.render() }} {{ '(' ~ columns_csv ~ ')' if columns_csv }}
```

dbt-snowflake/tests/functional/incremental/test_insert_overwrite_column_order.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def _model(select_list, tmp_relation_type="view", overwrite_columns=None):
+    overwrite = f"    overwrite_columns={overwrite_columns!r},\n" if overwrite_columns else ""
+    return f"""
+{{{{ config(
+    materialized='incremental',
+    incremental_strategy='insert_overwrite',
+    on_schema_change='append_new_columns',
+    tmp_relation_type='{tmp_relation_type}',
```
aahel, line 19: `on_schema_change='append_new_columns'` is hardcoded here, so every ordering test — including `TestInsertOverwriteContractColumnOrder` — goes through `process_schema_changes` and gets `dest_columns` from the *source* relation.

But `on_schema_change` defaults to `ignore` (`incremental.sql` L142), which is the path a contracted model takes unless the user opts out, and it's a materially different branch: `process_schema_changes` returns `{}`, so `dest_columns` falls back to `get_columns_in_relation(existing_relation)` — target columns, in contract order. The fix handles it correctly; it's just the variant of #1511's actual scenario that nothing pins.

Worth parameterizing (`_model(..., on_schema_change="ignore")`, defaulting to `append_new_columns`) so the contract-order test can cover it — `TestInsertOverwriteAppendedColumnOrder` still needs `append_new_columns` for its `alter table add column`.

dbt-snowflake/tests/functional/incremental/test_insert_overwrite_column_order.py
```diff
@@ -0,0 +1,237 @@
...
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"renamed.sql": _uncontracted_model("1 as a, 2 as b", "ignore")}
+
+    def test_rename_is_surfaced(self, project):
+        run_dbt(["run"])
+        run_dbt(["run"])
+        write_file(
+            _uncontracted_model("1 as a, 2 as b2", "ignore"),
```
aahel, line 201: There's a third behaviour change that isn't documented here or in the changelog: under the default `on_schema_change='ignore'`, a model that *gains* a column used to fail loudly (`Insert value list does not match column list expecting 2 but got 3`) because `select *` produced more columns than the target. It now silently drops the new column.

That's consistent with `append` / `merge` via `get_insert_into_sql`, so I think it's the right outcome — but it's a silent change on the default config. Worth a test alongside the rename/drop cases, and a mention in the changelog.

dbt-snowflake/tests/functional/incremental/test_insert_overwrite_column_order.py
```diff
@@ -0,0 +1,237 @@
...
+                f"select column3, column4 from {project.test_schema}.{tmp}_tmp", fetch="one"
+            ) == (33, 44), f"tmp_relation_type={tmp} did not map by column name"
+
+
+class TestInsertOverwriteAppendedColumnOrder:
+    """A column added mid-contract is appended to the end of the target."""
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        models = {
```
aahel: This `models` fixture is byte-identical to `TestInsertOverwriteContractColumnOrder`'s. A shared base (or module-level fixture) would make the difference between the two tests obvious.

Also: the PR description says only `view` actually diverges. That's true for the contract-order test, but not this one — after `alter table add column` the target is `column1, column2, column3, column4, column2b` while a contracted `table`/`transient` tmp is in contract order, so all three `tmp_relation_type`s diverge here and all three are meaningful.

dbt-snowflake/.changes/unreleased/Fixes-20260821-101500.yaml
```diff
@@ -0,0 +1,8 @@
+kind: Fixes
+body: Map values by column name rather than column position in the Snowflake `insert_overwrite`
+  incremental strategy, so an enforced contract or a column appended by `on_schema_change`
+  can no longer silently write values into the wrong columns
```
aahel: Besides the two behaviour changes already called out in the PR body (rename → `invalid identifier`; dropped column under `append_new_columns` → NULL), this also changes the default `on_schema_change='ignore'` path: a newly added column is now silently omitted instead of failing with a column-count mismatch. If we keep that (it matches the other strategies), it belongs in this changelog entry.

aahel (commented):
The approach is sound: naming both sides of `insert overwrite` is the right fix for the positional mapping bug in #1511, `dest_columns` is in scope, and `get_quoted_csv` matches what the other incremental strategies already do.

Three things worth a look — a whitespace trim that can swallow the DML when `sql_header` is set, a gap in coverage around the default `on_schema_change`, and one behaviour change that isn't documented.

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/incremental/insert_overwrite.sql
tauhid621: addressed

dbt-snowflake/tests/functional/incremental/test_insert_overwrite_column_order.py
tauhid621: removed the hardcoding

dbt-snowflake/tests/functional/incremental/test_insert_overwrite_column_order.py
tauhid621: added in the changelg. Thanks for catching this.

aahel (approved):

PR #2138, opened 2026-08-23
Add typed DDL materialization planning contracts and renderer

dataders:
## Summary

Capability-driven DDL/materialization planning: Python resolves strategy facts, offers, and plans; Jinja stays a compatibility-aware renderer, invoked only when a plan calls for it.

- Typed, immutable, serializable create-from-query facts/offers/plans, with rejection reasons and provenance
- Typed incremental mutation facts, schema-change plans, and staging capabilities
- Typed renderer: emits validated CTAS/sql_header SQL, or an explicit legacy-macro fallback with a reason
- Preserves existing adapter dispatch and macro/selector overrides

No adapter-specific behavior yet — Snowflake follow-on lives on `dataders:codex/typed-ddl-snowflake`.

## ELI5

Today, "what to do" and "how to write the SQL" are tangled together in Jinja. This defines a typed "recipe card" format so Python decides what to do first, and only hands off to Jinja/SQL to write the final text. The one recipe implemented end-to-end here is the simple case: create a table from a query (CTAS).

## Validation

- Full `dbt-adapters` unit suite: 232 passed
- Live dbt-duckdb table + incremental runs: passed

## Stack

Base for #2139.


dataders (commented):
maybe python dataclasses are a root cause. but i don't like the IR / data class representation of all this

PR #2139, opened 2026-08-24
Add semantic materialization lifecycle strategies

dataders:
## Summary

Stacked on #2138 — replaces ordered materialization opcodes with typed, executable semantic lifecycle strategies.

- Native view lifecycles: default, Redshift, Snowflake, BigQuery
- Native snapshot lifecycles (initial-create + staged-merge): default, Snowflake
- Preserves project/adapter overrides and custom snapshot strategies on the legacy Jinja path
- Needs companion dbt-core PR #16061 (generic runtime protocol)

## ELI5

#2138 defines the typed "recipe card" format and one easy recipe (create-table-from-query). Views and snapshots aren't one SQL statement — they're multi-step processes that differ per database (Snowflake snapshots work differently than the default approach; BigQuery views differ from Redshift's). This PR writes those harder, multi-step recipes using #2138's format, one version per adapter. It needs a small dbt-core change too, since these multi-step recipes run a bit differently than the old one-shot approach.

## Validation

- 45 planning tests + 15 BigQuery planning tests passed
- CI/adapter coverage still in progress — Snowflake suite can't run offline

Draft, still landing.


dbt-adapters/src/dbt/adapters/base/impl.py
```diff
@@ -112,6 +113,47 @@
...
+    MaterializationTransactionMode,
+    TableLifecyclePlan,
+    TableMaterializationFacts,
+    resolve_table_materialization_operations,
+    resolve_create_from_query_offers,
+    incremental_renderer_macro,
+    incremental_strategy,
+    resolve_incremental_mutation_offers,
+    resolve_incremental_schema_change_plan,
+)
```
dataders, line 148: why does this code not live in dbt-adapters? it sure is a lot to import!

dbt-adapters/src/dbt/adapters/base/impl.py
```diff
@@ -1840,44 +2264,410 @@ def builtin_incremental_strategies(self):
...
+                    requirements=requirements,
+                    reason=reason,
+                    provenance=(
+                        PlanProvenance(
+                            rule="incremental.requested_strategy.unsupported",
+                            detail=reason,
+                        ),
+                    ),
+                ),
             )
```
dataders, line 2452: yuck. there has to be a way to make this control flow easier! maybe there's a way to represent all the conditionals as a rules engine or something??

dbt-adapters/src/dbt/adapters/base/impl.py
```diff
@@ -1840,44 +2264,410 @@ def builtin_incremental_strategies(self):
...
+        )
+
+    @available
+    def plan_incremental_schema_change(
+        self, requested_strategy: Optional[str], default: str = "ignore"
+    ) -> IncrementalSchemaChangePlan:
+        return resolve_incremental_schema_change_plan(requested_strategy, default=default)
+
+    @available
+    def resolve_incremental_lifecycle_plan(
```
dataders: 200 line method??

also does all the plan, get, resolve stuff work?

dbt-adapters/src/dbt/adapters/planning/__init__.py
```diff
@@ -0,0 +1,113 @@
+from dbt.adapters.planning.create_from_query import (
```
dataders: bad smell imho!

dbt-adapters/src/dbt/include/global_project/macros/materializations/models/incremental/incremental.sql
```diff
@@ -4,13 +4,29 @@
   -- relations
   {%- set existing_relation = load_cached_relation(this) -%}
   {%- set target_relation = this.incorporate(type='table') -%}
-  {%- set temp_relation = make_temp_relation(target_relation)-%}
```
dataders: aren't all of these not needed anymore??

PR #2117, opened 2026-08-11
fix(spark): use relation in list_relations_without_caching

dgvj-work:
resolves #480
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) N/A

### Problem

`spark__list_relations_without_caching` used `{{ relation.schema }}`. When callers pass a schema/database name string (for example spark-utils iterating `SHOW DATABASES` results), `.schema` is empty and Spark gets invalid SQL like `show table extended in  like '*'`.

### Solution

Use `{{ relation }}` / `{{ schema_relation }}` in the list-relations macros so both Relation objects and plain schema name strings render correctly. Adds a unit regression check and changelog entry.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-adapters/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

Issue #496, opened 2023-05-08
[ADAP-518] Convert information to a `dict`


Fokko:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt-spark functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

In preparation for having three-part identifiers `catalog.schema.table` (https://github.com/dbt-labs/dbt-adapters/issues/495), I would like to change the `information` attribute on the `SparkRelation` into a `dict`:

https://github.com/dbt-labs/dbt-spark/blob/cb41ab049481bc458871d5c37fad47e59d6b759c/dbt/adapters/spark/relation.py#L36-L37

### Describe alternatives you've considered

The current way is unmaintainable with the regex that extracts useful information from the big blob of text. Also, I noticed that the types are missing currently:

![image](https://user-images.githubusercontent.com/1134248/236912536-9ddcb506-2cc5-4640-81ba-ba2456aaa424.png)

### Who will this benefit?

Mostly the developers because it is hard to maintain right now, and hard to extend the current situation

### Are you interested in contributing this feature?

Yes!

### Anything else?

I wanted to add the database to the configuration. In Spark, this is called a catalog: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Catalog.html

Since Spark 3.0 it can discover tables/views from multiple catalogs, such as a Hive or Glue catalog. I would love to add this, but this refactor needs to be done first and I also want to keep the PRs concise.

Issue #495, opened 2023-05-08
[ADAP-522] Three-part identifiers (`catalog.schema.table`)


dbeatty10:
Spark >= 3.0 can discover tables/views from multiple catalogs, such as a Hive or Glue catalog. This is a prerequisite to enable so-called three-part identifiers for dbt-spark (`catalog.schema.table`).

To keep PRs concise for any refactoring plus the implementation, we are splitting this over multiple issues.

```[tasklist]
### Three-part identifiers
- [ ] https://github.com/dbt-labs/dbt-spark/issues/751
```

Issue #480, opened 2024-06-02
[Bug] `spark__list_relations_without_caching` expects legacy `schema` field


JCZuurmond:
### Is this a new bug in dbt-spark?

- [X] I believe this is a new bug in dbt-spark
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

[spark__list_relations_without_caching](https://github.com/dbt-labs/dbt-spark/blob/944dbea08de9a8069a39ad92f4beecb2cbec608c/dbt/include/spark/macros/adapters.sql#L295) expects legacy field`relation.schema`

``` SQL
{% macro spark__list_relations_without_caching(relation) %}
  {% call statement('list_relations_without_caching', fetch_result=True) -%}
    show table extended in {{ relation.schema }} like '*'
  {% endcall %}

  {% do return(load_result('list_relations_without_caching').table) %}
{% endmacro %}
```

### Expected Behavior

[spark__list_relations_without_caching](https://github.com/dbt-labs/dbt-spark/blob/944dbea08de9a8069a39ad92f4beecb2cbec608c/dbt/include/spark/macros/adapters.sql#L295) expects `relation`

``` SQL
{% macro spark__list_relations_without_caching(relation) %}
  {% call statement('list_relations_without_caching', fetch_result=True) -%}
    show table extended in {{ relation }} like '*'
  {% endcall %}

  {% do return(load_result('list_relations_without_caching').table) %}
{% endmacro %}
```

### Steps To Reproduce

N.A.

### Environment

```markdown
Irrelevant
```

### Additional Context

See Spark SQL [migration guide](https://spark.apache.org/docs/3.2.0/sql-migration-guide.html#upgrading-from-spark-sql-31-to-32)

dgvj-work:
I'd like to take this — planning to update `spark__list_relations_without_caching` to use `{{ relation }}` instead of the legacy `{{ relation.schema }}`, with a regression test.

PR #2140, opened 2026-08-24
feat(spark): support three-part identifiers

### LINKED ISSUE #2129: [Bug] dbt-spark drops the catalog prefix when listing relations in a catalog-qualified schema (schema: my_catalog.my_namespace)
### Is this a new bug?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [x] dbt-spark

### Current Behavior

To reach a Spark v2 (DSv2) catalog, dbt-spark users have to put the catalog in the schema, because `database` is forced to `None` in `SparkCredentials.__post_init__`:

```yaml
schema: my_catalog.my_namespace

list_relations_without_caching then builds its relations from the namespace column returned by SHOW TABLES / SHOW TABLE EXTENDED. But Spark reports that column without the catalog prefix:

spark-sql> show tables in my_catalog.my_namespace;
namespace     tableName   isTemporary
my_namespace  my_table    false

_get_relation_information_using_describe takes that value verbatim (dbt-spark/src/dbt/adapters/spark/impl.py:185):

_schema, name, _ = row
...
table_name = f"{_schema}.{name}"

so _schema is my_namespace, not my_catalog.my_namespace. Three things go wrong as a result:

1. The auxiliary DESCRIBE targets the wrong table. It runs describe extended my_namespace.my_table, which resolves against the default catalog. The resulting DbtRuntimeError is swallowed and logged at debug only (impl.py:196-198), so the run looks clean. If a same-named namespace happens to exist in the default catalog, it is worse than a miss — DESCRIBE succeeds and returns metadata for an entirely different table.

2. Every cache lookup misses. The relation is created with schema="my_namespace" while dbt looks it up as my_catalog.my_namespace, so adapter.get_relation(...) returns None for tables that plainly exist. Incremental models take the create path instead of merging.

The same shape is present in _get_relation_information (impl.py:171-179), the SHOW TABLE EXTENDED path. I have not verified that path against a v2 catalog, since SHOW TABLE EXTENDED errors out for v2 tables anyway (SPARK-33393).

### Expected Behavior

Relations listed for `schema: my_catalog.my_namespace` should carry the schema dbt asked for, not the bare namespace Spark echoes back. Concretely:

1. The auxiliary statement should be `describe extended my_catalog.my_namespace.my_table`, so it resolves in the catalog the profile points at — and cannot silently hit a same-named table in the default catalog.
2. `information` should be populated, so views are typed as `RelationType.View` and `is_delta` / `is_iceberg` / `is_hudi` reflect the actual provider.
3. Listed relations should render as `my_catalog.my_namespace.my_table`, matching the name dbt uses for lookups, so the relation cache hits and an existing incremental model is merged into rather than rebuilt.

In short, `dbt run` against a catalog-qualified schema should behave exactly as it does against a plain one — the catalog prefix should be invisible to everything downstream of `list_relations_without_caching`.

### Steps To Reproduce

1. A Spark cluster with a DSv2 catalog configured, e.g. an Iceberg catalog named `my_catalog`:

   SET spark.sql.catalog.my_catalog=org.apache.iceberg.spark.SparkCatalog;
   SET spark.sql.catalog.my_catalog.type=hadoop;
   SET spark.sql.catalog.my_catalog.warehouse=hdfs:///hdfs_path;

2. Create a namespace and a v2 table in it:

```sql
create namespace if not exists my_catalog.my_namespace;
create table my_catalog.my_namespace.seed_table (id int) using iceberg;
insert into my_catalog.my_namespace.seed_table values (1);

3. Confirm Spark drops the catalog from the namespace column — this is the root cause, and it reproduces without dbt at all:
 show tables in my_catalog.my_namespace;
-- namespace = my_namespace   (not my_catalog.my_namespace)
4. profiles.yml:                                                               
 `my_project:
  target: dev
  outputs:
    dev:
      type: spark
      method: thrift
      host: <host>
      port: <port>
      schema: my_catalog.my_namespace`
5. A single incremental model, models/my_table.sql: 
{{ config(materialized='iiceberg') }} select 1 as id 
6. dbt run — succeeds, creates my_catalog.my_namespace.my_table.
7. dbt --debug run again, and observe two symptoms:                             
  - The debug log contains describe extended my_namespace.my_table — the catalog prefix is missing — followed by Error while retrieving information about my_namespace.my_table.
  - The model is created from scratch rather than merged into, even though the table exists and this is not a --full-refresh.

### Relevant log output

```shell
show tables in prod_dm_dbt.crm

return

namespace
---------
crm
```

### Environment

```markdown
- OS: CentOS 7 Linux
- Python: 3.11.0
- dbt-adapters: 1.7.0
- dbt-spark :  1.10.0a1
```

### Additional Context

````python
def _get_relation_information_using_describe(
        self, relation: BaseRelation, row: "agate.Row"
    ) -> RelationInfo:
        """Relation info fetched using SHOW TABLES and an auxiliary DESCRIBE statement"""
        try:
            _, name, _ = row
        except ValueError:
            raise DbtRuntimeError(
                f'Invalid value from "show tables ...", got {len(row)} values, expected 3'
            )

        table_name = f"{relation.schema}.{name}"
        try:
            table_results = self.execute_macro(
                DESCRIBE_TABLE_EXTENDED_MACRO_NAME, kwargs={"table_name": table_name}
            )
        except DbtRuntimeError as e:
            logger.debug(
                f"Error while retrieving information about {table_name}: {e.msg}"
            )
            table_results = AttrDict()

        information = ""
        for info_row in table_results:
            info_type, info_value, _ = info_row
            if not info_type.startswith("#"):
                information += f"{info_type}: {info_value}\n"

        return relation.schema, name, information
````

currently using relation.schema instead of _schema as a workaround

Amogh-2404:
addresses #495
resolves #2129
related to #480 and #496
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/9867) dbt-labs/docs.getdbt.com/#9867
[design decision](https://github.com/dbt-labs/dbt-spark/issues/281#issuecomment-1026835932)

### Problem

dbt-spark treats relations as two-part `schema.identifier` names and rejects a
distinct profile `database`. Modern Spark supports catalog-qualified
`catalog.namespace.identifier` names.

The existing workaround—placing `catalog.namespace` in `schema`—also fails
during relation discovery because Spark returns only the namespace from `SHOW
TABLES`. This can cause cache misses, incorrect `DESCRIBE` targets, and
incremental models failing to recognise existing relations.

### Solution

- Accept `catalog` as a profile alias for dbt's `database`.
- Use dbt's canonical `database` configuration for per-model and per-source
  catalog selection.
- Render catalog-qualified relations as `database.schema.identifier`.
- Preserve two-part behaviour when no catalog is supplied.
- Preserve legacy profiles where matching `database` and `schema` previously
  collapsed to a two-part identifier.
- Keep discovery, cache keys, column metadata, snapshots, and `docs generate`
  output catalog-aware.
- Preserve database/schema/identifier quote policy through discovery and the
  table, view, and snapshot materializations.
- Reject blank catalog values instead of rendering malformed identifiers.
- Use catalog-aware namespace listing and fully qualified `DESCRIBE`
  statements.
- Preserve the legacy dotted-schema workaround.
- Support documentation generation across multiple databases.

### Compatibility and scope

Restoring the standard `generate_database_name` behaviour means existing
model or source `database` / `+database` values that dbt-spark previously
ignored will now select a Spark catalog. Projects should audit those dormant
values before upgrading. `catalog` is a profile alias only; Spark namespaces
continue to use dbt's `schema` field, and no `namespace` profile alias is added.

The existing relation `information` string remains unchanged; converting it
to a dictionary remains scoped to #496. Catalog-qualified discovery uses the
existing Iceberg-compatible `SHOW TABLES` plus `DESCRIBE` fallback. During docs
generation this can mean two `DESCRIBE` queries per relation because the flat
`information` string does not retain parsed columns. That trade-off was already
accepted in #1870 for Iceberg tables; #496 remains the focused place to improve
the representation and query cost.

This includes the namespace-rendering correction independently proposed by
@dgvj-work in #2117 because full relation rendering is required for
catalog-qualified namespaces. #2117 remains linked as prior overlapping work.

The naming and metadata changes are above connection selection and apply to
`session`, `thrift`, `http`, and `odbc`. The committed functional regression
uses `session`; the external Spark/Iceberg acceptance uses Thrift. HTTP and
ODBC share the same adapter path but still require a server that supports and
configures the named Spark catalog.

### Validation

- `hatch run unit-tests` — 90 passed
- `hatch run code-quality` — passed
- `dagger/run_dbt_spark_tests.py --profile spark_session --test-path tests/functional/adapter/test_three_part_identifiers.py` — 1 passed
- `hatch build` — passed
- `hatch run build:check-all` — passed
- Spark 3.5.8 / Java 17 / Iceberg 1.11.0 named-catalog acceptance — 1 passed

The external Iceberg acceptance used a named `test_catalog` plus a same-name
shadow table in `spark_catalog`. It ran the incremental model twice and the
snapshot twice, then verified qualified queries, relation discovery, schema
lookup, catalog isolation, and model/snapshot docs metadata.

The committed Session regression ran incremental, table, and view models and
asserted the exact quoted three-part `CREATE TABLE` and `CREATE VIEW` SQL.

### Checklist

- [x] I have read the contributing guide and understand what is expected.
- [x] I have run this code in development and it appears to resolve the stated issue.
- [x] This PR includes tests.
- [ ] This interface change has received current Product or DX approval.

### DIFF
--- a/dbt-spark/.changes/unreleased/Features-20260824-193243.yaml
+++ b/dbt-spark/.changes/unreleased/Features-20260824-193243.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add catalog-qualified three-part identifiers to dbt-spark while preserving legacy matching database and schema profiles as two-part identifiers.
+time: 2026-08-24T19:32:43.000000+05:30
+custom:
+  Author: Amogh-2404
+  Issue: "495 2129"
--- a/dbt-spark/.changes/unreleased/Features-20260824-193243.yaml
+++ b/dbt-spark/.changes/unreleased/Features-20260824-193243.yaml
@@ -1,5 +1,5 @@
 kind: Features
-body: Add catalog-qualified three-part identifiers to dbt-spark while preserving legacy matching database and schema profiles as two-part identifiers.
+body: Add catalog-qualified three-part identifiers to dbt-spark. Profile `catalog` maps to dbt's `database`; model and source `database` configurations now select a catalog. Legacy profiles with matching `database` and `schema` remain two-part.
 time: 2026-08-24T19:32:43.000000+05:30
 custom:
   Author: Amogh-2404
--- a/dbt-spark/src/dbt/adapters/spark/connections.py
+++ b/dbt-spark/src/dbt/adapters/spark/connections.py
@@ -51,6 +51,7 @@ import time
 logger = AdapterLogger("Spark")
 
 NUMBERS = DECIMALS + (int, float)
+CATALOG_ALIAS_MARKER = "__dbt_spark_catalog_provided"
 
 # Exception types that indicate a lost connection during query polling
 # Built once at module load time to avoid reconstructing on every poll
@@ -108,11 +109,28 @@ class SparkCredentials(Credentials):
     query_timeout: Optional[int] = None
     poll_interval: int = 5  # Polling interval in seconds for async queries
     query_retries: int = 1  # Number of times to retry on connection loss during query execution
+    _ALIASES = {"catalog": "database"}
+
+    @classmethod
+    def translate_aliases(cls, kwargs: Dict[str, Any], recurse: bool = False) -> Dict[str, Any]:
+        # Profiles translate aliases before Credentials.from_dict runs. Preserve
+        # whether catalog was explicit so the legacy database == schema shape
+        # can remain a two-part identifier.
+        catalog_provided = kwargs.get(CATALOG_ALIAS_MARKER, False) or "catalog" in kwargs
+        translated = super().translate_aliases(kwargs, recurse)
+        if catalog_provided:
+            translated[CATALOG_ALIAS_MARKER] = True
+        return translated
 
     @classmethod
     def __pre_deserialize__(cls, data: Any) -> Any:
         data = super().__pre_deserialize__(data)
+        catalog_provided = data.pop(CATALOG_ALIAS_MARKER, False)
         if "database" not in data:
+            data["database"] = None
+        elif not catalog_provided and data["database"] == data.get("schema"):
+            # Before three-part support, this was the only accepted value for
+            # database and SparkCredentials always discarded it.
             data["database"] = None
         return data
 
@@ -128,15 +146,10 @@ class SparkCredentials(Credentials):
         if self.schema is None:
             raise DbtRuntimeError("Must specify `schema` in profile")
 
-        # spark classifies database and schema as the same thing
-        if self.database is not None and self.database != self.schema:
-            raise DbtRuntimeError(
-                f"    schema: {self.schema} \n"
-                f"    database: {self.database} \n"
-                f"On Spark, database must be omitted or have the same value as"
-                f" schema."
-            )
-        self.database = None
+        if self.database is not None:
+            self.database = self.database.strip()
+            if not self.database:
+                raise DbtConfigError("Catalog cannot be empty")
 
         if self.method == SparkConnectionMethod.ODBC:
             try:
@@ -199,7 +212,7 @@ class SparkCredentials(Credentials):
         return self.host  # type: ignore
 
     def _connection_keys(self) -> Tuple[str, ...]:
-        return "host", "port", "cluster", "endpoint", "schema", "organization"
+        return "host", "port", "cluster", "endpoint", "database", "schema", "organization"
 
 
 class SparkConnectionWrapper(ABC):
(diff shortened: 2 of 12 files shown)


Amogh-2404:
@cla-bot check

PR #2142, opened 2026-08-26
feat(snowflake): add interactive_table materialization

### LINKED ISSUE #1780: [Feature] Snowflake Interactive Table support

itsnamangoyal:
## Summary

Adds the `interactive_table` materialization to dbt-snowflake, covering static and dynamic
(auto-refreshing) interactive tables end to end: config parsing with compile-time validation,
change detection, create/replace/alter/drop/rename DDL, describe readback, and comment-DDL
dispatch.

**This is the v1 (dbt-adapters) port.** The v2/Fusion implementation is dbt-labs/fs#12664 (plus
dbt-labs/fs#13201 and dbt-labs/fs#14103) and was the behavioral reference for this branch. Built
from scratch — not a revival of #2042, which merged and was reverted 4 days later with no recorded
reason.

Resolves #1780.

## What's included

**Materialization and macros** — `materializations/interactive_table.sql` plus a full macro set
under `macros/relations/interactive_table/`: create, replace, alter, drop, rename, describe, and
`ddl_body`.

**Config and change detection** — `relation_configs/interactive_table.py` parses the config and
builds the changeset that decides ALTER vs `CREATE OR REPLACE` per field. Four validations run at
compile time: `cluster_by` presence/shape, iceberg rejection, `transient` rejection, and
`target_lag`/refresh-warehouse pairing.

**Adapter** — `describe_interactive_table`, interactive-table discrimination when listing
relations, and the new relation type registered in `relation.py` / `policies.py`.

**Also fixes dynamic tables** (separate `Fixes` changelog entry) — `cluster_by`, `target_lag`, and
`snowflake_initialization_warehouse` were compared without normalization on dynamic tables, so
every run emitted a no-op ALTER (LINEAR-prefix and case for `cluster_by`, unit for `target_lag`,
case for the warehouse). Normalization now lives in a shared `relation_configs/_normalize.py`
consumed by both materializations.

**Shared, not duplicated** — the target_lag/warehouse ALTER SET clause is one macro
(`relations/target_lag_warehouse_alter.sql`) parameterized by dynamic vs interactive; `create.sql`
and `replace.sql` share `interactive_table/ddl_body.sql`; the ALTER/REPLACE functional test classes
share one fixture base class.

## Parity with the v2/Fusion implementation

This branch now has full behavioral parity with v2/Fusion (dbt-labs/fs#12664 + #14103). Every
review-driven fix on the v2 side was audited against v1 and ported where applicable — 11 fixes in
total:

- **Interactive-warehouse attach/detach removed entirely** to match v2's scope: no
  `snowflake_interactive_warehouses` config, no `warehouses.sql`, no
  `describe_interactive_table_warehouses`, no sync call in the materialization. v1 and v2 now have
  identical surface area.
- `cluster_by` normalization is quote-aware, so a quoted mixed-case identifier (`"MyCol"` vs
  `"mycol"`) is no longer silently treated as unchanged.
- `snowflake__alter_relation_add_remove_columns` ALTERs interactive tables as plain `TABLE`s
  (previously it fell through to invalid `alter interactive_table` DDL) and rejects DROP COLUMN with
  a clear compiler error, since Snowflake rejects it with 010406. Reachable via `incremental` +
  `on_schema_change: sync_all_columns` against a relation that was previously an interactive table.
- `catalog.sql` now populates `stats:last_modified` for interactive tables; it previously only fired
  for `table_type = 'BASE TABLE'`.
- `cluster_by` rejects blank entries *within* a list (`["id", "  "]`), not just a fully-blank value.
- `snowflake_initialization_warehouse` clears (via the `NONE` literal) now take effect on the config
  side, not just on readback — previously a clear would re-fire on every run without ever applying.
- A warning fires when `snowflake_initialization_warehouse` is set on a *static* interactive table
  without `target_lag`, where it is inert. The sibling `snowflake_warehouse` case is deliberately
  left silent: it still selects the build warehouse.
- Whitespace-only warehouse values (`"   "`) are treated as unset across all three sites, instead of
  leaking into invalid DDL.
- Validation errors for `cluster_by`, `transient`, and `target_lag` now name the Snowflake error
  codes they correspond to (010405, 001003, 010412).
- `SHOW ... LIKE` escapes single quotes in interpolated identifiers, so a model or relation name
  containing `'` no longer terminates the SQL string literal early. Fixed at all three call sites —
  `describe_interactive_table`, `describe_dynamic_table`, and the transient-status check. Two of the
  three are pre-existing dynamic-table code, so this carries its own `Fixes` changelog entry.

## Behavior confirmed live against a real Snowflake account

`ALTER INTERACTIVE TABLE` is not in Snowflake's public `ALTER TABLE` reference, but it exists.
Verified live:

- `target_lag`, `refresh_warehouse`, and `snowflake_initialization_warehouse` value-to-value
  changes ALTER in place.
- `cluster_by` changes and any dynamic↔static `target_lag` transition force `CREATE OR REPLACE` —
  Snowflake rejects both directions (errors 001422/001420).
- A static interactive table rejects `TARGET_LAG`, `WAREHOUSE`, and `INITIALIZATION_WAREHOUSE`
  alike; Snowflake validates it as a plain `TABLE`. This caught a real bug — create/replace emitted
  `INITIALIZATION_WAREHOUSE` unconditionally. Fixed, with permanent unit coverage (verified by
  reverting the fix and confirming the new tests fail as expected).
- Every one of the 11 parity fixes above was re-verified against live Snowflake DDL and error
  responses, not just unit-tested. Notably: the DROP COLUMN rejection was confirmed to produce the
  exact compiler error during a live `on_schema_change` run; the `NONE`-clear was confirmed to stop
  re-firing on a third consecutive run; and the quote-escaping fix was confirmed with a negative
  control, showing the pre-fix code genuinely breaks on an apostrophe in a relation name.

## Testing

- **Unit: 418 pass** — the full dbt-snowflake suite, up from 372 at the previous review point.
  130 of the 418 live in the seven test files new on this branch
  (`test_interactive_table_config` / `_alter_macro` / `_create_replace_macros` / `_listing`,
  `test_describe_interactive_table`, `test_rename_relation_dispatch`,
  `test_alter_relation_add_remove_columns_macro`), and 13 further test functions were added to the
  existing dynamic-table config, renamed-relations, and comment-macro tests. Coverage spans config
  parsing and normalization, the changeset builder, `describe_interactive_table`, DDL rendering
  through real Jinja macros against real config objects, column add/remove dispatch, and rename
  dispatch. Every parity fix above was written test-first.
- **Code quality: clean** — all 8 pre-commit hooks, including mypy.
- **Live Snowflake pass: 86 tests run, 0 failures.** This is separate from the CI-gated functional
  suite, which skips by default without a configured account.
  - `tests/functional/relation_tests/interactive_table_tests/` — 22 passed. ALTER-vs-REPLACE per
    field, all four compile-time validations, static no-op idempotency, and
    `on_configuration_change` apply/continue/fail.
  - `tests/functional/relation_tests/test_relation_type_change.py` — 64 passed, 64 skipped, 128
    total. The skips are by design: the matrix enumerates mutually exclusive scenario pairs. The
    interactive-table subset is 15 passed / 15 skipped.
- **Conversion matrix** — `interactive_table` was added to the pairwise
  `test_relation_type_change.py` matrix, covering every conversion into and out of it, and the whole
  matrix now runs live (above). `interactive_table → incremental` requires `--full-refresh`:
  dbt-core's incremental materialization only special-cases an existing view or a `table_format`
  mismatch, so nothing else converts the relation type. Traced from `incremental.sql`, not inferred.

## Known limitations (deliberate)

- Iceberg rejection checks only the model's own `table_format` config, not the
  `catalogs:`/`catalog_name` routing path. A project using that route bypasses the guard; Snowflake
  still rejects the resulting DDL, loudly.
- A whitespace-only `snowflake_warehouse` — the general per-model warehouse override, not specific
  to `interactive_table` — fails with an opaque Snowflake error (002043) raised from a generic
  pre-model `use warehouse` hook, rather than a clear validation message. This predates
  `interactive_table` and affects every materialization, so it is out of scope here.
- `describe_dynamic_table` and the transient-status check take Snowflake's first `SHOW ... LIKE`
  result row without an exact-match filter, so a `_` wildcard in a relation name could in principle
  match a differently-named sibling relation in the same schema. `describe_interactive_table`
  already filters to an exact match; these two did not get the same treatment. This is the same
  class of issue already deferred on the v2/Fusion side, and fixing it here would expand this PR's
  scope into pre-existing dynamic-table behavior.


dbt-snowflake/src/dbt/adapters/snowflake/relation.py
tauhid621: We should also notmalize this maybe?

dbt-snowflake/src/dbt/include/snowflake/macros/adapters.sql
tauhid621, line 220: Do we need to include interactive tables here?

tauhid621 (commented):
Overall looks good. left two comments

dbt-snowflake/src/dbt/include/snowflake/macros/materializations/interactive_table.sql
```diff
@@ -0,0 +1,100 @@
...
+    {% elif full_refresh_mode or not existing_relation.is_interactive_table %}
+        {% set build_sql = get_replace_sql(existing_relation, target_relation, sql) %}
+    {% else %}
+
+        {% set on_configuration_change = config.get('on_configuration_change') %}
+        {% set configuration_changes = snowflake__get_interactive_table_configuration_changes(existing_relation, config) %}
+
+        {% if configuration_changes is none %}
+            {% set build_sql = '' %}
+            {{ exceptions.warn("No configuration changes were identified on: `" ~ target_relation ~ "`. Continuing.") }}
```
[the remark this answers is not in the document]
itsnamangoyal: we intentionally don't want to rebuild static tables unless --full-refresh flag is passed

dbt-snowflake/src/dbt/adapters/snowflake/relation_configs/_normalize.py
```diff
@@ -0,0 +1,157 @@
...
+                start = index + 1
+        index += 1
+    keys.append(text[start:])
+    return keys
+
+
+def normalize_key(key: str) -> str:
+    """A quoted identifier is case-SENSITIVE in Snowflake and kept exactly as
+    written; anything unquoted is folded, matching how Snowflake folds it."""
+    if len(key) >= 2 and key.startswith('"') and key.endswith('"'):
```
tauhid621: Should we consider single quotes as well?

dbt-snowflake/src/dbt/include/snowflake/macros/adapters.sql
itsnamangoyal: thanks, fixed.

dbt-snowflake/src/dbt/adapters/snowflake/relation.py
itsnamangoyal: Thanks, fixed.

tauhid621 (approved):

PR #2147, opened 2026-08-27
feat(snowflake): CREATE OR ALTER for dynamic table query evolution

HeisenBug-QAQ:
Addresses #1825

### Problem

For a dynamic table, `dbt run` compares only config — not the model's SQL. A SQL-only edit
therefore produces no changeset and no-ops, leaving the old query live, so users must remember
`--full-refresh` (unlike table/view/incremental models, which apply SQL changes automatically).

### Solution

Add the `snowflake_dynamic_table_create_or_alter` behavior flag (default off). When enabled,
native dynamic tables sync their definition in place via `CREATE OR ALTER DYNAMIC TABLE` on every
run — so a SQL edit deploys without `--full-refresh`. Change detection is delegated to Snowflake
rather than comparing SQL text. Applies to any `refresh_mode` and any `scheduler` setting: opting
in means accepting `CREATE OR ALTER` semantics wherever they are expressible.

- **INCREMENTAL / AUTO:** a definition edit reinitializes the table (full recompute, incremental
  state discarded) — a documented divergence from dbt incremental models, where a logic change
  requires `--full-refresh`. Called out loudly in the flag description.
- `transient` / `refresh_mode` changes route to `CREATE OR REPLACE` (dbt treats them as full rebuilds).
- Schema changes `CREATE OR ALTER` can't express (column reorder, mid-list insert, type narrowing,
  drop-all) surface Snowflake's error; the fallback is `--full-refresh` (no automatic fallback).
- A no-change run is an idempotent no-op (synced in place, not rebuilt).
- **Iceberg is out of scope for this PR** — the gate is native (`INFO_SCHEMA`) only. Snowflake added
  `CREATE OR ALTER DYNAMIC ICEBERG TABLE` on 2026-08-13; wiring that Iceberg DDL path is a planned
  follow-up, so for now Iceberg dynamic tables keep the standard create-or-replace behavior.

This adds a behavior flag, one `@available` adapter method, and new dynamic-table macros — all
additive and opt-in (default off).

**Tests:** unit tests for the gate + flag, and a functional matrix across `refresh_mode` ×
`scheduler` (FULL / INCREMENTAL / AUTO × disable / enable), plus the routing/veto (transient,
refresh_mode), `on_configuration_change` (fail / continue), incompatible-edit (errors +
`--full-refresh` recovers), transient-drift (errors + recovers), refresh_mode→AUTO (applies in
place), and no-op / idempotency cases.

Supersedes #1851 (deprecated by its author).

### Checklist

### DIFF
--- a/dbt-snowflake/.changes/unreleased/Features-20260827-174315.yaml
+++ b/dbt-snowflake/.changes/unreleased/Features-20260827-174315.yaml
@@ -1,0 +1,12 @@
+kind: Features
+body: 'Add the `snowflake_dynamic_table_create_or_alter` behavior flag. When enabled,
+    native dynamic tables are synced with CREATE OR ALTER DYNAMIC TABLE on every run,
+    so edits to the model''s SQL deploy without `--full-refresh`. Note: for INCREMENTAL
+    and AUTO refresh modes, a definition edit reinitializes the table (incremental
+    state is discarded). Schema changes that CREATE OR ALTER cannot express (e.g.
+    column reorder, mid-list insert) still require `--full-refresh`. Iceberg dynamic
+    tables are unaffected.'
+time: 2026-08-27T17:43:15.027603+00:00
+custom:
+    Author: HeisenBug-QAQ
+    Issue: "1825"
--- a/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
@@ -91,6 +91,28 @@ SNOWFLAKE_MANAGED_ICEBERG_DEFAULT = BehaviorFlag(
     ),
 )
 
+SNOWFLAKE_DYNAMIC_TABLE_CREATE_OR_ALTER = BehaviorFlag(
+    name="snowflake_dynamic_table_create_or_alter",
+    default=False,
+    description=(
+        "When enabled, native (info schema) dynamic tables are synced with CREATE OR ALTER "
+        "DYNAMIC TABLE on every run, so edits to the model's SQL are deployed without "
+        "--full-refresh (matching CTAS ergonomics). WARNING: for INCREMENTAL and AUTO refresh "
+        "modes, a definition edit reinitializes the dynamic table (full recompute, incremental "
+        "state discarded) -- this diverges from dbt incremental models, where a logic change "
+        "requires --full-refresh. CREATE OR ALTER does not support every schema change (e.g. "
+        "column reorder, mid-list insert); incompatible edits still require --full-refresh. "
+        "If `transient` is not set explicitly and the effective transient default diverges from "
+        "the live table (the default flag changed, or an explicit `transient` was removed), dbt "
+        "does not detect it and CREATE OR ALTER attempts a transient flip that Snowflake rejects "
+        "-- recover with --full-refresh. "
+        "With on_configuration_change='continue', a run carrying both a config change and a SQL "
+        "edit skips both. Iceberg dynamic tables are unaffected: this adapter does not apply "
+        "CREATE OR ALTER to them yet (they keep create-or-replace behavior). When disabled "
+        "(default), behavior is unchanged."
+    ),
+)
+
 # Guard against older dbt-adapters that don't have Capability.CatalogsV2 yet.
 # Remove once dbt-adapters lower bound is bumped to the version that adds it.
 _CATALOGS_V2_CAPABILITY = getattr(Capability, "CatalogsV2", None)  # type: ignore[attr-defined]
@@ -176,7 +198,11 @@ class SnowflakeAdapter(SQLAdapter):
 
     @property
     def _behavior_flags(self) -> list[BehaviorFlag]:
-        return [SNOWFLAKE_DEFAULT_TRANSIENT_DYNAMIC_TABLES, SNOWFLAKE_MANAGED_ICEBERG_DEFAULT]
+        return [
+            SNOWFLAKE_DEFAULT_TRANSIENT_DYNAMIC_TABLES,
+            SNOWFLAKE_MANAGED_ICEBERG_DEFAULT,
+            SNOWFLAKE_DYNAMIC_TABLE_CREATE_OR_ALTER,
+        ]
 
     def __init__(self, config, mp_context) -> None:
         super().__init__(config, mp_context)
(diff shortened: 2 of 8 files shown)


dbt-snowflake/.changes/unreleased/Features-20260827-174315.yaml
```diff
@@ -0,0 +1,10 @@
+kind: Features
+body: 'Add the `snowflake_dynamic_table_create_or_alter` behavior flag. When enabled,
+    a FULL-refresh-mode `scheduler=''disable''` native dynamic table is synced with
```
igorbelianski-cyber: any reason not allow incremental DTs here as well( especially if we are guarding this under flag )

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -663,6 +681,28 @@ def build_catalog_relation(self, model: RelationConfig) -> Optional[CatalogRelat
...
+            model (RelationConfig): `config.model` from the jinja context
+        """
+        if not self.behavior.snowflake_dynamic_table_create_or_alter.no_warn:
+            return False
+        dynamic_table = self.Relation.from_config(model)
+        catalog_relation = self.build_catalog_relation(model)
+        return (
+            str(getattr(dynamic_table, "refresh_mode", None)).upper() == "FULL"
+            and str(getattr(dynamic_table, "scheduler", None)).upper() == "DISABLE"
+            and getattr(catalog_relation, "catalog_type", None) == "INFO_SCHEMA"
```
igorbelianski-cyber: what's special about   catalog_type INFO

igorbelianski-cyber: may be add a comment with the link to the docs that this is not supported yet

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -663,6 +681,28 @@ def build_catalog_relation(self, model: RelationConfig) -> Optional[CatalogRelat
...
+        Args:
+            model (RelationConfig): `config.model` from the jinja context
+        """
+        if not self.behavior.snowflake_dynamic_table_create_or_alter.no_warn:
+            return False
+        dynamic_table = self.Relation.from_config(model)
+        catalog_relation = self.build_catalog_relation(model)
+        return (
+            str(getattr(dynamic_table, "refresh_mode", None)).upper() == "FULL"
+            and str(getattr(dynamic_table, "scheduler", None)).upper() == "DISABLE"
```
igorbelianski-cyber: does it match if DISABLED ( is a default choice not explicitly set? (

igorbelianski-cyber:
should we do automated fallback to Create or replace if CoA doesn't support given change: 
RE :    " surface Snowflake's error; the fallback is --full-refresh."

dbt-snowflake/tests/functional/relation_tests/dynamic_table_tests/test_create_or_alter.py
```diff
@@ -0,0 +1,379 @@
...
+
+# COA-incompatible edit: column reorder (Snowflake rejects it in CREATE OR ALTER).
+DT_REORDERED = """
+{{ config(materialized='dynamic_table', snowflake_warehouse='DBT_TESTING',
+          refresh_mode='FULL', scheduler='disable') }}
+select value, id from {{ ref('my_seed') }}
+"""
+
+
+class TestCreateOrAlterTransientFlipRebuilds:
```
igorbelianski-cyber, line 158: is there tests that keeping transient attribute TRUE ( as is)  works with CoA

igorbelianski-cyber (approved):
minor comments

dbt-snowflake/.changes/unreleased/Features-20260827-174315.yaml
HeisenBug-QAQ: as we synced offline, restrict removed

dbt-snowflake/tests/functional/relation_tests/dynamic_table_tests/test_create_or_alter.py
HeisenBug-QAQ: TestCreateOrAlterPreservesTransient added

HeisenBug-QAQ:
> should we do automated fallback to Create or replace if CoA doesn't support given change: RE : " surface Snowflake's error; the fallback is --full-refresh."



> should we do automated fallback to Create or replace if CoA doesn't support given change: RE : " surface Snowflake's error; the fallback is --full-refresh."

as we synced offline, we decided to populate error instead of silent fallback

igorbelianski-cyber:
Comment 1 — missing test for refresh_mode → AUTO gap:

▎ Missing test: refresh_mode transition to AUTO is undetected under COA
▎
▎ dynamic_table_config_changeset deliberately excludes transitions to AUTO from change detection:
▎ if (
▎     new_dynamic_table.refresh_mode != RefreshMode.AUTO
▎     and new_dynamic_table.refresh_mode != existing_dynamic_table.refresh_mode
▎ ):
▎ (relation.py:137-140)
▎
▎ Before this PR, an undetected transition to AUTO just fell into the legacy "no changes identified → skip" branch — harmless. Under COA, the same undetected transition now gets actively re-declared via a live CREATE OR ALTER ... refresh_mode = 'AUTO' ... statement. The flag description already documents this as a possible failure mode ("changing refresh_mode to AUTO... can make CREATE OR ALTER attempt an unsupported in-place change and error"), but I don't see a test exercising it end-to-end. Suggest something like:
▎
▎ class TestCreateOrAlterRefreshModeToAutoUndetectedErrors:
▎     """refresh_mode transitions TO 'AUTO' are excluded from change detection, so under COA
▎     the transition is silently re-declared via CREATE OR ALTER instead of routing to
▎     CREATE OR REPLACE. This pins the documented failure mode end-to-end."""
▎
▎     def test_refresh_mode_to_auto_is_undetected_then_errors_and_full_refresh_recovers(self, project):
▎         update_model(project, "dt_coa", DT_AUTO)  # only refresh_mode differs (FULL -> AUTO)
▎         run_dbt(["run"], expect_pass=False)        # COA attempts the undetected transition, errors
▎         results = run_dbt(["run", "--full-refresh"])
▎         assert len(results) == 1
▎
▎ This would confirm the documented behavior actually happens, and that --full-refresh recovers cleanly, rather than relying solely on prose in the flag description.

Comment 2 — missing test for transient default-drift gap:

▎ Missing test: transient default drift (flag changes, model never sets transient) is undetected under COA
▎
▎ dynamic_table_config_changeset only compares transient when both sides are explicitly known:
▎ if (
▎     new_dynamic_table.transient is not None
▎     and existing_dynamic_table.transient is not None
▎     and new_dynamic_table.transient != existing_dynamic_table.transient
▎ ):
▎ (relation.py:169-173)
▎
▎ If a model never sets transient in its config and relies on the snowflake_default_transient_dynamic_tables behavior flag default, a drift between that default and the live table's actual transient-ness is never flagged. Yet snowflake__create_or_alter_dynamic_table_info_schema_sql recomputes is_transient from the current flag default (create_or_alter.sql:259-267) and re-declares it — attempting to flip transient-ness via CREATE OR ALTER, which Snowflake rejects (transient is create-time-only).
▎
▎ The existing transient tests (TestCreateOrAlterTransientFlipRebuilds, TestCreateOrAlterPreservesTransient) both set transient explicitly in the model config, so they only cover the case the changeset does detect. A test for the "never set transient, drift comes from the flag" path — flip snowflake_default_transient_dynamic_tables between runs with no transient key in the model, confirm the run errors, then confirm --full-refresh recovers — would close the gap between what the flag description claims and what's actually verified.

igorbelianski-cyber (approved):

igorbelianski-cyber:
Pushed a trivial comment wording tweak onto `HeisenBug-QAQ:dt-query-evolution-coa`:

- `(matching CTAS ergonomics)` → `(matching CTAS expirience)`

in the `snowflake_dynamic_table_create_or_alter` behavior-flag description and the matching functional-test module docstring.

Commit: `1d97c39d2` (`docs(snowflake): tweak CREATE OR ALTER comment wording`) authored as igorbelianski-cyber.

PR #2149, opened 2026-08-29
Scope batch source freshness queries to one project at a time

### LINKED ISSUE #2107: [Bug] Batch metadata source freshness returns another project's table metadata when two sources share a schema/dataset name
### Is this a new bug?

- [x] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [x] dbt-bigquery
- [ ] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

With `bigquery_use_batch_source_freshness` enabled, `dbt source freshness` can report the
freshness of a **completely different physical table, in a different BigQuery project**, for a
source. The reported `max_loaded_at` is a real timestamp from the wrong table, so there is no
error and nothing in the logs indicates a problem — the source simply passes, warns, or errors
based on another table's modification time.

This happens when sources point at **different projects (`database`) that use the same dataset
name (`schema`)**. Batch freshness collapses them: *all* source relations end up in a single
batch, that batch is looked up inside **one** project's `INFORMATION_SCHEMA`, and results are
matched back to sources on `(schema, identifier)` with the `database` discarded.

**The result is nondeterministic.** Which project the batch queries depends on which relation
lands at `relations[0]`, and that varies between otherwise identical invocations — so the same
source flips between correct and incorrect freshness from run to run (evidence below).

Two further notes on scope:

- The colliding table **does not need to be declared as a source**. A physical table with a
  matching `schema.table_name` in the queried project is enough to answer for a declared source
  in a different project. Removing the other source declaration does not fix it.
- Silent incorrectness is the bigger concern, not the failures that make it visible. A source
  whose twin in the other project is updated normally will pass freshness indefinitely while the
  real table goes stale.

### Expected Behavior

Source freshness should be scoped to the specified database/project.

### Steps To Reproduce

Two BigQuery projects that both contain a dataset named `stripe`:

| Project | Table | `storage_last_modified_time` |
| --- | --- | --- |
| `project-a` | `stripe.plans` | `2026-04-23T21:21:48.869Z` (stale — pipeline retired) |
| `project-b` | `stripe.plans` | `2026-08-04T02:52:04.591Z` (fresh — updated nightly) |
| `project-b` | `stripe.prices` | `2026-08-04T03:46:50.504Z` (no counterpart in `project-a`) |

`dbt_project.yml`:

```yaml
name: repro
version: '1.0'
profile: default

flags:
  bigquery_use_batch_source_freshness: True
```

`models/sources.yml`:

```yaml
version: 2

sources:
  # Declared first, so it becomes relations[0] for the batch metadata query.
  - name: legacy
    database: project-a
    schema: stripe
    config:
      freshness:
        warn_after: {count: 2, period: day}
        error_after: {count: 4, period: day}
    tables:
      - name: accounts   # NOTE: `plans` is NOT declared under this source,
                         # but a table of that name exists physically in project-a.

  - name: v2
    database: project-b
    schema: stripe
    config:
      freshness:
        warn_after: {count: 2, period: day}
        error_after: {count: 4, period: day}
    tables:
      - name: plans      # fresh in project-b -> should PASS
      - name: prices     # control: no same-named table in project-a
```

```
dbt source freshness
```

Expected behavior: `v2.plans` reads `project-b.stripe.plans` (modified today) and passes.

```
1 of 3 START freshness of legacy.accounts ...... [RUN]
2 of 3 START freshness of v2.plans ............. [RUN]
3 of 3 START freshness of v2.prices ............ [RUN]
2 of 3 ERROR STALE freshness of v2.plans ....... [ERROR STALE in 0.09s]
1 of 3 PASS freshness of legacy.accounts ....... [PASS in 0.24s]
3 of 3 PASS freshness of v2.prices ............. [PASS in 0.76s]
```

`target/sources.json`:

```json
{"id":"source.repro.v2.plans",       "status":"error","max_loaded_at":"2026-04-23T21:21:48.869000+00:00"}
{"id":"source.repro.legacy.accounts","status":"pass", "max_loaded_at":"2026-08-03T22:21:47.090000+00:00"}
{"id":"source.repro.v2.prices",      "status":"pass", "max_loaded_at":"2026-08-04T03:46:50.504000+00:00"}
```

`v2.plans` reports `2026-04-23T21:21:48.869` — that is exactly, to the millisecond,
`project-a.stripe.plans`'s `storage_last_modified_time`. Note that `plans` is not even declared
under the `legacy` source; only the *physical* table needs to exist in `project-a`.

`v2.prices` is the control: because `project-a` has no `stripe.prices`, the batch returns no row
for it, dbt falls back to a per-relation query, and the value is correct. The timings reflect
this — `0.09s` served (wrongly) from the batch vs `0.76s` for the fallback query.

**Emitted SQL**
A single query, against `project-a`, carrying predicates for relations belonging to **both**
projects (from `logs/dbt.log`):

```sql
select table_schema as schema,
       table_name as identifier,
       storage_last_modified_time as last_modified,
       current_timestamp() as snapshotted_at
from `project-a`.`region-US`.INFORMATION_SCHEMA.TABLE_STORAGE
where (upper(table_schema) IN (upper('stripe'))
)
and ((upper(table_schema) = upper('stripe') and
     upper(table_name) = upper('accounts')) or (upper(table_schema) = upper('stripe') and
     upper(table_name) = upper('prices')) or (upper(table_schema) = upper('stripe') and
     upper(table_name) = upper('plans')))
```

`prices` and `plans` live in `project-b`, but are being resolved against `project-a`.

**Nondeterminism**
Selecting one source from each project and running the identical command four times against an
unchanged project — note both the queried project and the resulting status/value change:

```
run 1: from `project-a`.`region-US` -> plans: error 2026-04-23T21:21:48.869+00:00
run 2: from `project-b`.`region-US` -> plans: pass  2026-08-04T02:52:04.591+00:00
run 3: from `project-b`.`region-US` -> plans: pass  2026-08-04T02:52:04.591+00:00
run 4: from `project-b`.`region-US` -> plans: pass  2026-08-04T02:52:04.591+00:00
```

Only **one** `TABLE_STORAGE` query is issued per run even when the selected sources span two
projects, confirming that grouping is not per-database.

This makes the bug hard to spot in CI or scheduled jobs: a freshness check can pass repeatedly and
then fail without any change to code, data, or configuration.

### Relevant log output

```shell

```

### Environment

```markdown
- OS: macOS (arm64)
- Python: 3.12.13
- dbt-adapters: dbt-core==1.12.0b2 + dbt-bigquery==1.11.3
- Also observed on dbt Cloud's "Latest" release track (2026.7.28+709584c), where the behavior flag is enabled by default. This is how we first hit it in production.
```

### Additional Context

**Likely Root Cause**

Two contributing sites:

**1. The result-to-source lookup discards `database`** — `dbt/adapters/base/impl.py`, in
`calculate_freshness_from_metadata_batch`:

```python
# Track schema, identifiers of sources for lookup from batch query
schema_identifier_to_source = {
    (
        source.path.get_lowered_part(ComponentName.Schema),
        source.path.get_lowered_part(ComponentName.Identifier),
    ): source
    for source in sources
}
```

Keyed on `(schema, identifier)` only. Two sources in different databases sharing
`schema.identifier` collapse onto one key, so one silently overwrites the other. This is in the
base adapter and so is not BigQuery-specific.

**2. The batch query targets one project but filters without one** —
`dbt/include/bigquery/macros/metadata.sql`:

```sql
{% set location = adapter.get_dataset_location(relations[0]) %}
{% set information_schema = relations[0].incorporate(location=location).information_schema("TABLE_STORAGE") %}
```

The `INFORMATION_SCHEMA` to query comes from `relations[0]` alone, while the `WHERE` clause
matches only on `table_schema` / `table_name`. Any relation in the batch that is not in
`relations[0]`'s project is therefore either mismatched against a same-named table there, or
missing entirely.

**3. Relations are grouped across databases** — `_get_catalog_relations_by_info_schema` (same
file, ~L634) groups by `relation.information_schema_only()`. Confirmed empirically: selecting
sources from two different projects still produces exactly one `TABLE_STORAGE` query. Worth
checking whether `BigQueryInformationSchema` equality — declared
`@dataclass(frozen=True, eq=False, repr=False)` in `dbt/adapters/bigquery/relation.py:143` — is
comparing on the database component, since `eq=False` suppresses the generated `__eq__` and the
hash/equality behaviour it inherits determines the grouping. Whichever relation then lands at
`relations[0]` is unstable across runs, which is the source of the nondeterminism.

**Suggested Fix**
Include `database` in the lookup key, and either group strictly by database or add a project
qualifier to the batch query's predicates. Any one of these alone would prevent the
cross-project mismatch; the key change also fixes the `database`-blind overwrite.

A guard worth considering regardless: if a relation has no matching row in the batch result,
fall back to the per-relation query (as happens for `prices` today) rather than accepting a row
matched on a partial key.

afonsojanu:
resolves #2107
docs N/A

### Problem

With `bigquery_use_batch_source_freshness` enabled, `dbt source freshness` can report the freshness of a completely different physical table, in a different BigQuery project, for a source. This happens when two sources point at different projects that both use a dataset with the same name.

`calculate_freshness_from_metadata_batch` batches every source into one query regardless of which project it lives in, but the query's `information_schema` is built from `relations[0]`, so it only ever reaches a single project's `TABLE_STORAGE`. Rows that come back are matched to sources by `(schema, identifier)` alone, so a source in one project can be silently answered by a same-named table in whichever project `relations[0]` happened to name. The result is nondeterministic since it depends on iteration order, and there's no error or log line indicating anything went wrong, just a wrong `max_loaded_at`.

### Solution

Group sources by `database` before batching, so each query only ever contains sources from the project it actually reaches. This costs one extra query round trip per distinct project among the sources being checked, which is the correct tradeoff since BigQuery's `INFORMATION_SCHEMA` genuinely is project-scoped and there's no way to query more than one project at once.

Added a unit test with two sources sharing a schema/identifier across two different databases, confirming each triggers its own query and gets matched back to the right result.

### Checklist

- [x] This PR includes tests
- [x] This PR has no interface changes

### DIFF
--- a/dbt-bigquery/.changes/unreleased/Fixes-20260829-114624.yaml
+++ b/dbt-bigquery/.changes/unreleased/Fixes-20260829-114624.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Fix batch metadata source freshness reporting another project's table when
+    two sources in different projects share a dataset/schema name
+time: 2026-08-29T11:46:24.000000+00:00
+custom:
+    Author: afonsojanu
+    Issue: "2107"
--- a/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
+++ b/dbt-bigquery/src/dbt/adapters/bigquery/impl.py
@@ -1235,31 +1235,45 @@ class BigQueryAdapter(BaseAdapter):
                 freshness_responses[source] = freshness_response
             return adapter_responses, freshness_responses
 
-        # Track schema, identifiers of sources for lookup from batch query
-        schema_identifier_to_source = {
-            (
-                source.path.get_lowered_part(ComponentName.Schema),  # type: ignore
-                source.path.get_lowered_part(ComponentName.Identifier),  # type: ignore
-            ): source
-            for source in sources
-        }
+        # The batch query's information_schema is derived from relations[0], so
+        # one call can only ever reach a single project's TABLE_STORAGE. Batching
+        # sources from different projects together in one call, as before, sent
+        # every one of them at whichever project relations[0] happened to name,
+        # and matched rows back by (schema, identifier) alone, so a source in
+        # one project could silently be answered by a same-named table in
+        # another project's dataset. Grouping by database first keeps every
+        # batch scoped to the one project its own query actually reaches.
+        sources_by_database: Dict[Optional[str], List[BaseRelation]] = {}
+        for source in sources:
+            database = source.path.get_lowered_part(ComponentName.Database)  # type: ignore
+            sources_by_database.setdefault(database, []).append(source)
 
-        result = self.execute_macro(
-            GET_RELATION_LAST_MODIFIED_MACRO_NAME,
-            kwargs={
-                "information_schema": None,
-                "relations": sources,
-            },
-            macro_resolver=macro_resolver,
-            needs_conn=True,
-        )
-        adapter_response, table = result.response, result.table  # type: ignore[attr-defined]
-        adapter_responses.append(adapter_response)
+        for database_sources in sources_by_database.values():
+            # Track schema, identifiers of sources for lookup from batch query
+            schema_identifier_to_source = {
+                (
+                    source.path.get_lowered_part(ComponentName.Schema),  # type: ignore
+                    source.path.get_lowered_part(ComponentName.Identifier),  # type: ignore
+                ): source
+                for source in database_sources
+            }
 
-        for row in table:
-            raw_relation, freshness_response = self._parse_freshness_row(row, table)
-            source_relation_for_result = schema_identifier_to_source[raw_relation]
-            freshness_responses[source_relation_for_result] = freshness_response
+            result = self.execute_macro(
+                GET_RELATION_LAST_MODIFIED_MACRO_NAME,
+                kwargs={
+                    "information_schema": None,
+                    "relations": database_sources,
+                },
+                macro_resolver=macro_resolver,
+                needs_conn=True,
+            )
+            adapter_response, table = result.response, result.table  # type: ignore[attr-defined]
+            adapter_responses.append(adapter_response)
+
+            for row in table:
+                raw_relation, freshness_response = self._parse_freshness_row(row, table)
+                source_relation_for_result = schema_identifier_to_source[raw_relation]
+                freshness_responses[source_relation_for_result] = freshness_response
 
         return adapter_responses, freshness_responses
 
--- a/dbt-bigquery/tests/unit/test_calculate_freshness_from_metadata_batch.py
+++ b/dbt-bigquery/tests/unit/test_calculate_freshness_from_metadata_batch.py
@@ -1,13 +1,22 @@
+from types import SimpleNamespace
 from unittest.mock import MagicMock
 
 import pytest
 
 from dbt.adapters.bigquery.impl import BigQueryAdapter
+from dbt.adapters.bigquery.relation import BigQueryRelation
+from dbt.adapters.contracts.relation import RelationType
 
 
 @pytest.fixture
 def adapter():
     return BigQueryAdapter(MagicMock(), MagicMock())
+
+
+def _make_relation(database: str, schema: str, identifier: str) -> BigQueryRelation:
+    return BigQueryRelation.create(
+        database=database, schema=schema, identifier=identifier, type=RelationType.Table
+    )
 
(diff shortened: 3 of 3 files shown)


afonsojanu:
recheck

PR #2155, opened 2026-09-02
fix(snowflake): resolve catalog-linked relation identifier casing

dataders:
## Summary

- recognize Snowflake catalog-linked databases and retain catalog-stored relation casing
- refuse ambiguous case-only relation matches before DML
- apply the stored identity to relation lookup, `this`/`ref`/`source`, and incremental targets

This is intentionally scoped to relation identity (database/schema/table), not column-identifier casing.

## Validation

- unit coverage for catalog-linked matching and stored-case relation creation
- functional, lifecycle, concurrency, and cache-mode coverage is still needed

## Review status

Draft: the current implementation has correctness findings that need resolution before merge; they will be attached as review comments.

### DIFF
--- a/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
+++ b/dbt-snowflake/src/dbt/adapters/snowflake/impl.py
@@ -228,6 +228,41 @@ class SnowflakeAdapter(SQLAdapter):
             return False
         return self._strip_quotes(database).upper() in self._catalog_linked_databases
 
+    def _make_match(self, relations_list, database, schema, identifier):
+        """Refuse to resolve a relation identifier that another relation matches apart from case.
+
+        When the catalog holds two relations whose identifiers differ only by case, neither is a safe
+        answer: the other is equally plausible as the model's target, so binding to either silently
+        can read or overwrite the wrong object. Stock dbt picks whichever one its folded name matches
+        and ignores the twin; here we stop instead. That costs a case which "works" today, and the
+        trade is deliberate -- silently operating on the wrong table is worse than stopping.
+
+        The cache cannot show the collision: it keys on lowercased names, so the pair collapsed to
+        one entry before arriving here. It is recorded during the schema listing, while the catalog's
+        own listing is still un-deduped.
+
+        Everything else defers to default matching, unchanged.
+        """
+        if not self._is_catalog_linked_database(database):
+            return super()._make_match(relations_list, database, schema, identifier)
+
+        if identifier is not None and self._has_case_variants(database, schema, identifier):
+            raise DbtRuntimeError(
+                f"Cannot resolve '{self._strip_quotes(str(identifier))}' in catalog-linked database "
+                f"'{self._strip_quotes(str(database))}': the catalog holds more than one relation "
+                "whose identifier differs only by case. dbt cannot tell which one you mean, and "
+                "choosing either could read or overwrite the wrong object. Rename or remove one of "
+                "them, or set an explicit `alias` and enable `quoting` for the identifier."
+            )
+
+        return super()._make_match(relations_list, database, schema, identifier)
+
+    def _has_case_variants(self, database, schema, identifier) -> bool:
+        def norm(value) -> str:
+            return self._strip_quotes(str(value)).casefold() if value is not None else ""
+
+        return (norm(database), norm(schema), norm(identifier)) in self._case_variant_relations
+
     def _make_match_kwargs(self, database, schema, identifier):
         # if any path part is already quoted then consider same casing but without quotes
         quoting = self.config.quoting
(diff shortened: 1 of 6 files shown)


dbt-snowflake/src/dbt/adapters/snowflake/relation.py
```diff
@@ -82,6 +91,69 @@ def DynamicTable(cls) -> str:
...
+        in the project and render it unquoted, so Snowflake folds it. When the stored name is not
+        that folded form the reference fails with 002003, which surfaces inside the model's own SQL
+        (an `is_incremental()` filter selecting from `{{ this }}`) before any DML is generated.
+
+        Substituting the stored name and quoting it makes the reference resolve exactly. Quoting is
+        the operative part: the stored name often matches the configured name character for
+        character (both `t_orders`) and still fails, because unquoted `t_orders` folds to `T_ORDERS`.
+
+        Applies only where the stored case is known, which is only ever recorded for catalog-linked
+        databases and only once a schema has been listed. Before that -- at parse time, and on a
```
dataders, line 109: [P1] Ensure every runtime relation schema is hydrated

The fix only applies after a schema listing populates _stored_case, but dbt-core cache population does not necessarily list source-only schemas, unselected upstream refs with selected-only caching, or anything under no-populate-cache. In those paths source()/ref() still construct the configured spelling directly and remain unphysicalized. Please either hydrate every this/ref/source schema required by selected nodes before compilation or resolve logical-to-physical identity at runtime. Functional regressions should cover a source-only schema, an unselected ref, and disabled cache population.

dbt-snowflake/src/dbt/adapters/snowflake/relation.py
```diff
@@ -42,6 +42,15 @@ class SnowflakeRelation(BaseRelation):
     relation_configs = {
         SnowflakeRelationType.DynamicTable: SnowflakeDynamicTableConfig,
     }
+
+    # Stored relation-identifier casing for relations in catalog-linked databases, keyed by the
+    # casefolded (database, schema, relation identifier). Populated by SnowflakeAdapter as it lists
+    # a schema; read by create_from below. Only catalog-linked databases are recorded, so membership
+    # doubles as the signal that a relation needs its stored case applied.
+    _stored_case: ClassVar[Dict[Tuple[str, str, str], Tuple[str, str]]] = {}
```
dataders, line 50: [P1] Keep warehouse identity in adapter/cache-instance state

These mutable ClassVars are process-global and production code never clears them. A later adapter, target, parse, or project can inherit stale catalog-linked casing; cache clear/drop/rename does not update this state; and concurrent schema listing performs unlocked read/modify/write. Please move the normalized identity index into the adapter/cache instance, protect it with the cache lifecycle, and add sequential-project plus concurrent-listing coverage.

dbt-snowflake/src/dbt/adapters/snowflake/relation.py
```diff
@@ -82,6 +91,69 @@ def DynamicTable(cls) -> str:
...
+        the operative part: the stored name often matches the configured name character for
+        character (both `t_orders`) and still fails, because unquoted `t_orders` folds to `T_ORDERS`.
+
+        Applies only where the stored case is known, which is only ever recorded for catalog-linked
+        databases and only once a schema has been listed. Before that -- at parse time, and on a
+        first run when the relation does not exist -- the relation is returned unchanged, so nothing
+        about creation changes.
+        """
+        relation = super().create_from(quoting, relation_config, **kwargs)
+        key = cls._stored_case_key(relation)
```
dataders, line 114: [P1] Preserve explicitly quoted identity before applying stored-case adoption

This lookup ignores the effective quote policy already computed by create_from. If the catalog contains quoted Foo and quoted foo, a project explicitly requesting one can still be rewritten to the other stored spelling. The ambiguity check in _make_match also runs before exact matching, so the suggested alias plus quoting remedy cannot actually select an exact quoted candidate. Please make case-insensitive adoption conditional per component: exact matching for quoted components, folding only for unquoted components, and test that an explicitly quoted alias selects only its exact object.

dbt-snowflake/src/dbt/adapters/snowflake/impl.py
```diff
@@ -441,7 +551,39 @@ def list_relations_without_caching(
...
+        of the one it has. Recording it here costs no extra query -- this list is already fetched
+        once per schema per run.
+
+        Two consumers: SnowflakeRelation.create_from applies the stored case to relations built
+        from configuration, and _make_match refuses to resolve an identifier that collides by case.
+        """
+        if not self._is_catalog_linked_database(schema_relation.database):
+            return
+        self.Relation.record_stored_case(relations)
+        seen: Dict[str, str] = {}
```
dataders, line 573: [P2] Detect ambiguity over the complete physical path

This fresh per-listing map only detects identifier twins inside one schema listing. Database or schema names can also differ only by case, and the shared cache folds all three components, so those physical paths can collapse without entering _case_variant_relations. Please maintain one synchronized normalized multimap over the full (database, schema, identifier) tuple and derive exact, adopted, or ambiguous resolution from that single index.

PR #2161, opened 2026-09-04
fix(redshift): unconditionally type cast during sync_all_columns

### LINKED ISSUE #2159: redshift__alter_column_type falls through to an uncasted UPDATE for cross-category type changes, so on_schema_change: sync_all_columns fails

- [x] dbt-redshift

## Current Behavior

`redshift__alter_column_type` uses a native `ALTER COLUMN ... TYPE` for varchar/varbyte size changes and falls through to `default__alter_column_type` for everything else. That fallthrough copies the old column into the new one with no cast:

```sql
update {{ relation }} set "{{ col }}__dbt_alter" = "{{ col }}";
```

Redshift only has implicit assignment casts within a type category, so anything cross-category is rejected:

```
column "my_int_col__dbt_alter" is of type bigint but expression is of type character varying
column "my_ts_col__dbt_alter" is of type timestamp without time zone but expression is of type character varying
```

This is a type-level failure, not a data-level one. It happens even when every value in the column converts cleanly. `varchar(256)` -> `varchar(500)` in the same run succeeded; `varchar` -> `bigint` and `varchar` -> `timestamp` did not.

Net effect: `on_schema_change: sync_all_columns` detects a type change it then can't apply, and the model fails on every run until someone migrates the column by hand outside dbt.

## Expected Behavior

Cast explicitly in the copy step:

```sql
update {{ relation }} set "{{ col }}__dbt_alter" = cast("{{ col }}" as {{ new_type }});
```

Then the migration succeeds whenever the data is convertible, and where it isn't you get a real data error naming the bad value instead of a blanket type mismatch.

Scoped to `redshift__alter_column_type` (`dbt-redshift/src/dbt/include/redshift/macros/adapters.sql`). Not asking to touch the shared `default__alter_column_type` that postgres and others inherit; that's a separate, broader discussion. Redshift already owns its casting policy here via the varchar/varbyte branch, so this just extends that to the branch it currently punts on.

Precedent:

- **dbt-bigquery** (`bigquery/macros/adapters.sql:184-207`) casts unconditionally — `CAST(col AS new_type)`, no whitelist, no gating, every type change on every table. Different mechanism (rebuilds via `create_table_as`, can't ALTER in place), same policy. It also accepts the tradeoff that an explicit cast is more permissive: BigQuery `TIMESTAMP` -> `DATE` truncates silently there too.
- **dbt-athena** (`athena/macros/materializations/models/incremental/on_schema_change.sql:59-90`) uses the identical add/copy/drop/rename shape with `cast()` in the copy step. Iceberg tables only — Hive goes to `alter_relation_replace_columns` instead.
- **dbt-snowflake** / **dbt-spark** delegate to native `ALTER COLUMN ... TYPE` / `SET DATA TYPE`, so the engine's rules decide.

If maintainers would rather not cast, the fallback ask is to detect up front that no implicit cast exists and raise an actionable error suggesting `--full-refresh`, instead of emitting an `UPDATE` that's guaranteed to fail.

## Steps To Reproduce

1. Build an incremental model with a varchar column. Values are cleanly convertible to `bigint`, to show this isn't a data problem:

   ```sql
   -- models/type_change.sql
   {{ config(materialized='incremental', on_schema_change='sync_all_columns') }}

   select 1 as id, '123'::varchar(50) as my_col
   ```

2. `dbt run --select type_change` — builds with `my_col varchar(50)`.

3. Change the column to `bigint`:

   ```sql
   select 1 as id, 123::bigint as my_col
   ```

4. `dbt run --select type_change` — dbt detects the change, then fails:

   ```
   column "my_col__dbt_alter" is of type bigint but expression is of type character varying
   ```

Same shape for `varchar` -> `timestamp`, and generally any pair lacking an implicit assignment cast.

## Relevant log output

```
    In "<db>"."<schema>"."my_incremental_model":
        Schema changed: True
        Source columns not in target: []
        Target columns not in source: []
        New column types: [{'column_name': 'my_int_col', 'new_type': 'bigint'}, {'column_name': 'my_ts_col', 'new_type': 'timestamp without time zone'}]

On model.<project>.my_incremental_model:
    alter table "<db>"."<schema>"."my_incremental_model" add column "my_int_col__dbt_alter" bigint;
SQL status: SUCCESS in 9.928 seconds

On model.<project>.my_incremental_model:
    update "<db>"."<schema>"."my_incremental_model" set "my_int_col__dbt_alter" = "my_int_col";
Redshift adapter: Redshift error: column "my_int_col__dbt_alter" is of type bigint but expression is of type character varying
On model.<project>.my_incremental_model: ROLLBACK
Redshift adapter: Not retrying error: Database Error
  column "my_int_col__dbt_alter" is of type bigint but expression is of type character varying
```

## Environment

```markdown
- OS: dbt Platform runtime ("Latest" release track)
- Python:
- dbt-adapters: 1.24.5
- dbt-redshift: built from dbt-adapters commit 7d31313462e06637800f3c78ffc96813d9000ab3
```

## Additional Context

Not a data-corruption risk: the whole add/update/drop/rename sequence runs in one open transaction, so the failed `UPDATE` rolls back cleanly rather than leaving a half-altered column.

The cost is wasted work repeated forever. In this run three earlier column-widening migrations in the same `sync_all_columns` pass had already succeeded before the failing column was reached, and all three rolled back with it. The whole pass (817s) gets thrown away and retried from scratch on every subsequent run.

Only became reachable after #2097 fixed temp-relation introspection on datashare consumer connections. Before that the source columns came back empty, so dbt never detected the drift; now it detects it correctly and fails applying it.

trouze:
resolves #2159 
no docs changes needed

### Problem

`redshift__alter_column_type` handles varchar/varbyte size changes [with a native](https://github.com/dbt-labs/dbt-adapters/blob/2d5a9cef960651178bb66bb068eb0efaca8c7c59/dbt-redshift/src/dbt/include/redshift/macros/adapters.sql#L477) `ALTER COLUMN ... TYPE` and falls through to `default__alter_column_type` for everything else. That [fallthrough](https://github.com/dbt-labs/dbt-adapters/blob/2d5a9cef960651178bb66bb068eb0efaca8c7c59/dbt-adapters/src/dbt/include/global_project/macros/adapters/columns.sql#L124) copies the old column into the new one with no cast.

```sql
update {{ relation }} set "{{ col }}__dbt_alter" = "{{ col }}";
```

Redshift only has implicit assignment casts within a type category, so cross-catagory changes are rejected outright:

```
column "my_int_col__dbt_alter is of type bigint but expression is of type character varying
```

This is a type-level failure of the SQL being issued, not necessarily indicative of what Redshift can and can't do. `on_schema_change: sync_all_columns` can detect the change, but will fail to apply a cast due to the above. This forces customers on Redshift to migrate data types by hand and, as I call out in the [issue](https://github.com/dbt-labs/dbt-adapters/issues/2159), we have precedents for unconditionally type casting with BigQuery, Athena (when using iceberg tables), and we delegate this to Snowflake and Spark who have native rules when issuing `ALTER COLUMN ... TYPE / SET DATA TYPE` statements.

### Solution

I propose we replace the fallback to `default__alter_column_type` with something similar, just casting to the new column type instead.

```sql
update {{ relation }} set "{{ col }}__dbt_alter" = cast("{{ col }}" as {{ new_column_type }});
```

Alternatives considered:
- We could do nothing at all and put it on users to explicitly cast in their models, override this themselves, or full refresh/manually alter column types by hand. While I think there's precedent for unconditionally casting, at a minimum we should improve the error messaging to suggest that there's a data type mismatch and users should perform a full refresh to resolve the error
- We could attempt to whitelist a set of "safe pairs", but this would put burden on us to keep it in-sync rather than simply delegating to the database, and BigQuery doesn't do this

### Checklist

### DIFF
--- a/dbt-redshift/.changes/unreleased/Fixes-20260904-120000.yaml
+++ b/dbt-redshift/.changes/unreleased/Fixes-20260904-120000.yaml
@@ -1,0 +1,8 @@
+kind: Fixes
+body: Cast explicitly when migrating a column type in `redshift__alter_column_type`, so
+    `on_schema_change='sync_all_columns'` can apply type changes across type categories
+    (e.g. varchar to bigint) instead of failing on an uncasted copy
+time: 2026-09-04T12:00:00.000000-07:00
+custom:
+    Author: trouze
+    Issue: "2159"
--- a/dbt-redshift/src/dbt/include/redshift/macros/adapters.sql
+++ b/dbt-redshift/src/dbt/include/redshift/macros/adapters.sql
@@ -462,13 +462,13 @@
 {% macro redshift__alter_column_type(relation, column_name, new_column_type) -%}
   {#
     Redshift ALTER COLUMN TYPE only supports VARCHAR and VARBYTE (size changes).
-    For those, use native ALTER; for any other type change, fall back to
-    default add/copy/drop/rename.
+    For those, use native ALTER; for any other type change, migrate the column
+    with add/copy/drop/rename.
 
     The native ALTER TABLE ALTER COLUMN cannot run inside a transaction block.
     It is only safe to use when `redshift_skip_autocommit_transaction_statements`
     is enabled (i.e. we are not wrapping statements in BEGIN/COMMIT).
-    When the flag is off, always use the default migration path.
+    When the flag is off, always use the migration path.
   #}
   {% set type_lower = (new_column_type | lower) | trim %}
   {% set skip_txn = adapter.behavior.redshift_skip_autocommit_transaction_statements.no_warn %}
@@ -477,7 +477,28 @@
       alter table {{ relation.render() }} alter column {{ adapter.quote(column_name) }} type {{ new_column_type }}
     {% endcall %}
   {% else %}
-    {{ default__alter_column_type(relation, column_name, new_column_type) }}
+    {#
+      1. Create a new column (w/ temp name and correct type)
+      2. Copy data over to it, casting explicitly
+      3. Drop the existing column (cascade!)
+      4. Rename the new column to existing column
+
+      This mirrors `default__alter_column_type`, except the copy step casts
+      instead of relying on an implicit assignment cast. Redshift only has
+      implicit assignment casts within a type category, so an uncasted copy
+      fails outright for pairs like varchar -> bigint even when every value
+      would convert cleanly. The explicit cast lets the warehouse's own cast
+      rules decide, so the migration succeeds whenever the data is actually
+      convertible and otherwise fails with the offending value in the error.
+    #}
+    {%- set tmp_column = column_name + "__dbt_alter" -%}
+
+    {% call statement('alter_column_type') %}
+      alter table {{ relation.render() }} add column {{ adapter.quote(tmp_column) }} {{ new_column_type }};
+      update {{ relation.render() }} set {{ adapter.quote(tmp_column) }} = cast({{ adapter.quote(column_name) }} as {{ new_column_type }});
+      alter table {{ relation.render() }} drop column {{ adapter.quote(column_name) }} cascade;
+      alter table {{ relation.render() }} rename column {{ adapter.quote(tmp_column) }} to {{ adapter.quote(column_name) }}
+    {% endcall %}
   {% endif %}
 {% endmacro %}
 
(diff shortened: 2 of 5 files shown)


trouze:
copilot looks to be wrong - this line is directly from [here](https://github.com/dbt-labs/dbt-adapters/blob/2d5a9cef960651178bb66bb068eb0efaca8c7c59/dbt-redshift/src/dbt/include/redshift/macros/adapters.sql#L475) and unit tests pass

PR #2164, opened 2026-09-08
Fix(snowflake): honor transient config for seeds (CORE-903)

itsnamangoyal:
## Summary

Fixes [CORE-903](https://dbtlabs.atlassian.net/browse/CORE-903): `+transient: true` (and `transient: false`) was accepted as a config for Snowflake seeds but had no effect — seeds were always created as permanent tables.

**Root cause.** The Snowflake `seed` materialization calls `materialization_seed_default()`, which dispatches to `default__create_csv_table` — a plain `create table` with no adapter-specific DDL-prefix logic. Regular tables instead route through `snowflake__create_table_info_schema_sql`, which prefixes `transient` based on `adapter.build_catalog_relation(config.model).is_transient`. Seeds were the only resource type bypassing that path.

**Fix.** Add a `snowflake__create_csv_table` override in `dbt-snowflake/src/dbt/include/snowflake/macros/materializations/seed.sql`. It mirrors the default column DDL but sources the transient decision from the same `catalog_relation.is_transient` the table create path uses, so seeds and tables can't drift. `snowflake__reset_csv_table` (the full-refresh path) already calls the dispatched `create_csv_table(...)`, so re-seeding with `--full-refresh` picks up the fix automatically.

Companion fix for the Fusion (v2) engine: [dbt-labs/fs#14538](https://github.com/dbt-labs/fs/pull/14538).

## Test plan

TDD, functional (live Snowflake), mirroring `test_incremental_transient.py`'s `run_dbt_and_capture` pattern since this repo has no Jinja-macro unit-test harness:
- `TestSeedTransientDefault` — default config → compiled/executed SQL contains `create transient table`.
- `TestSeedTransientFalse` — `+transient: false` → plain `create table`, no `transient` keyword.

Both confirmed to fail before the fix (today's seed DDL never emits `transient`) and pass after. Also re-ran the existing `simple_seed`, `custom_schema_tests`, and `statement_test` suites — no regressions.

Changelog: `dbt-snowflake/.changes/unreleased/Fixes-20260908-190024.yaml`.

[CORE-903]: https://dbtlabs.atlassian.net/browse/CORE-903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


r-jais:
@itsnamangoyal Could you confirm this will not cause any failures for iceberg_rest catalogs during dbt seed?

PR #2166, opened 2026-09-10
fix(snowflake): interactive_table transient handling and dbt1005 message

itsnamangoyal:
## Summary
- Correct the dbt1005 `cluster_by` compilation-error message: it said "at least one non-blank column" but the code actually requires *every* entry to be non-blank, even alongside a valid one.
- Stop rejecting `transient=true` for `interactive_table` models. `transient` has no valid DDL for this materialization and is never emitted either way, so treat it as inert (like `change_tracking`) instead of raising. There's no way to distinguish an explicit per-model `transient: true` from one inherited via a project-wide `+transient` default, so the old check also broke the common case of a project-wide default reaching an interactive_table model that never mentions `transient` itself.
- Mirrors the equivalent fix on the v2 (Fusion) side: dbt-labs/fs#14733.

## Test plan
- [x] Unit tests updated and passing: `hatch run unit-tests -- tests/unit/test_interactive_table_config.py` (101 passed)
- [x] `hatch run code-quality` clean (black, flake8, mypy)
- [x] Functional tests updated and passing live against Snowflake (`ktb38830`), including the new `test_transient_true_is_inert`: `hatch run integration-tests -- tests/functional/relation_tests/interactive_table_tests/test_basic.py::TestCompileValidation` and `::TestBasic`


itsnamangoyal:
Closing to split into two separate PRs (dbt1005 message fix, transient-ignore fix) tracked individually in Jira. Will not resume until code freeze ends 2026-09-21.

PR #2168, opened 2026-09-12
fix(snowflake): correct dbt1005 message for interactive_table cluster_by

itsnamangoyal:
The dbt1005 validation message said cluster_by needs 'at least one non-blank column', but the code actually rejects any blank entry, even alongside valid ones. Fixed the message to match the real rule.

Tracks CORE-943. Split out of dbt-labs/dbt-adapters#2166.


itsnamangoyal:
⚠️ Do not merge until code freeze ends 2026-09-21.

ash2shukla (approved):

PR #2169, opened 2026-09-12
fix(snowflake): stop rejecting transient=true for interactive_table

itsnamangoyal:
transient has no valid DDL for interactive_table and is never emitted either way, so treat it as inert like change_tracking instead of hard-erroring. There's no way to tell an explicit per-model value from one inherited via a project-wide +transient default, so the error was rejecting that common case too. Verified live against Snowflake: a project-level transient default now builds successfully instead of failing.

Tracks CORE-944. Split out of dbt-labs/dbt-adapters#2166.


itsnamangoyal:
⚠️ Do not merge until code freeze ends 2026-09-21.

PR #2172, opened 2026-09-14
fix(postgres): name built-in types psycopg2 does not register during contract enforcement

### LINKED ISSUE #2170: [Bug] [dbt-postgres] Contract enforcement cannot name `uuid` (and other built-in types) because psycopg2 registers no typecaster for them
### Is this a new bug?

- [ ] dbt-adapters
- [ ] dbt-tests-adapter
- [ ] dbt-athena
- [ ] dbt-athena-community
- [ ] dbt-bigquery
- [x] dbt-postgres
- [ ] dbt-redshift
- [ ] dbt-snowflake
- [ ] dbt-spark

### Current Behavior

With `contract: enforced` on a model that has a `uuid` column, `dbt run` emits, for every such column:

```
The `type_code` 2950 was not recognized, which may affect error messages for enforced contracts that fail as well as `Column.data_type` values returned by `get_column_schema_from_query`
```

and compares the column as `unknown type_code 2950`. When a contract does fail, the mismatch table names the column's type `unknown type_code 2950` instead of `uuid`. Under `--warn-error` (or `DBT_ENGINE_WARN_ERROR_OPTIONS='{"error": "all"}'`) the event is promoted and every contracted model with a `uuid` column fails to build.

`uuid` is a PostgreSQL built-in with a fixed OID (2950), as are `money` (790), `inet` (869), `cidr` (650), `macaddr` (829), `bit` (1560), `bit varying` (1562), the geometric types and others, so none of them can be named today.

### Expected Behavior

Built-in PostgreSQL types are named during contract enforcement, so a valid `uuid` contract produces no event and a failing one reports `uuid`. Only types whose OID is assigned per database (enums, domains, extension types such as PostGIS) should still need the `unknown type_code <oid>` fallback.

### Steps To Reproduce

```sql
-- models/my_model.sql
select cast('00000000-0000-0000-0000-000000000000' as uuid) as id
```

```yaml
# models/schema.yml
models:
  - name: my_model
    config:
      contract:
        enforced: true
    columns:
      - name: id
        data_type: uuid
```

```
dbt --warn-error run
```

### Relevant log output

```
Compilation Error in model my_model
  [WARNING]: The `type_code` 2950 was not recognized, which may affect error
  messages for enforced contracts that fail as well as `Column.data_type`
  values returned by `get_column_schema_from_query`
  > in macro get_column_schema_from_query (macros/adapters/columns.sql)
  > called by macro assert_columns_equivalent (macros/relations/column/columns_spec_ddl.sql)
```

### Relevant log output

```shell

```

### Environment

```markdown
- OS: Linux
- Python: 3.11
- dbt-core: 1.12.4
- dbt-postgres: 1.11.0 (and unchanged on main as of 2026-09-14)
- psycopg2-binary: 2.9.13
```

### Additional Context

`PostgresConnectionManager.data_type_code_to_name` looks the OID up in `psycopg2.extensions.string_types`, which only holds the types psycopg2 has a typecaster for. psycopg2 treats `uuid` as an optional extra (`psycopg2.extras.register_uuid()`), which dbt-postgres never calls.

History: dbt-core #8353 (the original uuid report, 2023), #8877, and #8887 (which turned the crash into the current warning + placeholder). #8357 proposed calling `register_uuid()` at connection time and was closed when adapter code moved repositories. #8900 (a user-space type map) is still open. dbt-redshift ships a `TYPE_OID_TO_DATA_TYPE` table for the same class of problem.

I have a PR ready that adds the equivalent table for dbt-postgres, consulted after psycopg2 and before the fallback, with unit tests, updated functional tests, and a functional test that checks every mapped OID against the server's catalog.

tuday2:
resolves #2170

### Problem

`PostgresConnectionManager.data_type_code_to_name` names a result column's type by looking its OID up in `psycopg2.extensions.string_types`. psycopg2 registers no typecaster for `uuid`, `money`, `inet`, `cidr`, `macaddr`, `bit`, `bit varying`, the geometric types and a number of other PostgreSQL built-ins, so contract enforcement on a model with such a column emits `TypeCodeNotFound` and compares `unknown type_code <oid>`. A failing contract then names the type by number, and under `--warn-error` every contracted model with a `uuid` column fails to build.

Prior history: dbt-core #8353, #8877, #8887 (which made this a warning instead of a crash), #8357 (register the psycopg2 typecaster; closed unmerged on the repo move), #8900 (still open).

### Solution

Add `TYPE_OID_TO_DATA_TYPE`: the PostgreSQL built-in base types psycopg2 does not register, plus their array types, keyed by OID and valued by the name `format_type(oid, null)` returns. It was generated from `pg_catalog.pg_type` on PostgreSQL 16.13 (`typtype = 'b'`, excluding the pseudo, polymorphic and internal-use categories) minus the OIDs psycopg2 2.9.13 registers. This is the same shape as dbt-redshift's `TYPE_OID_TO_DATA_TYPE`.

`data_type_code_to_name` now consults psycopg2 first, then the table, then emits `TypeCodeNotFound` as before. So:

- every type psycopg2 does name keeps exactly the name it had (no change for existing users);
- the built-ins it does not name are named (`uuid`, `money`, `inet`, …), so valid contracts on them produce no event and failing ones report the type by name;
- user-defined and extension types, whose OIDs are assigned per database, still take the placeholder path unchanged.

Registering psycopg2's `uuid` typecaster instead (the #8357 approach) was deliberately not done: it would change the Python type `run_query` returns for uuid columns everywhere, and it would fix one type rather than the set.

### Tests

- `tests/unit/test_connection.py`: driver-registered types keep the driver's name; table types are named; the table and the driver's set are disjoint (so precedence can never matter); known codes do not warn; an unknown code still warns once with `TypeCodeNotFound` and returns the placeholder.
- `tests/functional/contracts/test_nonstandard_data_type.py`: the three `money` cases now assert `money` is named and no event is emitted (they previously asserted `unknown type_code 790`); two new `uuid` cases (valid contract passes without an event; `uuid` against a `text` contract fails naming both); one new case with an enum created in the test schema proving the placeholder path is unchanged for a per-database OID.
- `tests/functional/test_type_oid_mapping.py`: asks the server under test for `format_type(oid, null)` of every mapped OID and requires agreement, and requires every OID to exist on PostgreSQL 13+ (the newest mapped types, `xid8`, `pg_snapshot` and `regcollation`, arrived in 13).

Run locally against PostgreSQL 16.13: `tests/unit/test_connection.py` 19 passed / 2 skipped (pre-existing skips); `tests/functional/contracts` + `tests/functional/test_type_oid_mapping.py` 11 passed; `tests/functional/test_types.py` and `test_column_quotes.py` unchanged and passing. `black`, `flake8` and `mypy` clean on the changed module.

### Checklist


tuday2:
@cla-bot check

